PR #4, opened 2021-06-28
Fixes for some memory leaks

samuel-lee-msft:
+ In sc_ossl_rsa_finish free exdata if it is non-NULL
  + And do not dereference exdata if it is NULL
+ In ECC, do not initialize exdata until we know we will definitely be
  using it (and not falling back to default implementation)
  + Rather than using the curve to determine whether to free exdata
    members, use the initialized field
  + Always free the exdata if it is non-NULL
+ Also in ECDSA, handle DER encoded ECDSA signatures for P192 and P224

### DIFF
--- a/SymCryptEngine/src/sc_ossl_ecc.c
+++ b/SymCryptEngine/src/sc_ossl_ecc.c
@@ -27,12 +27,6 @@ typedef int (*PFN_eckey_verify)(
 typedef int (*PFN_eckey_verify_sig)(
                             const unsigned char* dgst, int dgst_len,
                             const ECDSA_SIG* sig, EC_KEY* eckey);
-typedef int (*PFN_eckey_init)(EC_KEY *key);
-typedef void (*PFN_eckey_finish)(EC_KEY *key);
-typedef int (*PFN_eckey_copy)(EC_KEY *dest, const EC_KEY *src);
-typedef int (*PFN_eckey_set_group)(EC_KEY *key, const EC_GROUP *grp);
-typedef int (*PFN_eckey_set_private)(EC_KEY *key, const BIGNUM *priv_key);
-typedef int (*PFN_eckey_set_public)(EC_KEY *key, const EC_POINT *pub_key);
 
 typedef int (*PFN_eckey_keygen)(EC_KEY *key);
 typedef int (*PFN_eckey_compute_key)(unsigned char **psec,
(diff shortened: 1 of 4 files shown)


SymCryptEngine/src/sc_ossl_ecc.c
```diff
@@ -748,8 +698,21 @@ int sc_ossl_get_context_ex(EC_KEY* eckey, SC_OSSL_ECC_KEY_CONTEXT** ppKeyCtx, BO
...
-        SC_OSSL_LOG_ERROR("SymCrypt Context Not Found.");
-        return SC_OSSL_ECC_GET_CONTEXT_ERROR;
+        SC_OSSL_ECC_KEY_CONTEXT *keyCtx = OPENSSL_zalloc(sizeof(*keyCtx));
+        if( !keyCtx )
+        {
+            SC_OSSL_LOG_ERROR("OPENSSL_zalloc failed");
+            return SC_OSSL_ECC_GET_CONTEXT_ERROR;
+        }
+
+        if( EC_KEY_set_ex_data(eckey, eckey_sc_ossl_idx, keyCtx) == 0)
```
mlindgren, line 708: >if( EC_KEY_set_ex_data(eckey, eckey_sc_ossl_idx, keyCtx) == 0) [](http://example.com/codeflow?start=8&length=62)

> All functions with a TYPE of DH, DSA, RSA and EC_KEY are deprecated. Applications should instead use EVP_PKEY_set_ex_data(), EVP_PKEY_get_ex_data() and EVP_PKEY_get_ex_new_index().

Just confirming: it's a known issue that we don't implement the EVP interface for ECC, making this required for now, right?

mlindgren (approved):
:shipit:

SymCryptEngine/src/sc_ossl_ecc.c
samuel-lee-msft: Correct - we should move the RSA and ECC support to the EVP layer going forward, but for now we rely on the deprecated lower level APIs (which are still available in 1.1.1)

SymCryptEngine/src/sc_ossl_ecc.c
```diff
@@ -748,8 +698,21 @@ int sc_ossl_get_context_ex(EC_KEY* eckey, SC_OSSL_ECC_KEY_CONTEXT** ppKeyCtx, BO
 
     if( *ppKeyCtx == NULL )
     {
-        SC_OSSL_LOG_ERROR("SymCrypt Context Not Found.");
-        return SC_OSSL_ECC_GET_CONTEXT_ERROR;
+        SC_OSSL_ECC_KEY_CONTEXT *keyCtx = OPENSSL_zalloc(sizeof(*keyCtx));
```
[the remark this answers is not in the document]
samuel-lee-msft: This is used elsewhere in the engine code already. I was also on the fence about whether it is clearer to use `sizeof(SC_OSSL_ECC_KEY_CONTEXT)` - but it seems very readable to me. I think we should ensure we are consistent about how we do this in the engine.

PR #5, opened 2021-07-08
Simplify digest memory allocation

samuel-lee-msft:
+ Hopefully handing allocation/freeing of SymCrypt digest state entirely
  to OpenSSL should avoid leaks in HMAC which Ming-Wei observes but I
  have not been able to reproduce
  + Reduces complexity and lines of code anyway, so seems like a step in
    the right direction
+ Free allocated public key on successful ECDH
+ Error in the engine when flen provided to RSA is too big, rather than
  changing the value to read a prefix of the provided buffer
  + Also explicitly write to the amount of memory that is specified in
    the RSA documentation:
    https://www.openssl.org/docs/man1.1.1/man3/RSA_private_decrypt.html

### DIFF
--- a/SymCryptEngine/src/sc_ossl_rsa.c
+++ b/SymCryptEngine/src/sc_ossl_rsa.c
@@ -26,6 +26,11 @@ typedef int (*PFN_RSA_meth_pub_dec)(int flen, const unsigned char* from,
 typedef int (*PFN_RSA_meth_priv_dec)(int flen, const unsigned char *from,
                         unsigned char *to, RSA *rsa, int padding);
 
+// The minimum PKCS1 padding is 11 bytes
+#define SC_OSSL_MIN_PKCS1_PADDING (11)
+// The minimum OAEP padding is 2*hashlen + 2, and the minimum hashlen is SHA1 - with 20B hash => minimum 42B of padding
+#define SC_OSSL_MIN_OAEP_PADDING (42)
+
 int sc_ossl_rsa_pub_enc(int flen, const unsigned char* from,
     unsigned char* to, RSA* rsa,
     int padding)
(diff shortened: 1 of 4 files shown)


SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -51,21 +51,30 @@ int sc_ossl_rsa_pub_enc(int flen, const unsigned char* from,
...
+    if( from == NULL )
+    {
+        goto err;
+    }
 
     switch( padding )
     {
     case RSA_PKCS1_PADDING:
         SC_OSSL_LOG_DEBUG("SymCryptRsaPkcs1Encrypt");
+        if( flen > cbModulus - 11 )
```
mlindgren, line 66: >if( flen > cbModulus - 11 ) [](http://example.com/codeflow?start=8&length=27)

Can you add comments indicating why we have these limits? I assume it has to do with the limit of how many bytes can be encrypted with a given padding scheme, but right now they're just magic numbers.

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -202,7 +223,7 @@ int sc_ossl_rsa_priv_dec(int flen, const unsigned char* from,
                        SYMCRYPT_NUMBER_FORMAT_MSB_FIRST,
                        0,
                        to,
-                       flen > cbModuls ? cbModuls : flen,
+                       cbModulus - 11,
```
mlindgren, line 226: >cbModulus - 11, [](http://example.com/codeflow?start=23&length=15)

Ditto

mlindgren (approved):
:shipit:

SymCryptEngine/src/sc_ossl_rsa.c
samuel-lee-msft: I made some new #defines for these values with some comments

PR #6, opened 2021-08-13
Digests: SAL annotations for parameters

SpencerNofzinger:
Preliminary SAL annotations - want to use this to ask some questions

### DIFF
--- a/SymCryptEngine/src/sc_ossl_digests.c
+++ b/SymCryptEngine/src/sc_ossl_digests.c
@@ -19,10 +19,10 @@ static int sc_ossl_digest_nids[] = {
 };
 
 /* MD5 */
-static int sc_ossl_digest_md5_init(EVP_MD_CTX *ctx);
-static int sc_ossl_digest_md5_update(EVP_MD_CTX *ctx, const void *data, size_t count);
-static int sc_ossl_digest_md5_final(EVP_MD_CTX *ctx, unsigned char *md);
-static int sc_ossl_digest_md5_copy(EVP_MD_CTX *to, const EVP_MD_CTX *from);
+static int sc_ossl_digest_md5_init(_Out_ EVP_MD_CTX *ctx);
+static int sc_ossl_digest_md5_update(_Inout_ EVP_MD_CTX *ctx, _In_reads_(count) const void *data, size_t count);
+static int sc_ossl_digest_md5_final(_Inout_ EVP_MD_CTX *ctx, _Out_writes_(SYMCRYPT_MD5_RESULT_SIZE) unsigned char *md);
+static int sc_ossl_digest_md5_copy(_Out_ EVP_MD_CTX *to, _In_ const EVP_MD_CTX *from);
 static EVP_MD *_hidden_md5_md = NULL;
 static const EVP_MD *sc_ossl_digest_md5(void)
 {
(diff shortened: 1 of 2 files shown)


SymCryptEngine/src/sc_ossl_digests.c
```diff
@@ -19,10 +19,10 @@ static int sc_ossl_digest_nids[] = {
 };
 
 /* MD5 */
-static int sc_ossl_digest_md5_init(EVP_MD_CTX *ctx);
-static int sc_ossl_digest_md5_update(EVP_MD_CTX *ctx, const void *data, size_t count);
-static int sc_ossl_digest_md5_final(EVP_MD_CTX *ctx, unsigned char *md);
-static int sc_ossl_digest_md5_copy(EVP_MD_CTX *to, const EVP_MD_CTX *from);
+static int sc_ossl_digest_md5_init(_Out_ EVP_MD_CTX *ctx);
+static int sc_ossl_digest_md5_update(_Inout_ EVP_MD_CTX *ctx, _In_reads_(count) const void *data, size_t count);
```
NielsFerguson, line 23: The _In_reads_(count) doesn't work if the data type pointed to is void. You're annotating that you read count values of type void. You should use _In_reads_bytes_ I think.

SymCryptEngine/src/sc_ossl_digests.c
```diff
@@ -142,10 +142,10 @@ static const EVP_MD *sc_ossl_digest_sha384(void)
 }
 
 /* SHA512 */
-static int sc_ossl_digest_sha512_init(EVP_MD_CTX *ctx);
-static int sc_ossl_digest_sha512_update(EVP_MD_CTX *ctx, const void *data, size_t count);
-static int sc_ossl_digest_sha512_final(EVP_MD_CTX *ctx, unsigned char *md);
-static int sc_ossl_digest_sha512_copy(EVP_MD_CTX *to, const EVP_MD_CTX *from);
+static int sc_ossl_digest_sha512_init(_Out_ EVP_MD_CTX *ctx);
+static int sc_ossl_digest_sha512_update(_Inout_ EVP_MD_CTX *ctx, _In_reads_(count) const void *data, size_t count);
+static int sc_ossl_digest_sha512_final(_Inout_ EVP_MD_CTX *ctx, _Out_writes_(SYMCRYPT_SHA256_RESULT_SIZE) unsigned char *md);
```
samuel-lee-msft, line 147: >SYMCRYPT_SHA256_RESULT_SIZE [](http://example.com/codeflow?start=77&length=27)

Should be SYMCRYPT_SHA512_RESULT_SIZE

samuel-lee-msft:
Other than moving to _In_reads_bytes_ and the one SHA512 typo, I think this seems good. :)

I like the comment on sc_ossl_digests, and I think the other digest functions are kind of self explanatory.

One possible additional improvement would be to typedef a return value (`SCOSSL_STATUS`?) for the digest functions with a `_Return_type_success_(return == 1)` annotation (see https://docs.microsoft.com/en-us/cpp/code-quality/annotating-function-behavior?view=msvc-160#successfailure-annotations), to make it more explicit that the functions return 1 on success and 0 on failure.

I think such a typedef could be used in a number of places in the engine, and help to make it more explicit when some functions stray from the 1 for success, 0 for failure shape.

SpencerNofzinger:
(It seems we can't reply directly to comments? Boo)
Sam, the SCOSSL_STATUS is a good idea. I've went ahead and done that. Please take one more look next week. Meanwhile, I'll start working on other files and addressing anything else I see in this digest file.

samuel-lee-msft:
LGTM 👍

PR #9, opened 2021-09-03
Code quality: RSA and rand

SpencerNofzinger:
Code quality improvements for RSA and rand, mostly SAL, debug removal, and comments.

### DIFF
--- a/SymCryptEngine/src/sc_ossl_rand.c
+++ b/SymCryptEngine/src/sc_ossl_rand.c
@@ -19,21 +19,21 @@ extern "C" {
 #endif
 
 // RAND_seed() returns 1 on success, 0 otherwise. Since an error internally is fatal, we always return 1 here.
-static int sc_ossl_rand_seed(const void *buf, int num)
+static SCOSSL_STATUS sc_ossl_rand_seed(const void *buf, int num)
 {
     SymCryptProvideEntropy(buf, num);
     return 1;
 }
 
 // RAND_bytes() returns 1 on success, 0 otherwise. Since an error internally is fatal, we always return 1 here.
-static int sc_ossl_rand_bytes(unsigned char *buf, int num)
+static SCOSSL_STATUS sc_ossl_rand_bytes(unsigned char *buf, int num)
 {
     SymCryptRandom(buf, num);
     return 1;
 }
 
 // RAND_add() returns 1 on success, 0 otherwise. Since an error internally is fatal, we always return 1 here.
-static int sc_ossl_rand_add(const void *buf, int num, double randomness)
+static SCOSSL_STATUS sc_ossl_rand_add(const void *buf, int num, double randomness)
 {
     SymCryptProvideEntropy(buf, num);
     return 1;
@@ -41,14 +41,14 @@ static int sc_ossl_rand_add(const void *buf, int num, double randomness)
 
 // RAND_pseudo_bytes() returns 1 if the bytes generated are cryptographically strong, 0 otherwise.
 // Since an error internally is fatal, we always return 1 here.
-static int sc_ossl_rand_pseudorand(unsigned char *buf, int num)
+static SCOSSL_STATUS sc_ossl_rand_pseudorand(unsigned char *buf, int num)
 {
     SymCryptRandom(buf, num);
     return 1;
 }
 
 // RAND_status() returns 1 if the PRNG has been seeded with enough data, 0 otherwise. Since we guarantee this, we return 1.
-static int sc_ossl_rand_status(void)
+static SCOSSL_STATUS sc_ossl_rand_status(void)
 {
     return 1;
 }
--- a/SymCryptEngine/src/sc_ossl_rand.c
+++ b/SymCryptEngine/src/sc_ossl_rand.c
@@ -19,21 +19,21 @@ extern "C" {
 #endif
 
 // RAND_seed() returns 1 on success, 0 otherwise. Since an error internally is fatal, we always return 1 here.
-static SCOSSL_STATUS sc_ossl_rand_seed(const void *buf, int num)
+static SCOSSL_STATUS sc_ossl_rand_seed(_In_reads_bytes_(num) const void *buf, _In_ int num)
 {
     SymCryptProvideEntropy(buf, num);
     return 1;
 }
 
 // RAND_bytes() returns 1 on success, 0 otherwise. Since an error internally is fatal, we always return 1 here.
-static SCOSSL_STATUS sc_ossl_rand_bytes(unsigned char *buf, int num)
+static SCOSSL_STATUS sc_ossl_rand_bytes(_Out_writes_bytes_(num) unsigned char *buf, _In_ int num)
 {
     SymCryptRandom(buf, num);
     return 1;
 }
 
 // RAND_add() returns 1 on success, 0 otherwise. Since an error internally is fatal, we always return 1 here.
-static SCOSSL_STATUS sc_ossl_rand_add(const void *buf, int num, double randomness)
+static SCOSSL_STATUS sc_ossl_rand_add(_In_reads_bytes_(num) const void *buf, _In_ int num, _In_ double randomness)
 {
     SymCryptProvideEntropy(buf, num);
     return 1;
@@ -41,7 +41,7 @@ static SCOSSL_STATUS sc_ossl_rand_add(const void *buf, int num, double randomnes
 
 // RAND_pseudo_bytes() returns 1 if the bytes generated are cryptographically strong, 0 otherwise.
 // Since an error internally is fatal, we always return 1 here.
-static SCOSSL_STATUS sc_ossl_rand_pseudorand(unsigned char *buf, int num)
+static SCOSSL_STATUS sc_ossl_rand_pseudorand(_Out_writes_bytes_(num) unsigned char *buf, _In_ int num)
 {
     SymCryptRandom(buf, num);
     return 1;
--- a/SymCryptEngine/src/sc_ossl_rsa.c
+++ b/SymCryptEngine/src/sc_ossl_rsa.c
@@ -46,6 +46,7 @@ SCOSSL_RETURNLENGTH sc_ossl_rsa_pub_enc(int flen, _In_reads_bytes_(flen) const u
     SYMCRYPT_ERROR SymError = SYMCRYPT_NO_ERROR;
     BN_ULONG cbModulus = 0;
     SIZE_T cbResult = -1;
+    int res = -1;
     const RSA_METHOD *ossl_rsa_meth = NULL;
     PFN_RSA_meth_pub_enc pfn_rsa_meth_pub_enc = NULL;
     SC_OSSL_RSA_KEY_CONTEXT *keyCtx = RSA_get_ex_data(rsa, rsa_sc_ossl_idx);
@@ -173,8 +174,10 @@ SCOSSL_RETURNLENGTH sc_ossl_rsa_pub_enc(int flen, _In_reads_bytes_(flen) const u
         break;
     }
 
+    res = (cbResult <= INT_MAX) ? cbResult : -1;
+
 cleanup:
-    return cbResult;
+    return res;
 }
 
 SCOSSL_RETURNLENGTH sc_ossl_rsa_priv_dec(int flen, _In_reads_bytes_(flen) const unsigned char* from,
@@ -183,6 +186,7 @@ SCOSSL_RETURNLENGTH sc_ossl_rsa_priv_dec(int flen, _In_reads_bytes_(flen) const 
     SYMCRYPT_ERROR SymError = SYMCRYPT_NO_ERROR;
     BN_ULONG cbModulus = 0;
     SIZE_T cbResult = -1;
+    int res = -1;
     const RSA_METHOD *ossl_rsa_meth = NULL;
     PFN_RSA_meth_priv_dec pfn_rsa_meth_priv_dec = NULL;
     SC_OSSL_RSA_KEY_CONTEXT *keyCtx = RSA_get_ex_data(rsa, rsa_sc_ossl_idx);
@@ -302,8 +306,10 @@ SCOSSL_RETURNLENGTH sc_ossl_rsa_priv_dec(int flen, _In_reads_bytes_(flen) const 
         break;
     }
 
+    res = (cbResult <= INT_MAX) ? cbResult : -1;
+
 cleanup:
-    return cbResult;
+    return res;
 }
 
 SCOSSL_RETURNLENGTH sc_ossl_rsa_priv_enc(int flen, _In_reads_bytes_(flen) const unsigned char* from,
@@ -363,10 +369,6 @@ SCOSSL_STATUS sc_ossl_rsa_sign(int type, _In_reads_bytes_(m_length) const unsign
     }
 
     cbModulus = SymCryptRsakeySizeofModulus(keyCtx->key);
-    if( siglen != NULL )
-    {
-        *siglen = cbModulus;
-    }
     if( sigret == NULL || siglen == NULL )
     {
         goto cleanup;
@@ -391,7 +393,7 @@ SCOSSL_STATUS sc_ossl_rsa_sign(int type, _In_reads_bytes_(m_length) const unsign
                        SYMCRYPT_FLAG_RSA_PKCS1_NO_ASN1,
                        SYMCRYPT_NUMBER_FORMAT_MSB_FIRST,
                        sigret,
-                       siglen != NULL ? (*siglen > cbModulus ? cbModulus : *siglen) : 0,
+                       cbModulus,
                        &cbResult);
 
         if( SymError != SYMCRYPT_NO_ERROR )
@@ -416,7 +418,7 @@ SCOSSL_STATUS sc_ossl_rsa_sign(int type, _In_reads_bytes_(m_length) const unsign
                        0,
                        SYMCRYPT_NUMBER_FORMAT_MSB_FIRST,
                        sigret,
-                       siglen != NULL ? (*siglen > cbModulus ? cbModulus : *siglen) : 0,
+                       cbModulus,
                        &cbResult);
 
         if( SymError != SYMCRYPT_NO_ERROR )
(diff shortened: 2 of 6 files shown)


SymCryptEngine/src/sc_ossl_rand.c
```diff
@@ -19,36 +19,36 @@ extern "C" {
 #endif
 
 // RAND_seed() returns 1 on success, 0 otherwise. Since an error internally is fatal, we always return 1 here.
-static int sc_ossl_rand_seed(const void *buf, int num)
+static SCOSSL_STATUS sc_ossl_rand_seed(_In_reads_bytes_(num) const void *buf, _In_ int num)
```
samuel-lee-msft, line 22: >_In_ [](http://example.com/codeflow?start=78&length=4)

I think generally we don't mark up non-pointer parameters - they are unambiguously only input to a function.

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -988,7 +958,6 @@ int sc_ossl_rsa_keygen(RSA* rsa, int bits, BIGNUM* e,
     SC_OSSL_LOG_BIGNUM_DEBUG("d", rsa_d);
```
samuel-lee-msft, line 958: These should also probably be removed.

We could keep some debug logging if we ensure the debug macros don't generate any code in a release mode build, but I think there is not much benefit to keeping these lines around now we are passing functional tests.

samuel-lee-msft:
>     SC_OSSL_LOG_BYTES_DEBUG("SymCrypt pbPrivateExponent", pbPrivateExponent, cbPrivateExponent);

ditto

---
Refers to: SymCryptEngine/src/sc_ossl_rsa.c:1140 in 1122f0f. [](commit_id = 1122f0f99bd70b4f30cf5ee8c47d783974f3f51d, deletion_comment = False)

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -31,11 +31,19 @@ typedef int (*PFN_RSA_meth_priv_dec)(int flen, const unsigned char *from,
...
+// Encrypts flen bytes at from using public key rsa and stores ciphertext in to.
+// Same parameters as RSA_public_encrypt
+// - flen must not be more than RSA_size(rsa) - 11 for the PKCS #1 v1.5 based padding modes, not more than RSA_size(rsa) - 42
+//   for RSA_PKCS1_OAEP_PADDING and exactly RSA_size(rsa) for RSA_NO_PADDING
+// - from and to may overlap
+// - padding is ones of RSA_PKCS1_PADDING, RSA_PKCS1_OAEP_PADDING, RSA_SSLV23_PADDING, or RSA_NO_PADDING
+// Returns size of encrypted data (RSA_size(rsa)), or -1 on error
+_Success_(return >= 0)
+int sc_ossl_rsa_pub_enc(_In_ int flen, _In_reads_bytes_(flen) const unsigned char* from,
+    _Out_ unsigned char* to, _In_ RSA* rsa,
```
samuel-lee-msft, line 44: >_Out_ [](http://example.com/codeflow?start=4&length=5)

I think SAL tooling might read this as saying to is a pointer to a single character.
I imagine tooling might also barf at _Out_writes_bytes_( RSA_size(rsa) ), but it's a bit more explicit about the function's contract to a human reader

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -378,21 +384,18 @@ int sc_ossl_rsa_sign(int type, const unsigned char* m,
...
-        SC_OSSL_LOG_DEBUG("sigret NOT present");
         goto CommonReturn; // Not error - this can be called with NULL parameter for siglen
     }
 
     switch( type )
     {
     case NID_md5_sha1:
-        SC_OSSL_LOG_DEBUG("NID_md5_sha1");
         SC_OSSL_LOG_INFO("SymCrypt engine warning using Mac algorithm MD5+SHA1 which is not FIPS compliant");
         if( m_length != 36 )
```
samuel-lee-msft, line 400: >36 [](http://example.com/codeflow?start=24&length=2)

Would be good to #define these constants (representing the expected lengths in bytes of various hashes)

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -1296,9 +1267,10 @@ void sc_ossl_rsa_free_key_context(SC_OSSL_RSA_KEY_CONTEXT *keyCtx)
     return;
 }
 
-int sc_ossl_rsa_finish(RSA *rsa)
+// Destroys instance of RSA object. The memory for rsa is not freed by this function.
+// Returns 1 on success
+SCOSSL_STATUS sc_ossl_rsa_finish(_In_ RSA *rsa)
```
samuel-lee-msft, line 1272: >_In_ [](http://example.com/codeflow?start=33&length=4)

_Inout_ ?

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -1000,11 +959,10 @@ int sc_ossl_rsa_keygen(RSA* rsa, int bits, BIGNUM* e,
     goto CommonReturn;
```
SpencerNofzinger, line 959: I'm not really a fan of these err blocks being below the actual return statement - thoughts from others?

samuel-lee-msft: Right - yeah, we want to replace from CommonReturn to something more like:
`
cleanup:
    if( ret == 0 )
    {
        sc_ossl_rsa_free_key_context(keyCtx);
    }
    return ret;
`
Error paths just go to cleanup early, with ret set to 0, the success path hits cleanup with ret set to 1. This is something we want to change across the board.

SymCryptEngine/src/sc_ossl_rsa.h
```diff
@@ -12,42 +13,41 @@ extern "C" {
 
 extern int rsa_sc_ossl_idx;
 
-int sc_ossl_rsa_pub_enc(int flen, const unsigned char* from,
-                         unsigned char* to, RSA* rsa,
-                         int padding);
+_Success_(return >= 0)
```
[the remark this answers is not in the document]
samuel-lee-msft: I think SAL annotations should be in the headers for consistency.
The function purpose comments should also be with the function declarations in the headers - probably they should be removed from the function definitions to avoid duplication.

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -31,11 +31,19 @@ typedef int (*PFN_RSA_meth_priv_dec)(int flen, const unsigned char *from,
...
-    int padding)
+// Public Encryption
+// Encrypts flen bytes at from using public key rsa and stores ciphertext in to.
+// Same parameters as RSA_public_encrypt
+// - flen must not be more than RSA_size(rsa) - 11 for the PKCS #1 v1.5 based padding modes, not more than RSA_size(rsa) - 42
+//   for RSA_PKCS1_OAEP_PADDING and exactly RSA_size(rsa) for RSA_NO_PADDING
+// - from and to may overlap
+// - padding is ones of RSA_PKCS1_PADDING, RSA_PKCS1_OAEP_PADDING, RSA_SSLV23_PADDING, or RSA_NO_PADDING
+// Returns size of encrypted data (RSA_size(rsa)), or -1 on error
+_Success_(return >= 0)
```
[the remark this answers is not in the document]
samuel-lee-msft: I like that idea

SymCryptEngine/src/sc_ossl_rsa.c
samuel-lee-msft: We can't reasonably capture the various possible sizes of to for decrypt in SAL (depending on the padding), but the maximum required size is still RSA_size(rsa). I think it is reasonable to use the maximum size in the SAL annotation?

SymCryptEngine/src/sc_ossl_rsa.c
SpencerNofzinger: Changed. Please review this since the logic is a little messy in places.

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -543,56 +546,42 @@ int sc_ossl_rsa_sign(int type, const unsigned char* m,
...
-        //     goto err;
-        // }
-        // return pfn_rsa_meth_sign(type, m, m_length, sigret, siglen, rsa);
-        goto err;
+        cbResult = 0;
+        goto cleanup;
     }
 
-CommonReturn:
+    cbResult = 1;
```
SpencerNofzinger, line 559: This function never set result to 1 on success... not sure how it was passing before

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -366,38 +367,35 @@ int sc_ossl_rsa_sign(int type, const unsigned char* m,
...
     {
         if( sc_ossl_initialize_rsa_key((RSA *)rsa, keyCtx) == 0 )
         {
-            goto err;
+            goto cleanup;
         }
     }
 
     cbModulus = SymCryptRsakeySizeofModulus(keyCtx->key);
     cbResult = cbModulus;
```
SpencerNofzinger, line 381: Not sure why we set this here, it complicates a lot of the error logic into setting cbResult back to 0, and we never return a size in this function

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -366,38 +367,35 @@ int sc_ossl_rsa_sign(int type, const unsigned char* m,
...
-    SC_OSSL_LOG_DEBUG("m_length= %d", m_length);
     if( siglen != NULL )
     {
         *siglen = cbResult;
     }
     if( sigret == NULL )
     {
-        SC_OSSL_LOG_DEBUG("sigret NOT present");
-        goto CommonReturn; // Not error - this can be called with NULL parameter for siglen
+        goto cleanup; // Not error - this can be called with NULL parameter for siglen
```
SpencerNofzinger, line 388: Is this true? The documentation doesn't say anything about returning a length when a NULL parameter is passed in.
https://www.openssl.org/docs/man1.1.1/man3/RSA_sign.html

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -601,12 +590,12 @@ int sc_ossl_rsa_verify(int dtype, const unsigned char* m,
     switch( dtype )
```
SpencerNofzinger, line 590: Not sure why we set cbResult = cbModulus here, it complicates a lot of the error logic into setting cbResult back to 0, and we never return a size in this function

SymCryptEngine/src/sc_ossl_rsa.c
samuel-lee-msft: On looking at this again, for the `RSA_sign` API we are handling here (which only supports PKCS1 signatures) we should not accept NULL sigret.

I got confused because for RSAPSS signing (sc_ossl_rsapss.c) we are instead intercepting https://www.openssl.org/docs/man1.1.1/man3/EVP_PKEY_sign.html which _does_ support being called with a null `sig`, in which case the `siglen` is set and success returned.

It looks like in both cases I had thought the _return_ value was `SCOSSL_RETURNLENGTH` rather than `SCOSSL_STATUS` (we set cbResult to the size of the result and then return it on success).

SymCryptEngine/src/sc_ossl_rsa.c
samuel-lee-msft: Looks like I incorrectly thought this function had `SCOSSL_RETURNLENGTH` return value. I think probably we should return another variable (initially zero and set to 1 only on success), and only set `cbResult` in calls to `SymCryptRsaPkcs1Sign`.

I think we should also set `siglen` _after_ successfully producing the signature, rather than always setting it to `cbModulus`. (In RSAPSS we would set `siglen` to `cbModulus` when `sig` is NULL, or to `cbResult` when `sig` is not NULL and after successfully producing the signature)

SymCryptEngine/src/sc_ossl_rsa.c
samuel-lee-msft: Good catch - I think probably in unit tests callers may only be checking for non-zero => success if they expect only 1 or 0 to be returned. To repeat another comment, I think we should return a different variable which is initially 0 and only set to 1 on success.

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -46,31 +53,32 @@ int sc_ossl_rsa_pub_enc(int flen, const unsigned char* from,
     if( keyCtx == NULL )
     {
         SC_OSSL_LOG_ERROR("SymCrypt Context Not Found.");
-        goto err;
+        goto cleanup;
```
samuel-lee-msft, line 56: >goto cleanup; [](http://example.com/codeflow?start=8&length=13)

missing cbResult = -1

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -31,11 +31,18 @@ typedef int (*PFN_RSA_meth_priv_dec)(int flen, const unsigned char *from,
...
+#define SC_OSSL_SHA512_DIGEST_LENGTH (64)
+
+SCOSSL_RETURNLENGTH sc_ossl_rsa_pub_enc(int flen, _In_reads_bytes_(flen) const unsigned char* from,
+    _Out_writes_bytes_(RSA_size(rsa)) unsigned char* to, _In_ RSA* rsa,
     int padding)
 {
-    SC_OSSL_LOG_DEBUG(NULL);
     SYMCRYPT_ERROR SymError = SYMCRYPT_NO_ERROR;
     BN_ULONG cbModulus = 0;
     BN_ULONG cbResult = 0;
```
samuel-lee-msft, line 48: >BN_ULONG cbResult = 0; [](http://example.com/codeflow?start=4&length=22)

There's an argument to set cbResult to -1 initially (assume error result).

It would remove a bunch of lines below, but we would need to be careful to set it to -1 again if `SymCryptRsa*` returns an error.

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -352,12 +355,10 @@ int sc_ossl_rsa_pub_dec(int flen, const unsigned char* from,
...
-    unsigned int m_length,
-    unsigned char* sigret, unsigned int* siglen,
-    const RSA* rsa)
+SCOSSL_STATUS sc_ossl_rsa_sign(int type, _In_reads_bytes_(m_length) const unsigned char* m, unsigned int m_length,
+    _Out_writes_bytes_(siglen) unsigned char* sigret, _Out_ unsigned int* siglen,
+    _In_ const RSA* rsa)
 {
-    SC_OSSL_LOG_DEBUG(NULL);
     BN_ULONG cbModulus = 0;
     size_t cbResult = 0;
```
samuel-lee-msft, line 363: >size_t [](http://example.com/codeflow?start=4&length=6)

Hmm, we should either have all `cbResult` variables be `SIZE_T` (probably better) or `BN_ULONG` - we currently have a mix.

SymCryptEngine/src/sc_ossl_rsa.c
samuel-lee-msft: Slight amendment, my previous comment was about the `sc_ossl_rsa_sign`;
here in `sc_ossl_rsa_verify` I think we can remove `cbResult` entirely and return a SCOSSL_STATUS (called `scosslStatus`?) variable.

SpencerNofzinger: Thanks, that makes sense. Alright, changing it to ret instead.

SpencerNofzinger: In regards to siglen after producing signature, its value is used in the SymCryptRsaPkcs1Sign functions. We could change that to just always be cbModulus, but I'll let you speak on the implications of that. For now I'm leaving siglen to cbModulus for this iteration.

SymCryptEngine/src/sc_ossl_rsa.c
SpencerNofzinger: Thanks, makes sense.

SymCryptEngine/src/sc_ossl_rsa.c
SpencerNofzinger: Thanks, I agree.

SpencerNofzinger: Additionally, these functions are supposed to return an int, and both proposed types for cbResult (BN_ULONG and size_t) are unsigned. Are there any concerns with this overflowing?

SpencerNofzinger: I can set the size_t to -1, and I think the conversion from unsigned to signed will preserve the -1

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -352,52 +339,47 @@ int sc_ossl_rsa_pub_dec(int flen, const unsigned char* from,
...
         {
-            goto err;
+            goto cleanup;
         }
     }
 
     cbModulus = SymCryptRsakeySizeofModulus(keyCtx->key);
-    cbResult = cbModulus;
-    SC_OSSL_LOG_DEBUG("m_length= %d", m_length);
     if( siglen != NULL )
```
samuel-lee-msft, line 366: I think we should remove this if, and `*siglen` should be set to `cbResult` before returning success.
The calls to `SymCryptRsaPkcs1Sign` can just specify `cbModulus` as the `cbSignature` (the penultimate) parameter.

SymCryptEngine/src/sc_ossl_rsa.c
samuel-lee-msft: Yeah, I just made a comment to that effect - the value passed to SymCryptRsaPkcs1Sign is already always going to be cbModulus, we are just doing confused logic to get there (probably due to similar but different logic being required for RSAPSS)

samuel-lee-msft (commented):
Awesome, I think this is looking very close to ready to merge - thanks for cleaning this up! :)
Just 1 outstanding comment, and I'll also double check unit tests on my side before approving.

SymCryptEngine/src/sc_ossl_rsa.c
samuel-lee-msft: Setting the unsigned value to -1 is always OK (will set the unsigned value to UINT_MAX)

The conversion back to int looks like it's actually implementation defined, so we should probably be safe and explicitly convert cbResult to an int res (i.e. by setting res to cbResult iff it's <= INT_MAX, and -1 otherwise).

samuel-lee-msft:
> Awesome, I think this is looking very close to ready to merge - thanks for cleaning this up! :)
> Just 1 outstanding comment, and I'll also double check unit tests on my side before approving.

I just ran unit tests - looks good to merge

samuel-lee-msft (approved):
I just ran unit tests, and had a look at latest. Looks good to merge from my side.

PR #10, opened 2021-09-07
Code quality: DH and DSA

SpencerNofzinger:


samuel-lee-msft:
LGTM - just may want to move the function comments into the headers

SpencerNofzinger:
> LGTM - just may want to move the function comments into the headers

Did so and used the new type SCOSSL_RETURNLENGTH from the other PR

samuel-lee-msft (approved):

PR #12, opened 2021-09-22
Code quality: ECC

SpencerNofzinger:


SpencerNofzinger:
Not done with the whole ecc source file but starting the review anyway

SymCryptEngine/src/sc_ossl_ecc.c
```diff
@@ -891,7 +853,7 @@ ECDSA_SIG* sc_ossl_eckey_sign_sig(const unsigned char* dgst, int dgst_len,
         }
         if( s != NULL )
         {
-            BN_free(r);
+            BN_free(s);
```
samuel-lee-msft, line 856: Good catch!

SymCryptEngine/src/sc_ossl_ecc.c
```diff
@@ -1036,7 +1036,7 @@ SCOSSL_RETURNLENGTH sc_ossl_eckey_compute_key(_Out_writes_bytes_(pseclen) unsign
 
     int res = -1; // fail
 
-    switch( sc_ossl_get_context(ecdh, &keyCtx) )
+    switch( sc_ossl_get_context((EC_KEY*)ecdh, &keyCtx) )
```
SpencerNofzinger, line 1052: I wanted to fix the the compiler warning as it was irritating - however not sure the best way to handle this. This get context function does a couple different things, and in certain paths may alter the ecdh parameter, but in this path it should not, so I think removing the const is fine. Thoughts?

SymCryptEngine/src/sc_ossl_ecc.h
```diff
@@ -17,33 +18,67 @@ typedef int (*PFN_eckey_set_private)(EC_KEY *key, const BIGNUM *priv_key);
...
 
 // int sc_ossl_eckey_init(EC_KEY *key);
-void sc_ossl_eckey_finish(EC_KEY *key);
+
+// Frees SymCrypt-specific components of key
+void sc_ossl_eckey_finish(_Inout_ EC_KEY *key);
+
 // int sc_ossl_eckey_copy(EC_KEY *dst, const EC_KEY *src);
 // int sc_ossl_eckey_set_group(EC_KEY *key, const EC_GROUP *grp);
 // int sc_ossl_eckey_set_private(EC_KEY *key, const BIGNUM *priv_key);
```
SpencerNofzinger, line 27: Do we want to keep these or remove them?

SpencerNofzinger: (The commented out functions)

samuel-lee-msft: I think we can remove them

SymCryptEngine/src/sc_ossl_ecc.c
samuel-lee-msft: It would probably be better to refactor `sc_ossl_get_context` so we can call a version with a const EC_KEY and fail if EC_KEY_get_ex_data returns NULL, rather than allocating and setting the ex_data in this case. I think we may still get errors in the Sanitize build with a cast?

For now I think it's fine to do the cast here saying we shouldn't hit the paths where we modify `ecdh` from here, with a TODO comment to refactor if you want to prioritize other stuff.

SpencerNofzinger: Agreed

SymCryptEngine/src/sc_ossl_ecc.h
```diff
@@ -16,34 +17,61 @@ typedef int (*PFN_eckey_set_group)(EC_KEY *key, const EC_GROUP *grp);
...
+
+// Generates a new public and private key for the supplied key object.
+// key must have an EC_GROUP object associated with it before calling this function.
+// Returns 1 on success or 0 on error.
+SCOSSL_STATUS sc_ossl_eckey_keygen(_Inout_ EC_KEY *key);
+
+// Computes shared secret psec and secret length pseclen using pub_key and ecdh.
+// Allocates psec on success.
+// Returns length of secret on success, or -1 on error.
+SCOSSL_RETURNLENGTH sc_ossl_eckey_compute_key(_Out_writes_bytes_(pseclen) unsigned char **psec,
```
samuel-lee-msft, line 31: >_Out_writes_bytes_(pseclen) unsigned char **psec, [](http://example.com/codeflow?start=46&length=49)

Sorry, I should have picked up on this earlier, but I think the correct annotation here is:
_Outptr_result_bytebuffer_(*pseclen)
?

The function allocates, rather than taking and updating an allocated buffer.

SymCryptEngine/src/sc_ossl_ecc.h
```diff
@@ -16,34 +17,61 @@ typedef int (*PFN_eckey_set_group)(EC_KEY *key, const EC_GROUP *grp);
...
-                        EC_KEY* eckey);
-int sc_ossl_eckey_sign_setup(EC_KEY* eckey, BN_CTX* ctx_in, BIGNUM** kinvp, BIGNUM** rp);
-ECDSA_SIG* sc_ossl_eckey_sign_sig(const unsigned char* dgst, int dgst_len,
-                                   const BIGNUM* in_kinv, const BIGNUM* in_r,
-                                   EC_KEY* eckey);
-int sc_ossl_eckey_verify(int type, const unsigned char* dgst, int dgst_len,
-                          const unsigned char* sigbuf, int sig_len, EC_KEY* eckey);
-int sc_ossl_eckey_verify_sig(const unsigned char* dgst, int dgst_len,
-                              const ECDSA_SIG* sig, EC_KEY* eckey);
+                        _Out_writes_bytes_(siglen) unsigned char* sig,
```
samuel-lee-msft, line 43: >siglen [](http://example.com/codeflow?start=43&length=6)

I think this should be *siglen

samuel-lee-msft (approved):
I think this is good to merge other than for the SAL annotations mentioned.

SymCryptEngine/src/sc_ossl_ecc.h
SpencerNofzinger: It seems our SAL definitions for compilation are coming from symcrypt_no_sal.h, which only includes certain SAL annotations at this point, so I get an error when using _Outptr_result_bytebuffer_. Leaving as-is for now, but we can discuss whether we want to just ad-hoc add more SAL annotations to that header when needed or if there's a better way.

PR #14, opened 2021-09-23
Code quality: AES

SpencerNofzinger:


SymCryptEngine/src/sc_ossl_ciphers.c
```diff
@@ -784,29 +752,28 @@ static int sc_ossl_aes_gcm_tls(struct cipher_gcm_ctx *cipherCtx, unsigned char *
...
     return ret;
-err:
-    ret = -1;
-    OPENSSL_cleanse(out, inl);
-    goto end;
 }
 
 // This is a EVP_CIPH_FLAG_CUSTOM_CIPHER do cipher method
 // return negative value on failure, and number of bytes written to out on success (may be 0)
 int sc_ossl_aes_gcm_cipher(EVP_CIPHER_CTX *ctx, unsigned char *out,
```
SpencerNofzinger, line 772: This one is a bit weird - claims to return a length for this do_cipher function, while the other ones just return 1 for success. Unclear which is correct, or if they need to be the same. I'll investigate more tomorrow

samuel-lee-msft: I think the OpenSSL documentation is a little sparse around `EVP_CIPH_FLAG_CUSTOM_CIPHER`, but my recollection is that the cipher function is called in a different way, expecting the return value to indicate the bytes written. I think I had to look at the [OpenSSL 3.0.0 source](https://github.com/openssl/openssl/blob/0f70d6013435308ada5d0eb662b31f370b07ebd7/crypto/evp/evp_enc.c#L514-L527) to figure this out.

SpencerNofzinger: Thanks, that makes a lot more sense.

SymCryptEngine/src/sc_ossl_ciphers.c
```diff
@@ -403,45 +390,40 @@ int sc_ossl_ciphers(ENGINE *e, const EVP_CIPHER **cipher,
...
             return 0;
         }
-        SC_OSSL_LOG_BYTES_DEBUG("key", key, EVP_CIPHER_CTX_key_length(ctx));
     }
     return 1;
 }
 
-int sc_ossl_aes_cbc_cipher(EVP_CIPHER_CTX *ctx, unsigned char *out,
-                               const unsigned char *in, size_t inl)
+// Encrypts or ecrypts in, storing result in out, depending on mode set in ctx.
```
mlindgren, line 419: >ecrypts [](http://example.com/codeflow?start=15&length=7)

decrypts?

SymCryptEngine/src/sc_ossl_ciphers.c
```diff
@@ -403,45 +390,40 @@ int sc_ossl_ciphers(ENGINE *e, const EVP_CIPHER **cipher,
...
         memcpy(ctx_iv, iv, EVP_CIPHER_CTX_iv_length(ctx));
-        SC_OSSL_LOG_BYTES_DEBUG("ctx_iv", ctx_iv, EVP_CIPHER_CTX_iv_length(ctx));
     }
     if( key )
     {
         SymError = SymCryptAesExpandKey(&cipherCtx->key, key, EVP_CIPHER_CTX_key_length(ctx));
         if( SymError != SYMCRYPT_NO_ERROR )
         {
-            SC_OSSL_LOG_SYMERROR_DEBUG("SymCryptAesExpandKey failed", SymError);
             return 0;
```
mlindgren, line 413: >0; [](http://example.com/codeflow?start=19&length=2)

Do we have symbolic values for these error codes? Since we've defined a status type?

SymCryptEngine/src/sc_ossl_ciphers.c
```diff
@@ -486,39 +465,34 @@ static int sc_ossl_aes_cbc_ctrl(EVP_CIPHER_CTX *ctx, int type, int arg,
...
  * AES-ECB Implementation
  */
-int sc_ossl_aes_ecb_init_key(EVP_CIPHER_CTX *ctx, const unsigned char *key,
-                             const unsigned char *iv, int enc)
+
+// Initializes ctx with the provided key and iv, along with enc/dec mode.
+// enc should be set to 1 for encryption, 0 for decryption, and -1 to leave value unchanged.
+// Returns 1 on success, or 0 on error.
+SCOSSL_STATUS sc_ossl_aes_ecb_init_key(_Inout_ EVP_CIPHER_CTX *ctx, _In_ const unsigned char *key,
+                             _In_ const unsigned char *iv, int enc)
```
mlindgren, line 473: >int enc [](http://example.com/codeflow?start=59&length=7)

Might also be nice to have a typedef for this with symbolic values for encrypt, decrypt, and leave unchanged. I assume we're implementing an open SSL interface here, but it should still be OK to typedef this if we typedef it to int.

SymCryptEngine/src/sc_ossl_ciphers.c
```diff
@@ -575,29 +550,26 @@ int sc_ossl_aes_xts_init_key(EVP_CIPHER_CTX *ctx, const unsigned char *key,
...
-            SC_OSSL_LOG_SYMERROR_DEBUG("SymCryptXtsAesExpandKey failed", SymError);
             return 0;
         }
     }
     return 1;
 }
 
-int sc_ossl_aes_xts_cipher(EVP_CIPHER_CTX *ctx, unsigned char *out,
-                               const unsigned char *in, size_t inl)
+// Encrypts or ecrypts in, storing result in out, depending on mode set in ctx.
```
mlindgren, line 566: >ecrypts [](http://example.com/codeflow?start=15&length=7)

Decrypts

mlindgren (approved):
:shipit:

SymCryptEngine/src/sc_ossl_ciphers.c
SpencerNofzinger: We don't, and it's a little tricky too since a general error is 0 for some and -1 for others. We can talk about adding this to quality work after I get back.

PR #17, opened 2021-09-24
Add AES-CCM support to the engine

samuel-lee-msft:
+ Mostly copying GCM, but with some differences for the different way it
  is invoked by callers
+ Also update GCM to use EVP constants rather than locally defined
  versions to improve clarity a bit


samuel-lee-msft:
@SpencerNofzinger sorry if this conflicts with your updates for code quality - I had wanted to get this done before you got to the cipher code but didn't quite make it!

samuel-lee-msft:
Hmm, on trying this against OpenSSL speed it turns out that I need to rejig the code to support all existing callers - I'll close this PR for now and come back to it after vacation.

PR #20, opened 2021-11-15
Remove RSA PKCS1 side-channel

samuel-lee-msft:


SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -227,13 +225,21 @@ SCOSSL_RETURNLENGTH sc_ossl_rsa_priv_dec(int flen, _In_reads_bytes_(flen) const
...
-        }
-        break;
+
+        // Constant-time error processing to avoid Bleichenbacher attack
+
+        // Set ret based on SymError and cbResult
+        // cbResult > INT_MAX               => err > 0
+        err = (UINT64)cbResult >> 31;
+        // SymError != SYMCRYPT_NO_ERROR    => err > 0
+        err |= (UINT32)(SymError ^ SYMCRYPT_NO_ERROR);
```
mlindgren, line 235: >err |= (UINT32)(SymError ^ SYMCRYPT_NO_ERROR); [](http://example.com/codeflow?start=8&length=46)

Since `SYMCRYPT_NO_ERROR == 0`, isn't this line the same as `err |= (UINT32) SymError`? Am I misunderstanding something, or is it written this way in case `SYMCRYPT_NO_ERROR` is redefined to be non-zero?

mlindgren (approved):
:shipit:

SymCryptEngine/src/sc_ossl_rsa.c
samuel-lee-msft: Right - the compiler can recognize `SYMCRYPT_NO_ERROR` is 0 if it is, but as a caller of SymCrypt we shouldn't assume `SYMCRYPT_NO_ERROR` is 0

PR #23, opened 2021-11-23
Add DH support

samuel-lee-msft:
+ Rename similar ECC functions to have _ecc in the name
+ Remove needless long-lived buffer in SC_OSSL_ECC_KEY_CONTEXT
+ Avoid redundant checks using BN_num_bytes preceding BN_bn2binpad calls
+ Add ECDH and DH tests to SslPlay


SymCryptEngine/src/sc_ossl_ecc.c
```diff
@@ -515,13 +512,7 @@ SCOSSL_STATUS sc_ossl_import_keypair(_In_ const EC_KEY* eckey, _In_ const EC_GRO
         goto cleanup;
     }
 
-    if( (ec_privkey != NULL) && (BN_num_bytes(ec_privkey) > cbPrivateKey) )
```
mlindgren, line 518: >(BN_num_bytes(ec_privkey) > cbPrivateKey) [](http://example.com/codeflow?start=32&length=41)

I'm guessing this can never happen because the curve size defines the key size? Is there any way the OpenSSL structure could have the wrong size?

SymCryptEngine/src/sc_ossl_dh.c
```diff
@@ -16,68 +21,575 @@ typedef int (*PFN_DH_meth_bn_mod_exp)(const DH* dh, BIGNUM* r,
...
+    {
+    case SC_OSSL_DH_GET_CONTEXT_ERROR:
+        SC_OSSL_LOG_ERROR("sc_ossl_get_dh_context_ex failed.");
         return 0;
+    case SC_OSSL_DH_GET_CONTEXT_FALLBACK:
+        ossl_dh_meth = DH_OpenSSL();
+        PFN_DH_meth_generate_key pfn_dh_meth_generate_key = DH_meth_get_generate_key(ossl_dh_meth);
+        if (pfn_dh_meth_generate_key == NULL)
+        {
+            return 0;
```
mlindgren, line 447: >return 0; [](http://example.com/codeflow?start=12&length=9)

Can we create some symbolic error codes for DH errors too?

mlindgren, line 447: >return 0; [](http://example.com/codeflow?start=12&length=9)

Maybe we should have `#define SC_OSSL_DH_ERROR (0)` and `#define SC_OSSL_DH_SUCCESS (1)`  (or just use generic error/success macros, if we have them) so that we don't have to return integer constants.

SslPlay/SslPlay.cpp
```diff
@@ -106,6 +148,136 @@ void TestEcc()
...
+    DH_get0_key(key2, &pubkey2, NULL);
+    if( pubkey1 == NULL || pubkey2 == NULL )
+    {
+        printOpenSSLError("DH_get0_key failed\n");
+        goto end;
+    }
+
+    printf("Command DH_compute_key_padded\n");
+    if( DH_compute_key_padded(sharedSecretBytes1, pubkey2, key1) != DH_size(key1) ||
+        DH_compute_key_padded(sharedSecretBytes2, pubkey1, key2) != DH_size(key2) )
```
mlindgren, line 223: Should we also have tests to compare our result to the upstream OpenSSL implementation? I know that's something we've discussed adding to the SymCrypt unittests, but it might also be good to have in this layer to ensure there is no mistranslation or other such issues.

SslPlay/SslPlay.cpp
```diff
@@ -89,15 +89,57 @@ void TestEcdsa(EC_KEY* key)
...
+    if( ECDH_compute_key(sharedSecretBytes1, SC_OSSL_ECDH_SHARED_SECRET_LEN, publicKey2, key1, NULL) <= 0 )
+    {
+        printOpenSSLError("ECDH_compute_key failed\n");
+        goto end;
+    }
+    
+    printf("Command ECDH_compute_key #2\n");
+    if( ECDH_compute_key(sharedSecretBytes2, SC_OSSL_ECDH_SHARED_SECRET_LEN, publicKey1, key2, NULL) <= 0 )
+    {
+        printOpenSSLError("ECDH_compute_key failed\n");
```
mlindgren, line 112: I didn't realize this before, but the tests in this file just print an error and continue if something fails. That makes it very easy to miss a failure, and harder to detect failures with automated tooling. Not blocking for this PR, but I would prefer to have the tests either exit with an error code or just crash on failure.

mlindgren (approved):
:shipit:

SymCryptEngine/src/sc_ossl_ecc.c
samuel-lee-msft: Actually we are moving to relying on the return value from `BN_bn2binpad` below. In the old code we weren't checking the return from `BN_bn2binpad` which could fail because `BN_num_bytes(ec_privkey) > cbPrivateKey` or for some other reason.

SymCryptEngine/src/sc_ossl_dh.c
samuel-lee-msft: Hmm, that is a fair comment - OpenSSL code is littered with these integer constant return values, we made a step in the right direction by marking up the return value as SCOSSL_STATUS (which means 1 is success, failure (normally 0) otherwise).

We should probably introduce an enum something like SYMCRYPT_ERROR which gives names to these return values and apply them across the board. I can draft this is another PR along with sc_ossl -> scossl renaming.

SslPlay/SslPlay.cpp
samuel-lee-msft: Yeah, I agree with this - would be good to have the test crash on failure. I created an issue: [https://github.com/microsoft/SymCrypt-OpenSSL/issues/27](https://github.com/microsoft/SymCrypt-OpenSSL/issues/27)

SymCryptEngine/src/sc_ossl_dh.c
samuel-lee-msft: Created another issue for this: [https://github.com/microsoft/SymCrypt-OpenSSL/issues/29](https://github.com/microsoft/SymCrypt-OpenSSL/issues/29)

PR #25, opened 2021-11-29
Use SymCryptDlkeySetPrivateKeyLength in DH

samuel-lee-msft:


SymCryptEngine/src/sc_ossl_dh.c
```diff
@@ -63,6 +63,7 @@ SCOSSL_STATUS scossl_dh_generate_keypair(
     SIZE_T cbPrivateKey = 0;
     PBYTE  pbPublicKey = NULL;
     SIZE_T cbPublicKey = 0;
+    UINT32 nBitsPriv;
```
mlindgren, line 66: >UINT32 nBitsPriv; [](http://example.com/codeflow?start=4&length=17)

Nit: initialize? Same below

mlindgren (approved):
:shipit:

PR #26, opened 2021-12-07
Address various issues found in Code Review

samuel-lee-msft:
+ Enable compiler warnings for C source files
+ Fix various compiler warnings (unsigned/signed comparison, unused
  variables, etc.)
+ Remove redundant information from SCOSSL data structures
+ Remove redundant includes
+ Initialize long-lived static variables once at Engine load time,
  rather than on demand to avoid race conditions when algorithms are
  called concurrently without needing to introduce locking.
+ Check various returns for OpenSSL functions (e.g. BN_bin2bn et. al)
  and gracefully error rather than potentially failing in unexpected
  ways in low memory conditions
+ Fix memory leaks in new FFC DH implementation
+ Ensure all externally visible symbols have a scossl or sc_ossl prefix
  (should settle on a single prefix in a followup PR)
+ Fallback to OpenSSL for ECDSA signatures with a specific kinv or r
+ Use stderr rather than stdout for error messages in debug build
+ Correctly set verify_init function in PKEY rsa method!
+ Only log error messages on signature verification error, rather than
  expected signature verification failure
+ Remove a lot of useless work in RSA key setup
+ Error in RSA key import/generation if the requested public exponent is
  larger than 2^64-1

### DIFF
--- a/SymCryptEngine/src/sc_ossl_ciphers.c
+++ b/SymCryptEngine/src/sc_ossl_ciphers.c
@@ -3,8 +3,6 @@
 //
 
 #include "sc_ossl_ciphers.h"
-#include "sc_ossl_helpers.h"
-#include <symcrypt.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -16,17 +14,14 @@ typedef int SC_OSSL_ENCRYPTION_MODE;
 #define SC_OSSL_ENCRYPTION_MODE_NOCHANGE (-1)
 
 struct cipher_cbc_ctx {
-    INT32 enc;                     /* COP_ENCRYPT or COP_DECRYPT */
     SYMCRYPT_AES_EXPANDED_KEY key;
 };
 
 struct cipher_ecb_ctx {
-    INT32 enc;                     /* COP_ENCRYPT or COP_DECRYPT */
     SYMCRYPT_AES_EXPANDED_KEY key;
 };
 
 struct cipher_xts_ctx {
-    INT32 enc;                     /* COP_ENCRYPT or COP_DECRYPT */
     BYTE iv[SYMCRYPT_AES_BLOCK_SIZE];
     SYMCRYPT_XTS_AES_EXPANDED_KEY key;
 };
@@ -36,7 +31,6 @@ struct cipher_xts_ctx {
 #define SCOSSL_GCM_MAX_TAG_LENGTH (16)
 
 struct cipher_gcm_ctx {
-    INT32 enc;                     /* COP_ENCRYPT or COP_DECRYPT */
     INT32 operationInProgress;
     BYTE iv[SCOSSL_GCM_IV_LENGTH];
     INT32 ivlen;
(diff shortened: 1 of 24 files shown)


SymCryptEngine/src/sc_ossl_rsapss.c
```diff
@@ -54,9 +51,7 @@ SCOSSL_STATUS sc_ossl_rsapss_sign(_Inout_ EVP_PKEY_CTX *ctx, _Out_writes_opt_(*s
     size_t cbResult = 0;
     SYMCRYPT_ERROR SymError = SYMCRYPT_NO_ERROR;
     int ret = -1;
-    // We should have this localKeyCtx kept in some extension to EVP_PKEY_CTX
-    // Currently not sure how to achieve this with 1.1.1 APIs (EVP_PKEY_get_ex_data is introduced in OpenSSL 3.0)
-    SC_OSSL_RSA_KEY_CONTEXT localKeyCtx;
+    SC_OSSL_RSA_KEY_CONTEXT *keyCtx = NULL;
     PCSYMCRYPT_HASH sc_ossl_mac_algo = NULL;
```
mlindgren, line 55: >sc_ossl_mac_algo [](http://example.com/codeflow?start=20&length=16)

Nit: we mostly use camel case for local variables, but this one is an exception.

SymCryptEngine/src/sc_ossl_rsa.c
```diff
@@ -49,7 +47,7 @@ SCOSSL_RETURNLENGTH sc_ossl_rsa_pub_enc(int flen, _In_reads_bytes_(flen) const u
     int ret = -1;
     const RSA_METHOD *ossl_rsa_meth = NULL;
     PFN_RSA_meth_pub_enc pfn_rsa_meth_pub_enc = NULL;
-    SC_OSSL_RSA_KEY_CONTEXT *keyCtx = RSA_get_ex_data(rsa, rsa_sc_ossl_idx);
+    SC_OSSL_RSA_KEY_CONTEXT *keyCtx = RSA_get_ex_data(rsa, scossl_rsa_idx);
```
mlindgren, line 50: >scossl_rsa_idx [](http://example.com/codeflow?start=59&length=14)

I noticed that in some places you changed sc_ossl to scossl, but in others you didn't. I guess you only changed things that you were going to rename anyway? If so, that's fine.

Maybe we should run `find . -type f -print0 | xargs -0 sed -i 's/sc_ossl/scossl/g` and do a follow up PR? (I love editing code with sed. What could go wrong? 😀)

mlindgren:
> // creates that structure if it doesn't already exist.

Nit: these comments are now slightly inaccurate

---
Refers to: SymCryptEngine/src/sc_ossl_pkey_meths.c:87 in 8f260e8. [](commit_id = 8f260e87c183bdf7365a7a750084c25af0f61cf3, deletion_comment = False)

SymCryptEngine/src/sc_ossl_rsapss.c
samuel-lee-msft: Fixed in latest

mlindgren:
>         // return 1;

Does this code currently work without this functionality? Seems like we should either be returning an error if this path doesn't work, or if it does and the above isn't actually necessary, just remove it.

---
Refers to: SymCryptEngine/src/sc_ossl_hkdf.c:300 in 8f260e8. [](commit_id = 8f260e87c183bdf7365a7a750084c25af0f61cf3, deletion_comment = False)

SymCryptEngine/src/sc_ossl_ciphers.c
```diff
@@ -126,14 +120,13 @@ static SCOSSL_STATUS sc_ossl_aes_cbc_ctrl(_In_ EVP_CIPHER_CTX *ctx, int type, in
...
-    if( _hidden_aes_128_cbc == NULL
-        && ((_hidden_aes_128_cbc = EVP_CIPHER_meth_new(NID_aes_128_cbc, SYMCRYPT_AES_BLOCK_SIZE , AES_128_KEY_SIZE)) == NULL
-            || !EVP_CIPHER_meth_set_iv_length(_hidden_aes_128_cbc,16)
-            || !EVP_CIPHER_meth_set_flags(_hidden_aes_128_cbc, AES_CBC_FLAGS)
-            || !EVP_CIPHER_meth_set_init(_hidden_aes_128_cbc, sc_ossl_aes_cbc_init_key)
-            || !EVP_CIPHER_meth_set_do_cipher(_hidden_aes_128_cbc, sc_ossl_aes_cbc_cipher)
-            || !EVP_CIPHER_meth_set_ctrl(_hidden_aes_128_cbc, sc_ossl_aes_cbc_ctrl)
-            || !EVP_CIPHER_meth_set_impl_ctx_size(_hidden_aes_128_cbc, sizeof(struct cipher_cbc_ctx))) )
+    if( (_hidden_aes_128_cbc = EVP_CIPHER_meth_new(NID_aes_128_cbc, SYMCRYPT_AES_BLOCK_SIZE , AES_128_KEY_SIZE)) == NULL
+        || !EVP_CIPHER_meth_set_iv_length(_hidden_aes_128_cbc,16)
```
mlindgren, line 124: Nit: Can we have a macro for the IV length as well?

SymCryptEngine/src/sc_ossl_ciphers.c
```diff
@@ -525,26 +527,16 @@ int sc_ossl_ciphers(ENGINE *e, const EVP_CIPHER **cipher,
...
-    SYMCRYPT_ERROR SymError = SYMCRYPT_NO_ERROR;
-    PBYTE ctx_iv = EVP_CIPHER_CTX_iv_noconst(ctx);
-    if( enc != SC_OSSL_ENCRYPTION_MODE_NOCHANGE )
-    {
-        cipherCtx->enc = enc;
-    }
-    if( iv )
-    {
-        memcpy(ctx_iv, iv, EVP_CIPHER_CTX_iv_length(ctx));
-    }
```
mlindgren, line 543: We're not doing anything with the iv or encryption mode now?

mlindgren (approved):
:shipit:

samuel-lee-msft:
>         // return 1;

I'm completing HKDF in a follow up PR - it's currently a draft as it relies on SymCrypt API changes. We will move to something not too distant from the commented out code.

---
In reply to: [989306961](https://github.com/microsoft/SymCrypt-OpenSSL/pull/26#issuecomment-989306961) [](http://example.com/codeflow?ancestors=989306961,989306961)

---
Refers to: SymCryptEngine/src/sc_ossl_hkdf.c:300 in 968ea7e. [](commit_id = 968ea7e1df95ac75ea8c7e986ffe860bb78a8b19, deletion_comment = False)

SymCryptEngine/src/sc_ossl_ciphers.c
samuel-lee-msft: Moved to just use SYMCRYPT_AES_BLOCK_SIZE here (for CBC the IV is just another block at the start).

SymCryptEngine/src/sc_ossl_ciphers.c
samuel-lee-msft: For CBC we're relying on OpenSSL to track the IV and encryption mode, rather than duplicating the logic in SCOSSL. For ECB I can actually remove a bit more IV stuff, as there is no IV in ECB.

For GCM / CCM we're handling the IV ourselves (we specify EVP_CIPH_CUSTOM_IV), but we rely on OpenSSL to track the encryption mode.

PR #34, opened 2022-01-07
When verifying RSA-PSS signature with auto-saltlen always fallback

samuel-lee-msft:
+ Previously would only fallback for RSA keys, not specifically RSA-PSS
  keys
+ Add test cases for DigestSign/Verify with RSA-PSS

### DIFF
--- a/SymCryptEngine/src/scossl_pkey_meths.c
+++ b/SymCryptEngine/src/scossl_pkey_meths.c
@@ -50,12 +50,31 @@ static int scossl_pkey_rsa_sign(_Inout_ EVP_PKEY_CTX *ctx, _Out_writes_bytes_(*s
     return _openssl_pkey_rsa_sign(ctx, sig, siglen, tbs, tbslen);
 }
 
+// Call SymCrypt engine RSA-PSS verify, unless auto salt-length specified (not yet supported by SymCrypt)
+static int scossl_pkey_rsapss_verify(_Inout_ EVP_PKEY_CTX *ctx, _In_reads_bytes_(siglen) const unsigned char *sig, size_t siglen,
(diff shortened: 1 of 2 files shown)


SymCryptEngine/src/scossl_pkey_meths.c
```diff
@@ -65,17 +84,7 @@ static int scossl_pkey_rsa_verify(_Inout_ EVP_PKEY_CTX *ctx, _In_reads_bytes_(si
...
-        if( cbSalt != RSA_PSS_SALTLEN_AUTO )
-        {
-
-            return scossl_rsapss_verify(ctx, sig, siglen, tbs, tbslen);
-        }
-        SCOSSL_LOG_INFO("SymCrypt Engine does not support RSA_PSS_SALTLEN_AUTO saltlen - falling back to OpenSSL");
+        return scossl_pkey_rsapss_verify(ctx, sig, siglen, tbs, tbslen);
     }
 
     return _openssl_pkey_rsa_verify(ctx, sig, siglen, tbs, tbslen);
```
mlindgren, line 90: >return _openssl_pkey_rsa_verify(ctx, sig, siglen, tbs, tbslen); [](http://example.com/codeflow?start=4&length=63)

Can this function be called with PSS padding and non-auto salt length? If so, shouldn't it go to our implementation?

mlindgren (approved):
:shipit:

SymCryptEngine/src/scossl_pkey_meths.c
samuel-lee-msft: Yes - so in this case we rely on the OpenSSL default implementation of RSA at the EVP_PKEY layer to call into lower level RSA methods. We intercept non-PSS padding (which can be FIPS certified) by setting SCOSSL's RSA methods as the defaults.

PR #35, opened 2022-01-07
Log SCOSSL errors to OpenSSL's ERR infrastructure

samuel-lee-msft:
+ This is in addition to logging to stderr or a log file specified by
  the caller
+ For now do not specify all the functions which can generate errors and
  error reasons up front, but instead log all SCOSSL errors as being for
  an unknown function and with a generic reason of "Error".
  + Log additional information with the errors indicating the function
    and error reason using the existing data that SCOSSL provides to
    error logging functions

### DIFF
--- a/SymCryptEngine/src/scossl_helpers.c
+++ b/SymCryptEngine/src/scossl_helpers.c
@@ -19,17 +19,78 @@ extern "C" {
 static int _traceLogLevel = SCOSSL_LOG_LEVEL_INFO;
 static char *_traceLogFilename = NULL;
 
-#define SCOSSL_ERR_UNKNOWN_FUNC_CODE        (1)
-#define SCOSSL_ERR_UNKNOWN_REASON_CODE      (1)
-
 static int _scossl_err_library_code = 0;
 
-static ERR_STRING_DATA SCOSSL_ERR_strings[] = {
-    {0,                                             "SCOSSL"},  // library name
-    {ERR_PACK(0, SCOSSL_ERR_UNKNOWN_FUNC_CODE, 0),  ""},        // unknown function name
-    {ERR_PACK(0, 0, SCOSSL_ERR_UNKNOWN_REASON_CODE),"Error"},   // unknown reason name
+static ERR_STRING_DATA SCOSSL_ERR_library_string[] = {
+    {0, "SCOSSL"},  // library name
     {0, NULL}
 };
+
+static ERR_STRING_DATA SCOSSL_ERR_function_strings[] = {
+    {ERR_PACK(0, SCOSSL_ERR_F_AES_CCM_CIPHER, 0), "scossl_aes_ccm_cipher"},
+    {ERR_PACK(0, SCOSSL_ERR_F_AES_CCM_CTRL, 0), "scossl_aes_ccm_ctrl"},
+    {ERR_PACK(0, SCOSSL_ERR_F_AES_CCM_TLS, 0), "scossl_aes_ccm_tls"},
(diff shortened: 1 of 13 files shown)


mlindgren:
>     FILE *fp = stderr;

Did we decide against logging to syslog?

---
Refers to: SymCryptEngine/src/scossl_helpers.c:131 in aeb7337. [](commit_id = aeb7337a15bc61967c3b3dbd9fa0bb70bb2fa462, deletion_comment = False)

SymCryptEngine/src/scossl_helpers.c
```diff
@@ -105,15 +192,28 @@ void _scossl_log(
...
+    // Add error string indicating the error details as error data
+    ERR_add_error_data(1, paraBuf);
+
+    // Log details to stderr or a log file
+    ERR_error_string_n(ERR_PACK(_scossl_err_library_code, func_code, reason_code), errStringBuf, sizeof(errStringBuf));
+
     fp = _open_trace_log_filename();
-    fprintf(fp, "[%s] %s: %s\n", trace_level_prefix, func, paraBuf);
+    fprintf(fp, "[%s] %s:%s at %s, line %d\n", trace_level_prefix, errStringBuf, paraBuf, file, line);
     _close_trace_log_filename(fp);
```
mlindgren, line 206: >_close_trace_log_filename [](http://example.com/codeflow?start=4&length=25)

Do we want to open and close this again every time, or should we have a static file that we only open once and close on shutdown? Not sure what the perf impact of re-opening the handle every time is.

SymCryptEngine/src/scossl_helpers.c
```diff
@@ -166,81 +269,84 @@ void _scossl_log_bignum(
...
+    const char *file,
+    int line,
     char *description,
     SYMCRYPT_ERROR scError)
 {
     switch( scError  )
     {
         case SYMCRYPT_NO_ERROR:
-            _scossl_log(trace_level, func, "%s - SYMCRYPT_NO_ERROR (0x%x)", description, scError);
+            _scossl_log(trace_level, func_code, reason_code, file, line, "%s - SYMCRYPT_NO_ERROR (0x%x)", description, scError);
```
mlindgren, line 289: >_scossl_log [](http://example.com/codeflow?start=12&length=11)

Seems like the log call is ultimately always the same. You could just use the switch statement to pick an error string and then only have one log call which takes the error string as a parameter.

mlindgren (approved):
:shipit:

samuel-lee-msft:
>     FILE *fp = stderr;

We rely on callers to log OpenSSL errors to the appropriate logging infrastructure for the platform. We could go as far as removing the stderr / logging file, but I think it's helpful for developer debugging. Something to consider going forward.

---
In reply to: [1016094574](https://github.com/microsoft/SymCrypt-OpenSSL/pull/35#issuecomment-1016094574) [](http://example.com/codeflow?ancestors=1016094574,1016094574)

---
Refers to: SymCryptEngine/src/scossl_helpers.c:131 in aeb7337. [](commit_id = aeb7337a15bc61967c3b3dbd9fa0bb70bb2fa462, deletion_comment = False)

SymCryptEngine/src/scossl_helpers.c
samuel-lee-msft: Hmm, probably we want to only open once and close on shutdown, and probably this should be thread safe to avoid attempting multiple concurrent fprintfs. Will update.

samuel-lee-msft: Done in latest

SymCryptEngine/src/scossl_helpers.c
samuel-lee-msft: Good point - refactored in latest

PR #36, opened 2022-01-19
Fix memory leak in DH when falling back to OpenSSL

samuel-lee-msft:
+ Add SslPlay test case which uses default OpenSSL implementation
+ Call DH_OpenSSL finish method in scossl_dh_finish to free anything
  which may have been allocated in fallback to OpenSSL
+ Free BNs in cleanup in DH (technically had some memory leaks in
  failure cases before)


samuel-lee-msft:
Going ahead with merging this without review as it's a small change and fixes a memory leak for OE.

mlindgren:
lgtm

---
In reply to: [1016801691](https://github.com/microsoft/SymCrypt-OpenSSL/pull/36#issuecomment-1016801691) [](http://example.com/codeflow?ancestors=1016801691,1016801691)

PR #37, opened 2022-01-24
Various logging tweaks

samuel-lee-msft:
+ Do not leak CRYPTO_RWLOCK
+ Introduce ability to set different levels of logging verbosity for
  using OpenSSL ERR and stderr/logfile. Set default level of logging for
  OpenSSL ERR to ERROR to avoid problems in OpenSSL unit tests.
+ Update Engine id and name

### DIFF
--- a/SymCryptEngine/src/scossl.c
+++ b/SymCryptEngine/src/scossl.c
@@ -20,8 +20,8 @@ extern "C" {
 int scossl_module_initialized = 0;
 
 /* The constants used when creating the ENGINE */
-static const char* engine_scossl_id = "symcrypt";
-static const char* engine_scossl_name = "Symcrypt Engine";
+static const char* engine_scossl_id = "SCOSSL";
+static const char* engine_scossl_name = "SymCrypt engine for OpenSSL";
 static EC_KEY_METHOD* scossl_eckey_method = NULL;
 static RSA_METHOD* scossl_rsa_method = NULL;
 // static DSA_METHOD* scossl_dsa_method = NULL;
@@ -45,6 +45,7 @@ SCOSSL_STATUS scossl_destroy(ENGINE* e)
     DH_meth_free(scossl_dh_method);
     scossl_dh_method = NULL;
     CRYPTO_free_ex_index(CRYPTO_EX_INDEX_DH, scossl_dh_idx);
+    scossl_destroy_logging();
 
     return SCOSSL_SUCCESS;
 }
@@ -60,7 +61,7 @@ static SCOSSL_STATUS scossl_bind_engine(ENGINE* e)
     }
 
     scossl_eckey_method = EC_KEY_METHOD_new(EC_KEY_OpenSSL());
-    scossl_rsa_method = RSA_meth_new("SymCrypt RSA Method", 0);
+    scossl_rsa_method = RSA_meth_new("SCOSSL RSA Method", 0);
     // scossl_dsa_method = DSA_meth_dup(DSA_OpenSSL());
     scossl_dh_method = DH_meth_dup(DH_OpenSSL());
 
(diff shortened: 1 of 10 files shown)


mlindgren:
>         _traceLogFile = fopen(_traceLogFilename, "a");

`_traceLogFilename` may be (usually is) NULL here. It seems to sort of work anyway - I guess fopen just returns NULL - but valgrind catches this as as an error. Can you change this to only call fopen when `_traceLogFileName != NULL`?

---
Refers to: SymCryptEngine/src/scossl_helpers.c:160 in 2291285. [](commit_id = 22912858fce46cd2fe3b61ea70b7a946bf739ad2, deletion_comment = False)

SymCryptEngine/src/scossl.c
```diff
@@ -20,8 +20,8 @@ extern "C" {
 int scossl_module_initialized = 0;
 
 /* The constants used when creating the ENGINE */
-static const char* engine_scossl_id = "symcrypt";
-static const char* engine_scossl_name = "Symcrypt Engine";
+static const char* engine_scossl_id = "SCOSSL";
```
mlindgren, line 23: Changing this potentially breaks existing callers. We probably don't have any "in the wild," but I'd have to update my patch code at the very least. Not sure if we'd have to make any changes to our other engines.

I think it makes sense for the engine ID to be "SymCrypt" anyway. In any context where this string appears, it should be obvious that this is an OpenSSL engine.

SymCryptEngine/src/scossl.c
```diff
@@ -20,8 +20,8 @@ extern "C" {
...
+static const char* engine_scossl_name = "SymCrypt engine for OpenSSL";
```
mlindgren, line 24: >for OpenSSL [](http://example.com/codeflow?start=57&length=11)

Likewise - there's no compatibility concern with changing this, but the only place I can think of where it's used is when you dump/log OpenSSL engine info (e.g. via `openssl engine`). So it should be obvious that it's an OpenSSL engine.

SymCryptEngine/src/scossl.c
```diff
@@ -60,7 +61,7 @@ static SCOSSL_STATUS scossl_bind_engine(ENGINE* e)
     }
 
     scossl_eckey_method = EC_KEY_METHOD_new(EC_KEY_OpenSSL());
-    scossl_rsa_method = RSA_meth_new("SymCrypt RSA Method", 0);
+    scossl_rsa_method = RSA_meth_new("SCOSSL RSA Method", 0);
```
mlindgren, line 64: >SCOSSL [](http://example.com/codeflow?start=38&length=6)

Where is this string used? I think it makes more sense to keep SCOSSL as an internal prefix and use SymCrypt or "SymCrypt engine" anywhere that might be user-visible. Niels might disagree with me here though; I know he wanted a clear distinction between SymCrypt itself and SCOSSL. But SCOSSL wraps SymCrypt, so really the implementation here _is_ SymCrypt.

mlindgren (changes requested):
:clock1:

SymCryptEngine/src/scossl.c
samuel-lee-msft: I want it to be obvious to callers from a dump that they need to look at this repo. I'll change this to "SCOSSL (SymCrypt engine for OpenSSL)". I also want to put a version number in here when we snap to a LTSB.

SymCryptEngine/src/scossl.c
samuel-lee-msft: Sure, I'll leave this as "symcrypt" to avoid compatibility issues.

SymCryptEngine/src/scossl.c
samuel-lee-msft: I'm not sure it is used much by callers, but the string is available by calling [`RSA_meth_get0_name`](https://www.openssl.org/docs/man1.1.1/man3/RSA_meth_get0_name.html). I think similar to the Engine name, this should obviously pointer callers to this repo first; I'll update to "SCOSSL (SymCrypt engine for OpenSSL) RSA method"

mlindgren (approved):
:shipit:

PR #38, opened 2022-01-26
Prepare for upcoming SymCrypt API changes

mlindgren:
This PR includes a few changes to prepare for some upcoming SymCrypt changes. It should be backwards-compatible with existing versions of SymCrypt.

- Use `SYMCRYPT_MODULE_INIT()` macro instead of explicitly calling `SymCryptModuleInit`. The macro automatically passes in the right arguments, and will work with the new version which is removing the patch version as an argument.
- Remove reference to `SYMCRYPT_UNUSED` error code, which is being removed,
- Use CMake's [find_library](https://cmake.org/cmake/help/latest/command/find_library.html) function to find the symcrypt library on the system. This will hopefully be slightly more robust than just directly specifying the library name in `target_link_libraries`. In the future, we may want to create a CMake package config for SymCrypt and use `find_package`, but this requires additional work.


samuel-lee-msft (approved):
Lgtm

PR #42, opened 2022-02-07
Various tidying to reduce warnings in build

samuel-lee-msft:
+ Allow libsymcrypt to be found from root of SymCrypt-OpenSSL directory
+ Update README to make compilation instructions a little clearer
+ Add CMake project version/description

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 3.13.0)
 
-project(scossl)
+project(SymCrypt-OpenSSL
+    VERSION 1.0.0
+    DESCRIPTION "The SymCrypt engine for OpenSSL (SCOSSL)"
+    HOMEPAGE_URL "https://github.com/microsoft/SymCrypt-OpenSSL")
 
 include_directories(${SYMCRYPT_ROOT_DIR}/inc)
 
@@ -26,5 +29,3 @@ endif()
 add_subdirectory (SymCryptEngine/static)
 add_subdirectory (SymCryptEngine/dynamic)
 add_subdirectory (SslPlay)
-
-
--- a/README.md
+++ b/README.md
@@ -43,19 +43,34 @@ Note that just because an algorithm is FIPS certifiable, does not mean it is rec
 2. Use of unsupported digests in RSA signatures and TLS PRF
 3. Use of multi-prime (more than 2-prime) RSA
 
+## Versioning and Servicing
+
+As of version 1.0.0, SCOSSL uses the versioning scheme defined by the [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html) specification. This means:
+
+- Major version changes introduce ABI and/or API breaking changes
+- Minor version changes introduce backwards compatible additional functionality or improvements, and/or bug fixes
+- Patch version changes introduce backwards compatible bug fixes
+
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -47,14 +47,16 @@ Note that just because an algorithm is FIPS certifiable, does not mean it is rec
...
-
-```
-cp <SymCryptRepo>/bin/module/<arch>/LinuxUserMode/<module_name>/libsymcrypt.so ./
-mkdir bin; cd bin
-cmake .. -DSYMCRYPT_ROOT_DIR=<SymCryptRepo> -DOPENSSL_ROOT_DIR=<OpensslInstallDirectory> -DCMAKE_TOOLCHAIN_FILE=../cmake-toolchain/LinuxUserMode-<arch>.cmake -DCMAKE_BUILD_TYPE=Release
-cmake --build .
-```
+1. Follow Linux build instructions from SymCrypt repository [SymCrypt](https://github.com/Microsoft/SymCrypt) to build the Linux SymCrypt module
+    * You can either install the Linux SymCrypt module (i.e libsymcrypt.so* to /usr/lib/, and inc/* to /usr/include/), or
+    * Copy the built module to the root of the SymCrypt-OpenSSL repo `cp <SymCryptRepo>/bin/module/<arch>/LinuxUserMode/<module_name>/libsymcrypt.so ./`
```
mlindgren, line 52: >`cp <SymCryptRepo>/bin/module/<arch>/LinuxUserMode/<module_name>/libsymcrypt.so [](http://example.com/codeflow?start=69&length=79)

I'm still looking for a better way to handle these cross-depot dependencies in general. CMake has its own packaging system that would probably work, but it's CMake-specific and looks kind of unpleasant to deal with. Not sure if there's a better option.

CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 3.13.0)
 
-project(scossl)
+project(SymCrypt-OpenSSL
+    VERSION 1.0.0
```
mlindgren, line 4: >VERSION 1.0.0 [](http://example.com/codeflow?start=4&length=13)

I assume we'll be using semantic versioning with this repo too. Maybe we can put a blurb about that in the SCOSSL readme, too?

SymCryptEngine/dynamic/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++1y -Wall -Wextra -Wno-unused-par
 find_package(OpenSSL REQUIRED)
 include_directories(${OPENSSL_INCLUDE_DIR})
 
-find_library(SYMCRYPT_LIBRARY symcrypt)
+find_library(SYMCRYPT_LIBRARY symcrypt PATHS ../../)
```
mlindgren, line 15: >../../ [](http://example.com/codeflow?start=45&length=6)

Use `${CMAKE_SOURCE_DIR}` instead so that this doesn't have to change if the directory structure changes.

mlindgren (dismissed):
:clock1:

README.md
samuel-lee-msft: I think what we have here makes sense for right now, but definitely think we should look at how to make this easier.

mlindgren (approved):
:shipit:

PR #44, opened 2022-03-28
Add selftest flags for RSA, ECC and DH

samuel-lee-msft:


mlindgren (approved):
Thanks for taking care of this. LGTM.

PR #45, opened 2022-04-25
Avoid redundant EC_KEY tests on key generation

samuel-lee-msft:
+ Use EC_KEY_set_public_key with an EC_POINT created by SCOSSL, rather
  than using EC_KEY_set_public_key_affine_coordinates, which also
  performs validation that SymCrypt has already taken care of


mlindgren (commented):
Sorry I missed this. Changes look good to me.

PR #46, opened 2022-04-28
Ensure method flags are compatible with FIPS mode OpenSSL

samuel-lee-msft:
+ Copy flags from base methods for RSA and EVP methods
+ Specify the normally unused FIPS flags for ciphers and digests


SymCryptEngine/src/scossl_ciphers.c
```diff
@@ -373,7 +375,8 @@ SCOSSL_RETURNLENGTH scossl_aes_ccm_cipher(
 static _Success_(return > 0) int scossl_aes_ccm_ctrl(_Inout_ EVP_CIPHER_CTX *ctx, int type, int arg, _Inout_ void *ptr);
 #define AES_CCM_FLAGS   (EVP_CIPH_FLAG_DEFAULT_ASN1|EVP_CIPH_CCM_MODE|EVP_CIPH_CUSTOM_COPY \
                         |EVP_CIPH_CUSTOM_IV|EVP_CIPH_CUSTOM_IV_LENGTH|EVP_CIPH_FLAG_CUSTOM_CIPHER \
-                        |EVP_CIPH_ALWAYS_CALL_INIT|EVP_CIPH_CTRL_INIT|EVP_CIPH_FLAG_AEAD_CIPHER)
+                        |EVP_CIPH_ALWAYS_CALL_INIT|EVP_CIPH_CTRL_INIT|EVP_CIPH_FLAG_AEAD_CIPHER \
+                        |EVP_CIPH_FLAG_FIPS )
```
mlindgren, line 379: >EVP_CIPH_FLAG_FIPS [](http://example.com/codeflow?start=25&length=18)

I think always setting this is fine for our purposes for now, but it could lead to future callers shooting themselves in the foot if they compile their own modules. I wonder if we need to add an API (or just a public global) to the SymCrypt interface to indicate whether or not it is running in/was compiled with FIPS checks enabled

That said, compiling it with FIPS checks enabled is also not sufficient since you still need to get the binary certified, so we can't completely prevent callers from using our code incorrectly.

mlindgren (approved):
:shipit:

PR #49, opened 2022-08-01
Do not set -DDBG=1 when build type is RelWithDebInfo

mlindgren:
In some environments, we use CMake's RelWithDebInfo build type so that we can do an optimized build but still get debug symbols, which we then copy and strip from the binary in a separate step. When building this way, we don't want the `DBG` flag to be set.


samuel-lee-msft (approved):
LGTM

PR #53, opened 2022-10-27
Fix RSA-PSS maximum salt length calculation

samuel-lee-msft:
+ Previously was wrong when RSA keys had a modulus size in bits which is 1 modulo 8. In this case using SCOSSL to sign/verify signatures with RSA_PSS_SALTLEN_MAX would fail with invalid argument error in SymCrypt
  + Problem is with calculating length of EM as per RFC 3447 section 8.1.1 (step 1) - "Note that the octet length of EM will be one less than k if modBits - 1 is divisible by 8 and equal to k otherwise."
+ Amend SslPlay testcases to test this corner case

### DIFF
--- a/SymCryptEngine/src/scossl_rsapss.c
+++ b/SymCryptEngine/src/scossl_rsapss.c
@@ -60,6 +60,7 @@ SCOSSL_STATUS scossl_rsapss_sign(_Inout_ EVP_PKEY_CTX *ctx, _Out_writes_opt_(*si
     EVP_MD *mgf1Digest;
     int type = 0;
     int cbSalt = RSA_PSS_SALTLEN_DIGEST;
+    int cbDigest, cbSaltMax;
 
     if( EVP_PKEY_CTX_get_signature_md(ctx, &messageDigest) <= 0 )
     {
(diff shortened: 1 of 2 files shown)


SymCryptEngine/src/scossl_rsapss.c
```diff
@@ -97,16 +98,19 @@ SCOSSL_STATUS scossl_rsapss_sign(_Inout_ EVP_PKEY_CTX *ctx, _Out_writes_opt_(*si
         return SCOSSL_UNSUPPORTED;
     }
 
+    cbDigest = EVP_MD_size(messageDigest);
+    cbSaltMax = (( RSA_bits(rsa) + 6 ) / 8) - cbDigest - 2; // ceil((RSA_bits(rsa) - 1) / 8) - cbDigest - 2
```
mlindgren, line 102: >cbDigest [](http://example.com/codeflow?start=46&length=8)

Any risk of this underflowing if `cbDigest` is too large? I'm not sure what limits `EVP_MD_size` places on it.

mlindgren (approved):
:shipit:

ccalik (approved):

PR #54, opened 2022-12-08
Explicitly remove stitched EVP ciphers from OpenSSL internal list

samuel-lee-msft:
+ On SCOSSL bind, load all EVP ciphers and then remove the stitched ciphers that we do not support in SCOSSL. This is a workaround for callers using EVP_get_cipherbyname.


mlindgren (approved):
:shipit:

PR #57, opened 2023-02-01
SymCrypt provider base

mamckee:
This PR adds a baseline SymCrypt provider + digest implementation for OpenSSL 3. It also prepared the existing SymCrypt engine code for further refactoring into common ScOSSL code.
- Added SymCryptProvider to the project
  - Added digest implementation for the following algorithms using the default provider as a model
    - MD5 (for app compat and parity with SymCrypt engine)
    - SHA1 (for app compat and parity with SymCrypt engine)
    - SHA2 (256, 384, and 512)
    - SHA3 (256, 384, and 512) _New_
  - Added dispatch table generation macros for additional provider functions (subject to change as implemented)
- Refactored SymCrypt engine
  - Prefixed engine-specific functions to distinguish from the provider and common functions
  - Moved common functionality to ScosslCommon
  - Added deprecation warning ignore, as most engine code has been marked for deprecation in OpenSSL 3.


SymCryptEngine/src/scossl_dh.c
```diff
@@ -338,7 +338,7 @@ SCOSSL_STATUS scossl_dh_import_keypair(
     return res;
 }
 
-PSYMCRYPT_DLGROUP scossl_initialize_safeprime_dlgroup(_Inout_ PSYMCRYPT_DLGROUP* ppDlgroup,
+PSYMCRYPT_DLGROUP e_scosslinitialize_safeprime_dlgroup(_Inout_ PSYMCRYPT_DLGROUP* ppDlgroup,
```
samuel-lee-msft, line 341: >e_scosslinitialize_safeprime_dlgroup [](http://example.com/codeflow?start=18&length=36)

e_scossl_initialize_safeprime_dlgroup

SymCryptEngine/src/scossl_dh.c
```diff
@@ -380,7 +380,7 @@ SCOSSL_STATUS scossl_dh_init_static()
 // returns SCOSSL_FALLBACK when the dh is not supported by the engine, so we should fallback to OpenSSL
 // returns SCOSSL_FAILURE on an error
 // returns SCOSSL_SUCCESS and sets pKeyCtx to a pointer to an initialized SCOSSL_DH_KEY_CONTEXT on success
-SCOSSL_STATUS scossl_get_dh_context_ex(_Inout_ DH* dh, _Out_ PSCOSSL_DH_KEY_CONTEXT* ppKeyCtx, BOOL generate)
+SCOSSL_STATUS e_scosslget_dh_context_ex(_Inout_ DH* dh, _Out_ PSCOSSL_DH_KEY_CONTEXT* ppKeyCtx, BOOL generate)
```
samuel-lee-msft, line 383: >e_scosslget_dh_context_ex [](http://example.com/codeflow?start=14&length=25)

e_scossl_get_dh_context_ex

SymCryptEngine/src/scossl_dh.c
mamckee: Doing a double pass now. Looks like I missed this a few places

SymCryptProvider/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+set(SCOSSL_SOURCES
+    ./src/p_scossl_base.c
+    ./src/p_scossl_digests.c
+)
+
+# if (SCOSSL_SSHKDF)
+#     list(APPEND SCOSSL_SOURCES ./src/p_scossl_sshkdf.c)
+#     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DSCOSSL_SSHKDF")
+#     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DSCOSSL_SSHKDF")
+# endif()
```
samuel-lee-msft, line 28: Pretty sure this won't be necessary for the provider - the KDF API comes with OpenSSL 3 out of the box, the reason for the divergence on the 1.1.1 side is that SSHKDF in SCOSSL only works with a patched 1.1.1 OpenSSL.

SymCryptProvider/src/p_scossl_digests.c
```diff
@@ -0,0 +1,53 @@
...
+
+SCOSSL_STATUS p_scossl_digest_get_params(_Inout_ OSSL_PARAM params[], size_t blocksize, size_t size)
+{
+    OSSL_PARAM *p = NULL;
+
+    p = OSSL_PARAM_locate(params, OSSL_DIGEST_PARAM_BLOCK_SIZE);
+    if (p != NULL && !OSSL_PARAM_set_size_t(p, blocksize))
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_SET_PARAMETER);
+        return 0;
```
samuel-lee-msft, line 31: >0 [](http://example.com/codeflow?start=15&length=1)

SCOSSL_FAILURE

SymCryptProvider/src/p_scossl_digests.h
```diff
@@ -0,0 +1,73 @@
...
+
+#include <openssl/core_dispatch.h>
+#include <openssl/params.h>
+#include "scossl_helpers.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// lowercase, CamelCase, and UPPERCASE must be provided to reconcile differences
```
samuel-lee-msft, line 13: This looks good, but it looks like it would make sense to also support OSSL_FUNC_DIGEST_DUPCTX (SymCrypt##Ccalg##StateCopy) and OSSL_FUNC_DIGEST_DIGEST (SymCrypt##Ccalg) - this should enable more performant callers.

I also wonder whether it makes sense to name the macro something like IMPLEMENT_SCOSSL_DIGEST_BASIC - if we want to expose SHAKE / CSHAKE through the provider we have some more interesting interactions to contend with

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -0,0 +1,261 @@
...
+
+static const OSSL_DISPATCH p_scossl_base_dispatch[] = {
+    {OSSL_FUNC_PROVIDER_TEARDOWN, (void (*)(void))p_scossl_teardown},
+    {OSSL_FUNC_PROVIDER_GETTABLE_PARAMS, (void (*)(void))p_scossl_gettable_params},
+    {OSSL_FUNC_PROVIDER_GET_PARAMS, (void (*)(void))p_scossl_get_params},
+    {OSSL_FUNC_PROVIDER_QUERY_OPERATION, (void (*)(void))p_scossl_query_operation},
+    {0, NULL}};
+
+int OSSL_provider_init(_In_ const OSSL_CORE_HANDLE *handle,
+                       _In_ const OSSL_DISPATCH *in,
```
samuel-lee-msft, line 237: >_In_ const OSSL_DISPATCH *in, [](http://example.com/codeflow?start=23&length=29)

I notice this argument is not being used. My understanding it that this is a bunch of functions that OpenSSL passes to the provider.

Do you know what the implication is for the Provider code calling OpenSSL methods directly (i.e. OPENSSL_malloc), rather than using methods from the dispatch table (i.e. OSSL_FUNC_CRYPTO_MALLOC) passed in here?

SymCryptProvider/src/p_scossl_digests.h
```diff
@@ -0,0 +1,73 @@
...
+        return SCOSSL_SUCCESS;                                                      \
+    }                                                                               \
+    static SCOSSL_STATUS scossl_##lcalg##_final(                                    \
+        _Inout_ SYMCRYPT_##UCALG##_STATE *dctx,                                     \
+        _Out_writes_(SYMCRYPT_##UCALG##_RESULT_SIZE) unsigned char *out,            \
+        _Out_ size_t *outl,                                                         \
+        size_t outsz)                                                               \
+    {                                                                               \
+        if (outsz < SYMCRYPT_##UCALG##_RESULT_SIZE)                                 \
+            return 0;                                                               \
```
samuel-lee-msft, line 46: >0 [](http://example.com/codeflow?start=19&length=1)

SCOSSL_FAILURE

SymCryptProvider/src/p_scossl_digests.h
mamckee: We may need to build those implementations out on their own then and not be able to generalize with a macro. I think for now we can just leave it though.

SymCryptProvider/src/p_scossl_base.c
mamckee: The default provider just uses the core BIO functions and entropy functions. They do copy the memory mangement functions for the FIPS provider though. 

I'm still trying to figure out why though, because the default provider, oqs provider, and symcrypt provider use those functions without issue. My best guess is it has something to do with avoiding a dependency on libcrypto from the provider.

SymCryptProvider/src/p_scossl_digests.h
```diff
@@ -0,0 +1,113 @@
...
+    static SCOSSL_STATUS p_scossl_##lcalg##_update(                                   \
+        _Inout_ SYMCRYPT_##UCALG##_STATE *dctx,                                       \
+        _In_reads_bytes_(inl) const unsigned char *in,                                \
+        size_t inl)                                                                   \
+    {                                                                                 \
+        if (in == NULL)                                                               \
+            return SCOSSL_FAILURE;                                                    \
+                                                                                      \
+        if (inl > 0)                                                                  \
+            SymCrypt##CcAlg##Append(dctx, in, inl);                                   \
```
samuel-lee-msft, line 50: Should this return failure with in == NULL, and inl == 0?

I don't think we need either check here to be honest - just call SymCrypt#CcAlg##Append? This is closer to what we do in the Engine

mamckee: Took a look back at the default provider too. I don't see these parameters checked anywhere in that call stack, so it looks like this isn't the responsibility of the provider. I'll keep that in mind for future implementations too. That makes me a little concerned that the responsibility is assumed to lie on the caller, but if that's the normal pattern I'll make sure we follow it.

samuel-lee-msft: In general checking for NULL seems a bit redundant if it's something the caller passed - there are all kinds of addresses that can cause access violation if the caller is passing them.

If the pointer is something that is internal to the provider / OpenSSL, then erroring on NULL makes more sense - it indicates some bug with non-caller code, and the provider / OpenSSL should fail more gracefully to enable callers to diagnose and log the bug.

Checking for zero size should only be done if the SymCrypt API requires a non-zero size - otherwise we may be duplicating checks down the call stack and slowing down the expected case.

SymCryptProvider/src/p_scossl_digests.h
samuel-lee-msft: Looks good

SymCryptProvider/src/p_scossl_base.c
samuel-lee-msft: Yeah, I guess it could be beneficial to use the provided memory management functions to avoid circular dependencies. We currently have a problem with using statically linked libcrypto + dynamic SCOSSL engine, because the SCOSSL engine then dynamically loads libcrypto and there is symbol confusion.

If it is possible to make the provider only use OpenSSL functions it is initialized with, rather than general OpenSSL functions, then that could sidestep the issue neatly. Of course, currently in SCOSSL engine we use a lot more OpenSSL functions to fallback to base OpenSSL for functionality we don't support.

mamckee: Agreed, avoiding the circular dependencies would be nice. I don't know yet if it can be totally avoided, but this would be a start. We can always ask to add functions to the core dispatch table too if we absolutely need them and it would be reasonable to add. How about I put that in a separate PR then since this PR is already rather large?

samuel-lee-msft: Sounds good to me

samuel-lee-msft (approved):

PR #58, opened 2023-03-08
Cmake improvements

pemensik:
Some improvements in CMake for possible system-wide installed package.

Related to Fedora package review: https://bugzilla.redhat.com/show_bug.cgi?id=2162851


pemensik:
It seems that CLA bot does not work. Github does not auto-complete that user id and even manually it does not help. @microsoft-github-copilot is the only thing offered from microsoft-github prefix. I am doing the response wrong way?

samuel-lee-msft (approved):
LGTM

PR #60, opened 2023-03-20
Provider Ciphers and Rand

mamckee:
This PR adds cipher and RNG implementations to the Symcrypt Provider. The source files for the SymCrypt engine have all been prefixed with "e_" to avoid confusion with common source files.

The implementation for GCM and CCM has been refactored into common code shared between the provider and engine. 

The following ciphers have been added:
- AES-128-CBC
- AES-192-CBC
- AES-256-CBC
- AES-128-CFB (128-bit shift)
- AES-192-CFB (128-bit shift)
- AES-256-CFB (128-bit shift)
- AES-128-CFB (8-bit shift)
- AES-192-CFB (8-bit shift)
- AES-256-CFB (8-bit shift)
- AES-128-ECB
- AES-192-ECB
- AES-256-ECB
- AES-128-GCM
- AES-192-GCM
- AES-256-GCM
- AES-128-CCM
- AES-192-CCM
- AES-256-CCM
- AES-128-XTS (disabled for now)
- AES-256-XTS (disabled for now)


ScosslCommon/src/scossl_aes_aead.c
```diff
@@ -0,0 +1,754 @@
...
+
+    return SCOSSL_SUCCESS;
+}
+
+_Use_decl_annotations_
+SCOSSL_STATUS scossl_aes_gcm_get_aead_tag(SCOSSL_CIPHER_GCM_CTX *ctx, BOOL encrypt,
+                                          unsigned char *tag, size_t taglen)
+{
+    if (taglen < SCOSSL_GCM_MIN_TAG_LENGTH || taglen > SCOSSL_GCM_MAX_TAG_LENGTH ||
+        taglen > ctx->taglen || encrypt)
```
samuel-lee-msft, line 219: >encrypt [](http://example.com/codeflow?start=32&length=7)

!encrypt

ScosslCommon/inc/scossl_helpers.h
```diff
@@ -36,6 +36,10 @@ typedef _Return_type_success_(return >= 0) int SCOSSL_RETURNLENGTH; // For funct
 #define SCOSSL_LOG_LEVEL_INFO       (2) // DEFAULT for stderr / logging to logfile
 #define SCOSSL_LOG_LEVEL_DEBUG      (3)
 
+// Many parameters required by the engine and provider APIs are unused
+#
+#define _scossl_unused_
```
samuel-lee-msft, line 41: This isn't used anywhere as far as I can see?

ScosslCommon/src/scossl_aes_aead.c
```diff
@@ -0,0 +1,754 @@
...
+        if (scError != SYMCRYPT_NO_ERROR)
+        {
+            return SCOSSL_FAILURE;
+        }
+        ctx->keylen = keylen;
+    }
+    return SCOSSL_SUCCESS;
+}
+
+static SCOSSL_STATUS scossl_aes_gcm_tls(_Inout_ SCOSSL_CIPHER_GCM_CTX *ctx, BOOL encrypt,
```
samuel-lee-msft, line 61: >static SCOSSL_STATUS [](http://example.com/codeflow?start=0&length=20)

This was previously returning SCOSSL_RETURNLENGTH (i.e. length of out on success, or -1 on error). Is the change intended - are you sure this does not break OpenSSL 1.1.1?

samuel-lee-msft: Ah - I see, you're wrapping the common function scossl_aes_gcm_cipher in e_scossl_aes_gcm_cipher to tweak the return value.

ScosslCommon/inc/scossl_aes_aead.h
```diff
@@ -0,0 +1,144 @@
...
+    SYMCRYPT_GCM_STATE state;
+    SYMCRYPT_GCM_EXPANDED_KEY key;
+    SIZE_T keylen;
+    BYTE tag[EVP_GCM_TLS_TAG_LEN];
+    SIZE_T taglen;
+    BYTE tlsAad[EVP_AEAD_TLS1_AAD_LEN];
+    INT32 tlsAadSet;
+    UINT64 ivInvocation;
+    INT32 useInvocation;
+    BOOL encrypt;
```
samuel-lee-msft, line 59: >BOOL encrypt; [](http://example.com/codeflow?start=4&length=13)

Maybe note this is only used in the provider?
I've been using INT32 for booleans in structs elsewhere (I think I was following convention in OpenSSL here) - we have 3 of them already in this struct.

I don't have a huge preference either way, but it would be better to be consistent, either all BOOLs or all INT32s.

ScosslCommon/inc/scossl_aes_aead.h
```diff
@@ -0,0 +1,144 @@
...
+{
+    SCOSSL_CCM_STAGE_INIT = 0,   // The initial state
+    SCOSSL_CCM_STAGE_SET_CBDATA, // The state after a call providing the total input length
+    SCOSSL_CCM_STAGE_SET_AAD,    // The state after a call providing the AAD
+    SCOSSL_CCM_STAGE_COMPLETE,   // The state after a call providing the plain/ciphertext
+} SCOSSL_CCM_STAGE;
+
+typedef struct
+{
+    INT32 enc; /* COP_ENCRYPT or COP_DECRYPT */
```
samuel-lee-msft, line 86: >INT32 enc; /* COP_ENCRYPT or COP_DECRYPT */ [](http://example.com/codeflow?start=4&length=43)

There was a previous change to remove tracking enc from the engine structures, as it duplicated data kept in the EVP_CIPHER_CTX: [https://github.com/microsoft/SymCrypt-OpenSSL/commit/4534082f2040f34269fbb373d752e6521081fab1#diff-8360f0e71990cca77b2ad2bf9c73530ae51263784db4e2fc20e9909139fd5edb](https://github.com/microsoft/SymCrypt-OpenSSL/commit/4534082f2040f34269fbb373d752e6521081fab1#diff-8360f0e71990cca77b2ad2bf9c73530ae51263784db4e2fc20e9909139fd5edb)

Probably we can also remove SCOSSL_ENCRYPTION_MODE entirely, and just use (unused) int in all of the cipher *_init_key function declaration/definitions.

Seems I missed the enc field here in CCM, and it's unused in the engine. We only need one field in the provider.

ScosslCommon/src/scossl_aes_aead.c
```diff
@@ -0,0 +1,754 @@
...
+SCOSSL_STATUS scossl_aes_gcm_iv_gen(SCOSSL_CIPHER_GCM_CTX *ctx,
+                                    unsigned char *out, size_t outsize)
+{
+    if (ctx->useInvocation == 0)
+    {
+        return SCOSSL_FAILURE;
+    }
+    // Place invocation field into IV
+    SYMCRYPT_STORE_MSBFIRST64(ctx->iv + ctx->ivlen - EVP_GCM_TLS_EXPLICIT_IV_LEN, ctx->ivInvocation);
+    if (outsize <= 0 || outsize > ctx->ivlen)
```
samuel-lee-msft, line 252: ><= [](http://example.com/codeflow?start=16&length=2)

Maybe use == here - just as we're now comparing with unsigned size_t, <= looks like a mistake even if it's functionally fine.

ScosslCommon/src/scossl_aes_aead.c
```diff
@@ -0,0 +1,754 @@
...
+    return SCOSSL_SUCCESS;
+}
+
+_Use_decl_annotations_
+SCOSSL_STATUS scossl_aes_gcm_set_iv_inv(SCOSSL_CIPHER_GCM_CTX *ctx, BOOL encrypt,
+                                        unsigned char *iv, size_t ivlen)
+{
+    if (ctx->useInvocation == 0 ||
+        encrypt ||
+        ivlen > (size_t)ctx->ivlen)
```
samuel-lee-msft, line 310: >ivlen > (size_t)ctx->ivlen) [](http://example.com/codeflow?start=8&length=27)

We also used to fail with ivlen == 0, did you intend to remove this failure case?

ScosslCommon/inc/scossl_aes_aead.h
```diff
@@ -0,0 +1,144 @@
...
+    BYTE tag[EVP_CCM_TLS_TAG_LEN];
+    SIZE_T taglen;
+    UINT64 cbData;
+    BYTE tlsAad[EVP_AEAD_TLS1_AAD_LEN];
+    INT32 tlsAadSet;
+    BOOL encrypt;
+} SCOSSL_CIPHER_CCM_CTX;
+
+void scossl_aes_gcm_init_ctx(_Inout_ SCOSSL_CIPHER_GCM_CTX *ctx,
+                             size_t keylen,
```
samuel-lee-msft, line 102: >size_t keylen, [](http://example.com/codeflow?start=29&length=14)

I don't think we need this argument here - this init_ctx functions are only used by the Engine, and there we rely on OpenSSL to track the keylength anyway.

i.e. the key length can just be set at the point of init_key I think

ScosslCommon/src/scossl_aes_aead.c
```diff
@@ -0,0 +1,754 @@
...
+
+        ctx->ivlen = ivlen;
+        memcpy(ctx->iv, iv, ctx->ivlen);
+    }
+    if (key)
+    {
+        if (keylen != ctx->keylen)
+        {
+            return SCOSSL_FAILURE;
+        }
```
samuel-lee-msft, line 404: I don't think this should cause a failure - there's no particular reason we can't reuse a key context for different length keys

ScosslCommon/src/scossl_aes_aead.c
mamckee: Exactly, it kept the provider code a little cleaner. I'm writing any common functions more in line with the provider model since that's where the bulk of future changes will be.

ScosslCommon/inc/scossl_aes_aead.h
mamckee: What about SCOSSL_ENCRYPTION_MODE? I see the engine uses that as a parameter some places. I'll use INT32 for other booleans.

ScosslCommon/inc/scossl_aes_aead.h
mamckee: I'll remove the typedef and mark those parameters as unused. Will do a quick pass on e_scossl_ciphers.c to mark some of the other unused parameters.

samuel-lee-msft:
>     LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}/ossl-modules"

Probably a good plan to use a new variable like OPENSSL_PROVIDERS, like: [https://github.com/microsoft/SymCrypt-OpenSSL/pull/58/files#diff-f79d526d98446e8093575181d6dec16feb7cf2cd4c3882cc9bee910fa54fd5ce](https://github.com/microsoft/SymCrypt-OpenSSL/pull/58/files#diff-f79d526d98446e8093575181d6dec16feb7cf2cd4c3882cc9bee910fa54fd5ce)

---
Refers to: SymCryptProvider/CMakeLists.txt:50 in d5c545f. [](commit_id = d5c545f7c790636e5c7a99a21900a7dbbe004214, deletion_comment = False)

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -56,34 +51,37 @@ extern const OSSL_DISPATCH p_scossl_aes256gcm_functions[];
 extern const OSSL_DISPATCH p_scossl_aes128ccm_functions[];
 extern const OSSL_DISPATCH p_scossl_aes192ccm_functions[];
 extern const OSSL_DISPATCH p_scossl_aes256ccm_functions[];
-extern const OSSL_DISPATCH p_scossl_aes256xts_functions[];
-extern const OSSL_DISPATCH p_scossl_aes128xts_functions[];
+
+// Disabling XTS for now, also disabled in engine pending
+// the same dependencies.
+// Need to push the FIPS mode XTS flag into the SymCrypt API,
+// and expose an XTS key copy method. Addi
```
samuel-lee-msft, line 58: >Addi [](http://example.com/codeflow?start=38&length=4)

Nit: typo

SymCryptProvider/src/p_scossl_rand.c
```diff
@@ -0,0 +1,137 @@
...
+// Copyright (c) Microsoft Corporation. Licensed under the MIT license.
+//
+
+#include <openssl/core_dispatch.h>
+#include <openssl/core_names.h>
+#include <openssl/proverr.h>
+
+#include "scossl_helpers.h"
+
+#define SYMCRYPT_DRGB_STRENGTH 256
```
samuel-lee-msft, line 11: >SYMCRYPT_DRGB_STRENGTH [](http://example.com/codeflow?start=8&length=22)

Better to use a SCOSSL_ prefix for definitions here to avoid confusion with where to look for the definition.

ScosslCommon/inc/scossl_aes_aead.h
mamckee: IMPLEMENT_SCOSSL_AES_AEAD_CIPHER uses this and it needs to be set at the time of context creation in case the caller calls *_ctx_get_params. Given that it's only used by the provider thought I can move it out of the common code.

ScosslCommon/src/scossl_aes_aead.c
mamckee: This is consistent with the default OpenSSL behavior at least for providers. Given that I'm removing the keylen from *init_ctx thought I'll move this check to the provider.

SymCryptProvider/src/p_scossl_rand.c
```diff
@@ -0,0 +1,137 @@
...
+static void p_scossl_rand_freectx(ossl_unused void *ctx){}
+
+// RNG state is internally managed by SymCrypt. This function is
+// required, but does not actually instantiate SymCrypt's RNG state
+static SCOSSL_STATUS p_scossl_rand_instantiate(ossl_unused void *ctx,
+                                               ossl_unused unsigned int strength,
+                                               ossl_unused int prediction_resistance,
+                                               _In_reads_bytes_opt_(addin_len) const unsigned char *addin, size_t addin_len,
+                                               ossl_unused const OSSL_PARAM params[])
+{
```
samuel-lee-msft, line 44: Maybe this should return an error if the requested strength > 256 or prediction resistance is non-zero?

ScosslCommon/inc/scossl_aes_aead.h
samuel-lee-msft: Ahh - I see, I missed this because I searched for scossl_aes_gcm_init_ctx, rather than scossl_aes_.*_init_ctx. Fine to keep it in the common code.

ScosslCommon/src/scossl_aes_aead.c
samuel-lee-msft: Heads up that this failure is only currently in CCM, not GCM.

SymCryptProvider/src/p_scossl_rand.c
mamckee: There's no defined expected behavior but that makes sense to me.

SymCryptProvider/src/p_scossl_rand.c
```diff
@@ -0,0 +1,137 @@
...
+                                          _In_reads_bytes_opt_(addin_len) const unsigned char *addin, size_t addin_len)
+{
+    SymCryptProvideEntropy(addin, addin_len);
+    return SCOSSL_SUCCESS;
+}
+
+// SymCrypt internally mangages locking, so the provider does not need
+// to. This function is required but does nothing
+static SCOSSL_STATUS p_scossl_rand_enable_locking(ossl_unused void *ctx)
+{
```
samuel-lee-msft, line 88: Do we need to do anything for OSSL_FUNC_rand_lock / OSSL_FUNC_rand_unlock?

Given that SymCrypt always locks around accessing the RNG state, we would want these to always return success rather than taking any additional locks.

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -0,0 +1,539 @@
...
+    if (!ctx->cipher(ctx, out_int, &outl_int, SYMCRYPT_AES_BLOCK_SIZE, ctx->buf, ctx->cbBuf))
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_CIPHER_OPERATION_FAILED);
+        return SCOSSL_FAILURE;
+    }
+    ctx->cbBuf = 0;
+
+    if (ctx->pad)
+    {
+        SymCryptPaddingPkcs7Remove(
```
samuel-lee-msft, line 292: >SymCryptPaddingPkcs7Remove [](http://example.com/codeflow?start=8&length=26)

You are not checking the error return here!

Note that if SYMCRYPT_INVALID_ARGUMENT is returned we need to map it to SCOSSL_FAILURE in a side-channel safe way (i.e. using SymCryptMapUint32).

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -0,0 +1,539 @@
...
+    size_t cbBufRemaining = SYMCRYPT_AES_BLOCK_SIZE - *cbBuf;
+    if (inl < cbBufRemaining)
+    {
+        cbBufRemaining = inl;
+    }
+    memcpy(buf + *cbBuf, in, cbBufRemaining);
+
+    // Advance in
+    *cbBuf += cbBufRemaining;
+    in += cbBufRemaining;
```
samuel-lee-msft, line 69: >in += cbBufRemaining; [](http://example.com/codeflow?start=4&length=21)

I think this doesn't do what you intend here - this is just modifying the local copy of in rather than in from p_scossl_aes_generic_update.

I think it would be clearer to:
a) pass cbBuf by value as a const size_t
b) rename cbBufRemaining to something like cbBytesBuffered
c) in p_scossl_aes_generic_update, set a variable to the returned cbBytesBuffered, and use it to update in and ctx->cbBuf

Or you could manually inline the function

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -0,0 +1,539 @@
...
+        if (!ctx->cipher(ctx, out, NULL, outsize, ctx->buf, ctx->cbBuf))
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_CIPHER_OPERATION_FAILED);
+            return SCOSSL_FAILURE;
+        }
+
+        out += SYMCRYPT_AES_BLOCK_SIZE;
+        outl_int += SYMCRYPT_AES_BLOCK_SIZE;
+        ctx->cbBuf = 0;
+    }
```
samuel-lee-msft, line 183: We should have an invariant that ctx->cbBuf is in the range [0, SYMCRYPT_AES_BLOCK_SIZE-1] outside of the functions which modify ctx->buf.

Then this block should trigger iff `ctx->cbBuf > 0` on line 166, and we then filled the buffer. So we can move a branch off the expected path (I expect most callers will update with a multiple of block size) by moving this block into the block above.

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -0,0 +1,539 @@
...
+        else if (ctx->cbBuf == 0)
+        {
+            *outl = 0;
+            return SCOSSL_SUCCESS;
+        }
+        else if (ctx->cbBuf != SYMCRYPT_AES_BLOCK_SIZE)
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_WRONG_FINAL_BLOCK_LENGTH);
+            return SCOSSL_FAILURE;
+        }
```
samuel-lee-msft, line 255: I think if ctx->cbBuf == SYMCRYPT_AES_BLOCK_SIZE this is also a failure condition, so I would make this be an unconditional else

The two happy paths should be:
a) padding is enabled, and buf has 0-15 bytes in it, so we pad buf to 16 bytes and encrypt
b) padding is disabled and buf is empty, so we return nothing

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -0,0 +1,539 @@
...
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_WRONG_FINAL_BLOCK_LENGTH);
+        return SCOSSL_FAILURE;
+    }
+
+    unsigned char out_int[SYMCRYPT_AES_BLOCK_SIZE];
+    size_t outl_int;
+
+    // Decrypt
+    if (!ctx->cipher(ctx, out_int, &outl_int, SYMCRYPT_AES_BLOCK_SIZE, ctx->buf, ctx->cbBuf))
```
samuel-lee-msft, line 283: >ctx->buf, ctx->cbBuf [](http://example.com/codeflow?start=71&length=20)

Hmm, for this to work we need to change the update function to not return the last decrypted block eagerly; with the current setup there's no guarantee there's any data in buf when we get here.

I guess in the decrypt flow with padding buf can be in 1 of 3 states:
a) empty at initialization
b) contain 1-15 bytes of partial block
c) contain 16 bytes of the last block (could be decrypted in update, or delay to decrypt until here)
when we get to a final call with padding we fail if we are not in state c.

Worth documenting a bit what the different flows are in SCOSSL_AES_CTX's definition

SymCryptProvider/src/ciphers/p_scossl_aes.c
mamckee: Missed this in refactoring. I thought I might need this code elsewhere but that isn't the case, so I'll inline this again.

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -0,0 +1,539 @@
...
+        return SCOSSL_FAILURE;
+    }
+    p = OSSL_PARAM_locate(params, OSSL_CIPHER_PARAM_BLOCK_SIZE);
+    if (p != NULL && !OSSL_PARAM_set_size_t(p, block_size))
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_SET_PARAMETER);
+        return SCOSSL_FAILURE;
+    }
+    p = OSSL_PARAM_locate(params, OSSL_CIPHER_PARAM_AEAD);
+    if (p != NULL && !OSSL_PARAM_set_int(p, flags & SCOSSL_FLAG_AEAD))
```
samuel-lee-msft, line 370: >flags & SCOSSL_FLAG_AEAD [](http://example.com/codeflow?start=44&length=24)

My understanding here is that we should be returning 1 if flags & SCOSSL_FLAG_AEAD is SCOSSL_FLAG_AEAD and 0 otherwise?

For SCOSSL_FLAG_AEAD this works now, as SCOSSL_FLAG_AEAD == 1, but for SCOSSL_FLAG_CUSTOM_IV we're currently returning 2 or 0 I think?

Might be better to have ternary expressions here to make the return value 1 or 0, regardless of the definition of SCOSSL_FLAG_*

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -0,0 +1,539 @@
...
+            ERR_raise(ERR_LIB_PROV, PROV_R_CIPHER_OPERATION_FAILED);
+            return SCOSSL_FAILURE;
+        }
+
+        out += SYMCRYPT_AES_BLOCK_SIZE;
+        outl_int += SYMCRYPT_AES_BLOCK_SIZE;
+        ctx->cbBuf = 0;
+    }
+
+    cBlocksRemaining = inl / SYMCRYPT_AES_BLOCK_SIZE;
```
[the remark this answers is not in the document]
mamckee: Since the block size is a power of 2 `cbBytesInFullBlocks = inl & ~(SYMCRYPT_AES_BLOCK_SIZE-1);` will get the same result. It looks like the default provider does this too.

SymCryptProvider/src/ciphers/p_scossl_aes.c
mamckee: This is the decrypt case. I'll add a comment higher up to indicate that better. In this case we either expect an empty buffer or a full buffer with the last block. I realize looking back at update we need to ensure the last block is left in the buffer in case it's padded.

Looking at the EVP tests, all the plaintexts are a length multiple of 16 so this wasn't caught. I'll update those to properly test the padding.

SymCryptProvider/src/ciphers/p_scossl_aes.c
mamckee: OpenSSL just checks non-zero but I agree this would be cleaner.

SymCryptProvider/src/ciphers/p_scossl_aes.c
samuel-lee-msft: I think this comment was on the encrypt case (lines 251-255), where I still think it makes sense to only have the 2 success paths I described.

In the case of encrypt w/ padding, if you allow cbBuf == 16 on entry to the final call, then SymCryptPaddingPkcs7Add will fail (there's no space for padding).
So I think it makes sense, in encryption, to guarantee cbBuf in [0, 15] after a call to update.
In the final call, with padding we pad the contents of buf unconditionally. Without padding we fail if cbBuf != 0.

On decrypt, I think the behavior depends on padding.
Without padding, the decryption should be done eagerly, so, like encryption, cbBuf should be in [0, 15] after an update call, and in a final call we fail if cbBuf != 0.
With padding, we need to remove the padding from the last block before returning to the caller, so we need cbBuf to be in [1,16] after an update call. Now in a final call we fail if cbBuf != 16.

SymCryptProvider/src/ciphers/p_scossl_aes_aead.c
```diff
@@ -0,0 +1,546 @@
...
+extern "C" {
+#endif
+
+static const OSSL_PARAM p_scossl_aes_gcm_gettable_ctx_param_types[] = {
+    OSSL_PARAM_size_t(OSSL_CIPHER_PARAM_KEYLEN, NULL),
+    OSSL_PARAM_size_t(OSSL_CIPHER_PARAM_AEAD_IVLEN, NULL),
+    OSSL_PARAM_size_t(OSSL_CIPHER_PARAM_AEAD_TAGLEN, NULL),
+    OSSL_PARAM_octet_string(OSSL_CIPHER_PARAM_IV, NULL, 0),
+    OSSL_PARAM_octet_string(OSSL_CIPHER_PARAM_UPDATED_IV, NULL, 0),
+    OSSL_PARAM_octet_string(OSSL_CIPHER_PARAM_AEAD_TAG, NULL, 0),
```
samuel-lee-msft, line 22: I think we might also need OSSL_CIPHER_PARAM_AEAD_TLS1_AAD_PAD here and in CCM to enable EVP_CTRL_AEAD_TLS1_AAD with the provider.

It should be fine to return ctx->taglen.

SymCryptProvider/src/ciphers/p_scossl_aes_aead.c
```diff
@@ -0,0 +1,546 @@
...
+static SCOSSL_CIPHER_CCM_CTX *p_scossl_aes_ccm_dupctx(_In_ SCOSSL_CIPHER_CCM_CTX *ctx)
+{
+    SCOSSL_CIPHER_CCM_CTX *copy_ctx = OPENSSL_malloc(sizeof(SCOSSL_CIPHER_CCM_CTX));
+    if (copy_ctx != NULL)
+    {
+        memcpy(copy_ctx, ctx, sizeof(SCOSSL_CIPHER_CCM_CTX));
+
+        SymCryptAesKeyCopy(&ctx->key, &copy_ctx->key);
+        copy_ctx->state = ctx->state;
+        copy_ctx->state.pExpandedKey = &ctx->key;
```
samuel-lee-msft, line 292: Here we are reaching into SymCrypt state in a way which isn't ideal - would be good to keep a comment here that we would like to use SymCryptCcmStateCopy for this.

mamckee: Looking at OpenSSL behavior this should be
`ctx->tlsAadSet ? EVP_GCM_TLS_TAG_LEN : 0` for gcm
`ctx->tlsAadSet ? ctx->taglen : 0` for ccm

samuel-lee-msft: I think you meant that reply to go on the OSSL_CIPHER_PARAM_AEAD_TLS1_AAD_PAD comment, but from [the documentation for OSSL_CIPHER_PARAM_AEAD_TLS1_AAD_PAD](https://www.openssl.org/docs/man3.1/man3/EVP_CIPHER_CTX_flags.html#:~:text=%22-,tlsaadpad,-%22%20(OSSL_CIPHER_PARAM_AEAD_TLS1_AAD_PAD)%20%3Cunsigned), I see don't see that you _need_ to return 0 when `ctx->tlsAadSet` is not set, but fine to implement either way.

SymCryptProvider/src/p_scossl_rand.c
samuel-lee-msft: I think you got the check exactly backwards - we should fail if the caller requests prediction resistance (not supported) or asks for a strength which is higher than we support.

SymCryptProvider/src/ciphers/p_scossl_aes.c
samuel-lee-msft: This code could still be in the preceding block (we only check whether apply the cipher to the buffer if we already successfully checked it's not empty).

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -0,0 +1,556 @@
...
+    }
+
+    // Get the remaining number of whole blocks in inl
+    cbBytesInFullBlocks = inl & ~(SYMCRYPT_AES_BLOCK_SIZE-1);
+
+    // Decrypt with padding. Ensure the last block is buffered
+    // for the call to cipher_final so padding is removed
+    if (!ctx->encrypt && ctx->pad && cbBytesInFullBlocks * SYMCRYPT_AES_BLOCK_SIZE == inl)
+    {
+        cbBytesInFullBlocks--;
```
samuel-lee-msft, line 173: >cbBytesInFullBlocks-- [](http://example.com/codeflow?start=8&length=21)

cbBytesInFullBlock -= SYMCRYPT_AES_BLOCK_SIZE

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -0,0 +1,556 @@
...
+        outl_int += SYMCRYPT_AES_BLOCK_SIZE;
+        ctx->cbBuf = 0;
+    }
+
+    // Get the remaining number of whole blocks in inl
+    cbBytesInFullBlocks = inl & ~(SYMCRYPT_AES_BLOCK_SIZE-1);
+
+    // Decrypt with padding. Ensure the last block is buffered
+    // for the call to cipher_final so padding is removed
+    if (!ctx->encrypt && ctx->pad && cbBytesInFullBlocks * SYMCRYPT_AES_BLOCK_SIZE == inl)
```
samuel-lee-msft, line 171: >cbBytesInFullBlocks * SYMCRYPT_AES_BLOCK_SIZE == inl [](http://example.com/codeflow?start=37&length=52)

cbBytesInFullBlocks == inl

samuel-lee-msft: We also need to check cbBytesInFullBlocks > 0

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -0,0 +1,556 @@
...
+    if (cbBytesInFullBlocks > 0)
+    {
+        outl_int += cbBytesInFullBlocks;
+        // Need to verify outsize here is not smaller than the 
+        // result of the entire update operation
+        if (outsize < outl_int)
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_OUTPUT_BUFFER_TOO_SMALL);
+            return SCOSSL_FAILURE;
+        }
```
samuel-lee-msft, line 186: I think it's cleaner to let the cipher function handle the output buffer being too small.

i.e. We subtract SYMCRYPT_AES_BLOCK_SIZE from outsize above when we output from cbBuf to out. The subtraction can't underflow as otherwise we would fail in the cipher function with buffer too small.
Then here we let the cipher function fail if the updated outsize is too small, rather than doing an explicit check.

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -0,0 +1,556 @@
...
+            scError = SYMCRYPT_NO_ERROR;
+            goto cleanup;
+        }
+        else
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_WRONG_FINAL_BLOCK_LENGTH);
+            goto cleanup;
+        }
+
+        if (!ctx->cipher(ctx, out, outl, outsize, ctx->buf, ctx->cbBuf))
```
samuel-lee-msft, line 271: >ctx->cbBuf [](http://example.com/codeflow?start=60&length=10)

SYMCRYPT_AES_BLOCK_SIZE

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -0,0 +1,556 @@
...
+    size_t outl_int;
+
+    // The return value of SymCryptPaddingPkcs7Remove must be mapped in a
+    // side-channel safe way with SymCryptMapUint32. For all other failure
+    // cases scError is just set to 1 so it maps to SCOSSL_FAILURE
+    SYMCRYPT_ERROR scError = 1;
+    SYMCRYPT_UINT32_MAP scErrorMap[1] = {
+        {SYMCRYPT_NO_ERROR, SCOSSL_SUCCESS}};
+
+    // Encrypt final
```
samuel-lee-msft, line 238: Probably clearer if we refactor to handle the no padding case first. For both encrypt and decrypt if !ctx->pad return success and set outl to 0 if cbBuf == 0 and fail otherwise.

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -0,0 +1,556 @@
...
+            &outl_int);
+    }
+
+    if (outsize < outl_int)
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_OUTPUT_BUFFER_TOO_SMALL);
+        goto cleanup;
+    }
+
+    memcpy(out, out_int, outl_int);
```
samuel-lee-msft, line 317: Sorry I wasn't clear about the constant-time requirement here / should have picked up on this before.

We can't raise an error or branch based on outl_int, and the memcpy here is not constant time w.r.t outl_int.

I think on decrypt with padding we must unconditionally fail if outsize < SYMCRYPT_AES_BLOCK_SIZE (does not need to be constant time and can raise an err), then in the SymCryptPaddingPkcs7Remove, we write straight to out with cbDst = SYMCRYPT_AES_BLOCK_SIZE, and pcbResult = outl.

It's not clear to me whether OpenSSL will use this correctly in TLS in a constant-time fashion though - we need to dig into how it's wired up in TLS in OpenSSL 3

SymCryptProvider/src/ciphers/p_scossl_aes.c
mamckee: Ah right, inl could be 0 at this point

SymCryptProvider/src/ciphers/p_scossl_aes.c
mamckee: I see. That makes sense to me.

SymCryptProvider/src/ciphers/p_scossl_aes.c
samuel-lee-msft: I think `ctx->cbBuf == SYMCRYPT_AES_BLOCK_SIZE` on line 152 if and only if `ctx->cbBuf > 0` on line 134. So the whole if statement 152-164 can move into the block 134-147?

We still need to check the conditions in 152 and 153 before calling ctx->cipher, it's just we do less branching on the path when the buffer is empty.

mamckee:
@samuel-lee-msft, I added CFB to this PR. SymCrypt only supports 128- and 8-bit shifts, but OpenSSL also supports 1-bit shifts registered as separate algorithms. Is this something we need to support in the SymCrypt provider or can we fall back to the default provider?

samuel-lee-msft:
Fine to fallback here - but when we update the README it should be clear.

---
In reply to: [1500728890](https://github.com/microsoft/SymCrypt-OpenSSL/pull/60#issuecomment-1500728890) [](http://example.com/codeflow?ancestors=1500728890,1500728890)

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -0,0 +1,590 @@
...
+{
+    return p_scossl_aes_generic_init_internal(ctx, 0, key, keylen, iv, ivlen, params);
+}
+
+static SCOSSL_STATUS p_scossl_aes_generic_update(_Inout_ SCOSSL_AES_CTX *ctx,
+                                                 _Out_writes_bytes_(*outl) unsigned char *out, _Out_ size_t *outl, size_t outsize,
+                                                 _In_reads_bytes_(inl) const unsigned char *in, size_t inl)
+{
+    size_t cbBytesInFullBlocks = 0;
+    *outl = 0;
```
samuel-lee-msft, line 130: >*outl = 0; [](http://example.com/codeflow?start=4&length=10)

If this function is called with inl == 0 I think we should return success unconditionally having not done anything.

I think currently we will fail if it's the first call in a padded decryption, as cbBuf will remain 0. Probably simplest to just check for the inl==0 case at the start and return in this case.

SymCryptProvider/src/p_scossl_rand.c
```diff
@@ -0,0 +1,142 @@
...
+// Copyright (c) Microsoft Corporation. Licensed under the MIT license.
+//
+
+#include <openssl/core_dispatch.h>
+#include <openssl/core_names.h>
+#include <openssl/proverr.h>
+
+#include "scossl_helpers.h"
+
+#define SCOSSL_DRGB_STRENGTH 256
```
samuel-lee-msft, line 11: >DRGB [](http://example.com/codeflow?start=15&length=4)

typo across this file: DRGB should be DRBG

samuel-lee-msft (commented):
I think this is looking good to merge other than for the couple of comments - I'll perform some local testing next week and approve.

SymCryptProvider/src/ciphers/p_scossl_aes.c
mamckee: Agreed. It's not likely we'll encounter this scenario but it at least saves some cycles checking here rather than at the end.

SymCryptProvider/src/ciphers/p_scossl_aes_aead.c
```diff
@@ -0,0 +1,569 @@
...
+                                                                                                             \
+        return ctx;                                                                                          \
+    }                                                                                                        \
+    SCOSSL_STATUS p_scossl_aes_##kbits##_##lcmode##_get_params(_Inout_ OSSL_PARAM params[])                  \
+    {                                                                                                        \
+        return p_scossl_aes_generic_get_params(params, EVP_CIPH_##UCMODE##_MODE, kbits >> 3,                 \
+                                               ivlen, 1, SCOSSL_FLAG_AEAD | SCOSSL_FLAG_CUSTOM_IV);          \
+    }                                                                                                        \
+                                                                                                             \
+    const OSSL_DISPATCH p_scossl_aes##kbits####lcmode##_functions[] = {                                      \
```
samuel-lee-msft, line 542: p_scossl_aes##kbits##_##lcmode##_functions

samuel-lee-msft (dismissed):
I have run these changes against 1.1.1 to check the engine is still looking good, and compiled the provider and engine against 3.2.

Found one compile-time issue in p_scossl_aes_aead.c.

Need to do more work to test the provider against 3.x, but I think we can go ahead with merging the changes, with the one fix.

samuel-lee-msft (approved):

PR #61, opened 2023-03-23
Use dispatch table passed to provider

mamckee:
The SymCrypt provider receives a dispatch table for upcalls in its init function. The provider should use these functions to avoid a circular dependency on libcrypto.


samuel-lee-msft (approved):
LGTM

Issue #59, opened 2023-03-14
Is anywhere documented memory alignment requirements used with symcryptengine?


pemensik:
We have a customer, who tried to use SymCrypt and this engine to implement FIPS 140-2 like compliance. However they hit a crashes with some software, for example bind-export-libs and dhcp client.

I have tried reproducing it by using the engine with named service. It crashes early in the start, because it uses [MOVAPS instructions](https://www.felixcloutier.com/x86/movaps) requiring 16B alignment. Normal OpenSSL does not have any similar requirement and typical application does not have a need for alignment checks. I think that is important difference from default OpenSSL engine, be it on Windows platform or Linux. Is such requirement documented anywhere?

I think such requirement prevents the use of *symcryptengine* as a system wide default. Is it possible to build version without such limitation? Ie. which would accept also non-aligned memory buffers? I had expected some requirements for RSA keys and such, but requirement of digest functions is a bit surprising.

```
(gdb) bt
#0  0x00007ffff7f2b91d in SymCryptMd5AppendBlocks (pChain=0x5555558a6908, pbData=0x5555558a68c8 "\200", cbData=64, pcbRemaining=0x7fffffffdfb0) at /__w/1/s/lib/md5.c:388
#1  0x00007ffff7f26839 in SymCryptHashCommonPaddingMd4Style (pHash=pHash@entry=0x7ffff7f79880 <SymCryptMd5Algorithm_default>, pState=pState@entry=0x5555558a68a8) at /__w/1/s/lib/hash.c:125
#2  0x00007ffff7f2b1dc in SymCryptMd5Result (pState=0x5555558a68a8, pbResult=0x7fffffffe060 "(\341\212UUU") at /__w/1/s/lib/md5.c:221
#3  0x00007ffff1e22e19 in scossl_digest_md5_final () from /usr/lib64/symcryptengine.so
#4  0x00007ffff5fae808 in EVP_DigestFinal_ex (ctx=0x5555558b25f8, md=0x7fffffffe060 "(\341\212UUU", size=0x0) at crypto/evp/digest.c:199
#5  0x00007ffff5fae882 in EVP_DigestFinal (ctx=0x5555558b25f8, md=<optimized out>, size=<optimized out>) at crypto/evp/digest.c:185
#6  0x00007ffff63667cd in do_detect_available () from /lib64/libisc.so.1107
[...]

PR #62, opened 2023-05-18
Ensure that SymCrypt structures have sufficient alignment

samuel-lee-msft:
+ Previously SCOSSL would rely on allocations made by libcrypto being at least 16-byte aligned for 64-bit platforms. This was normally true, but in some environments where the allocator is modified, only 8-byte alignment was being provided.
+ Should address #59

### DIFF
--- a/ScosslCommon/inc/scossl_helpers.h
+++ b/ScosslCommon/inc/scossl_helpers.h
@@ -51,6 +51,9 @@ typedef _Return_type_success_(return >= 0) int SCOSSL_RETURNLENGTH; // For funct
 #define SCOSSL_ALIGNED_SIZEOF(typename)         (sizeof(typename) + SYMCRYPT_ALIGN_VALUE)
 #define SCOSSL_ALIGN_UP(ptr)                    (SYMCRYPT_ALIGN_UP(ptr))
 
+// We need to be able to represent the offset into our allocation in a single byte
+C_ASSERT( SYMCRYPT_ALIGN_VALUE < 256 );
+
 #define SCOSSL_COMMON_ALIGNED_ALLOC(ptr, allocator, typename)               \
     typename *ptr;                                                          \
     {                                                                       \
--- a/ScosslCommon/inc/scossl_helpers.h
+++ b/ScosslCommon/inc/scossl_helpers.h
@@ -51,20 +51,26 @@ typedef _Return_type_success_(return >= 0) int SCOSSL_RETURNLENGTH; // For funct
 #define SCOSSL_ALIGNED_SIZEOF(typename)         (sizeof(typename) + SYMCRYPT_ALIGN_VALUE)
 #define SCOSSL_ALIGN_UP(ptr)                    (SYMCRYPT_ALIGN_UP(ptr))
 
-#define SCOSSL_COMMON_ALIGNED_ALLOC(ptr, allocator, typename)   \
-    PBYTE alloc = allocator(SCOSSL_ALIGNED_SIZEOF(typename));   \
-    PBYTE aligned_alloc = NULL;                                 \
-    if (alloc)                                                  \
-    {                                                           \
-        aligned_alloc = SCOSSL_ALIGN_UP(alloc+1);               \
-        *(aligned_alloc - 1) = aligned_alloc - alloc;           \
-    }                                                           \
-    typename *ptr = (typename *) aligned_alloc;
+#define SCOSSL_COMMON_ALIGNED_ALLOC(ptr, allocator, typename)               \
+    typename *ptr;                                                          \
+    {                                                                       \
+        PBYTE scossl_alloc = allocator(SCOSSL_ALIGNED_SIZEOF(typename));    \
+        PBYTE scossl_aligned = NULL;                                        \
+        if (scossl_alloc)                                                   \
+        {                                                                   \
+            scossl_aligned = SCOSSL_ALIGN_UP(scossl_alloc+1);               \
+            *(scossl_aligned - 1) = scossl_aligned - scossl_alloc;          \
+        }                                                                   \
+        ptr = (typename *) scossl_aligned;                                  \
+    }
 
-#define SCOSSL_COMMON_ALIGNED_FREE(ptr, deallocator, typename)  \
-    PBYTE aligned_alloc = (PBYTE) ptr;                          \
-    PBYTE alloc = aligned_alloc - *(aligned_alloc-1);           \
-    deallocator(alloc, SCOSSL_ALIGNED_SIZEOF(typename));
+#define SCOSSL_COMMON_ALIGNED_FREE(ptr, deallocator, typename)      \
+    {                                                               \
+        PBYTE scossl_aligned = (PBYTE) ptr;                         \
+        PBYTE scossl_alloc = scossl_aligned - *(scossl_aligned-1);  \
+        deallocator(scossl_alloc, SCOSSL_ALIGNED_SIZEOF(typename)); \
+        ptr = NULL;                                                 \
+    }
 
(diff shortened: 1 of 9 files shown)


ScosslCommon/inc/scossl_helpers.h
```diff
@@ -36,6 +36,36 @@ typedef _Return_type_success_(return >= 0) int SCOSSL_RETURNLENGTH; // For funct
...
+    PBYTE aligned_alloc = NULL;                                 \
+    if (alloc)                                                  \
+    {                                                           \
+        aligned_alloc = SCOSSL_ALIGN_UP(alloc+1);               \
+        *(aligned_alloc - 1) = aligned_alloc - alloc;           \
+    }                                                           \
+    typename *ptr = (typename *) aligned_alloc;
+
+#define SCOSSL_COMMON_ALIGNED_FREE(ptr, deallocator, typename)  \
+    PBYTE aligned_alloc = (PBYTE) ptr;                          \
```
mlindgren, line 65: >aligned_alloc [](http://example.com/codeflow?start=10&length=13)

Probably not a problem given the way it's used, but this shadows the standard library [aligned_alloc](https://en.cppreference.com/w/c/memory/aligned_alloc).

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -249,22 +249,7 @@ SCOSSL_STATUS OSSL_provider_init(_In_ const OSSL_CORE_HANDLE *handle,
                                  _Out_ const OSSL_DISPATCH **out,
                                  _Out_ void **provctx)
 {
-    SCOSSL_PROVCTX *p_ctx = OPENSSL_malloc(sizeof(SCOSSL_PROVCTX));
-    if (p_ctx != NULL)
-    {
-        p_ctx->handle = handle;
-        *provctx = p_ctx;
-    }
```
mlindgren, line 257: Just curious, why was this moved?

mlindgren (dismissed):
:shipit:

SymCryptProvider/src/p_scossl_base.c
samuel-lee-msft: The intent is to use the malloc function we are provided in the `in` dispatch table to allocate the provider context; rather than maybe accidentally linking libcrypto for the call to OPENSSL_malloc.

samuel-lee-msft: @mamckee does this change make sense to you?

ScosslCommon/inc/scossl_helpers.h
samuel-lee-msft: Hrmm, that's a fair point - I'll rename to avoid ambiguity.

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -541,7 +541,7 @@ static SCOSSL_STATUS scossl_aes_cfb8_cipher(_Inout_ SCOSSL_AES_CTX *ctx,
 #define IMPLEMENT_SCOSSL_AES_BLOCK_CIPHER(kbits, ivlen, lcmode, UCMODE)                                   \
     SCOSSL_AES_CTX *p_scossl_aes_##kbits##_##lcmode##_newctx()                                            \
     {                                                                                                     \
-        SCOSSL_AES_CTX *ctx = OPENSSL_zalloc(sizeof(SCOSSL_AES_CTX));                                     \
+        SCOSSL_COMMON_ALIGNED_ALLOC(ctx, OPENSSL_malloc, SCOSSL_AES_CTX);                                 \
```
ccalik, line 544: OPENSSL_zalloc

SymCryptProvider/src/ciphers/p_scossl_aes_xts.c
```diff
@@ -38,8 +38,8 @@ static SCOSSL_STATUS p_scossl_aes_xts_set_ctx_params(_Inout_ SCOSSL_AES_XTS_CTX
 
 static SCOSSL_AES_XTS_CTX *p_scossl_aes_xtx_newctx_internal(size_t keylen)
 {
-    SCOSSL_AES_XTS_CTX *ctx = OPENSSL_zalloc(sizeof(SCOSSL_AES_XTS_CTX));
-    if (ctx != NULL)
+    SCOSSL_COMMON_ALIGNED_ALLOC(ctx, OPENSSL_malloc, SCOSSL_AES_XTS_CTX);
```
ccalik, line 41: OPENSSL_zalloc

SymCryptProvider/src/ciphers/p_scossl_aes_xts.c
```diff
@@ -48,7 +48,7 @@ static SCOSSL_AES_XTS_CTX *p_scossl_aes_xtx_newctx_internal(size_t keylen)
 
 static SCOSSL_AES_XTS_CTX *p_scossl_aes_xts_dupctx(SCOSSL_AES_XTS_CTX *ctx)
 {
-    SCOSSL_AES_XTS_CTX *copy_ctx = OPENSSL_zalloc(sizeof(SCOSSL_AES_XTS_CTX));
+    SCOSSL_COMMON_ALIGNED_ALLOC(copy_ctx, OPENSSL_malloc, SCOSSL_AES_XTS_CTX);
```
ccalik, line 51: OPENSSL_zalloc

ccalik (dismissed):

SymCryptProvider/src/ciphers/p_scossl_aes.c
samuel-lee-msft: Ah - I had intentionally made the change to be more consistent with AEAD versions; there's no security reason we need to zero before use and it seemed to be a waste of cycles.

However, we don't initialize cbBuf without the zalloc. We could explicitly initialize cbBuf to 0, or just zero the whole ctx as before.
Probably it's better to use zalloc in all newctx functions and malloc in the dupctx just to be consistent.

samuel-lee-msft: Hmm, when I went to make the change I decided it was more consistent to just ensure that everything that needs to be initialized is initialized. I believe only cbBuf was not being set.

SymCryptProvider/src/ciphers/p_scossl_aes_xts.c
samuel-lee-msft: Intentionally moving to malloc for consistency - there is no need for zero-ing

SymCryptProvider/src/ciphers/p_scossl_aes_xts.c
samuel-lee-msft: Intentionally moving to malloc for consistency - there is no need for zero-ing

ScosslCommon/inc/scossl_helpers.h
samuel-lee-msft: Fixed

SymCryptProvider/src/p_scossl_base.c
mamckee: Yes, we don't need the provider handle to map these functions, so moving the function table fetch to the top makes sense.

SymCryptProvider/src/ciphers/p_scossl_aes.c
mamckee: It gets explicitly set in encrypt/decrypt init, but we should also zero it here. p_scossl_aes_generic_update assumes this value is between 0 and 15. The caller should call encrypt/decrypt init but if they don't, we might get an access violation rather than a simple error from p_scossl_aes_generic_update.

mamckee (dismissed):

mlindgren (approved):
:shipit:

ccalik (approved):

PR #63, opened 2023-06-14
Provider RSA signature and encryption

mamckee:
- Added RSA key management interface to provider
- Added RSA signature interface to provider
  - PKCS1
  - PSS
- Added RSA asym cipher interface to provider
  - PKCS1
  - OAEP
  - No padding
- Refactored existing signature and encryption code from engine into common code


ScosslCommon/src/scossl_rsa.c
```diff
@@ -0,0 +1,959 @@
...
+#define SCOSSL_SHA512_DIGEST_LENGTH   (64)
+
+typedef struct
+{
+    PCSYMCRYPT_OID pHashOIDs;
+    SIZE_T         nOIDCount;
+    UINT32         flags;
+} SCOSSL_RSA_PKCS1_PARAMS;
+
+static const SCOSSL_RSA_PKCS1_PARAMS scossl_rsa_pkcs1_sha1md5_params  = {NULL, 0, SYMCRYPT_FLAG_RSA_PKCS1_NO_ASN1};
```
samuel-lee-msft, line 31: >sha1md5 [](http://example.com/codeflow?start=54&length=7)

NP: md5sha1 seems more consistent

ScosslCommon/src/scossl_rsa.c
```diff
@@ -0,0 +1,959 @@
...
+    }
+
+    if (cbDst == (SIZE_T)-1)
+    {
+        // cbDst is not caller supplied for engine
+        cbDst = cbModulus;
+    }
+    else if (cbDst < cbModulus)
+    {
+        goto cleanup;
```
samuel-lee-msft, line 524: >goto cleanup; [](http://example.com/codeflow?start=8&length=13)

Looks like *pcbDst is not set for this path, so the return value from the function will depend on *pcbDst 's uninitialized value?

ScosslCommon/src/scossl_rsa.c
```diff
@@ -0,0 +1,959 @@
...
+    default:
+        SCOSSL_LOG_INFO(SCOSSL_ERR_F_RSA_PUB_ENC, SCOSSL_ERR_R_OPENSSL_FALLBACK,
+                        "Unsupported Padding: %d.", padding);
+        break;
+    }
+
+    *pcbDst = (cbResult <= INT_MAX) ? (INT32)cbResult : -1;
+
+cleanup:
+    return *pcbDst >= 0;
```
samuel-lee-msft, line 615: >return *pcbDst >= 0; [](http://example.com/codeflow?start=4&length=20)

Generally I think it's cleaner to use a local variable to set the return value, rather than reading from caller provided memory which could theoretically change (harder to reason about)

ScosslCommon/src/scossl_rsa.c
```diff
@@ -0,0 +1,959 @@
...
+    scError = SymCryptRsaPssSign(
+        keyCtx->key,
+        pbHashValue,
+        cbHashValue,
+        scossl_mac_algo,
+        cbSalt,
+        0,
+        SYMCRYPT_NUMBER_FORMAT_MSB_FIRST,
+        pbSignature,
+        pcbSignature != NULL ? (*pcbSignature) : 0,
```
samuel-lee-msft, line 389: >pcbSignature != NULL ? (*pcbSignature) : 0, [](http://example.com/codeflow?start=8&length=43)

This construct has been here since I wrote the original PSS implementation, but I'm not sure it makes sense looking at it now.

We only get here if pbSignature is non-NULL, and trying to sign into a zero-length signature will always result in a SYMCRYPT_INVALID_ARGUMENT in SymCrypt.

My best guess for why it's like this is that originally I did not check pbSignature != NULL before calling into SymCrypt (so we wanted to accept pbSignature==NULL and pcbSignature==0).


I would rephrase this as:
if( pcbSignature == NULL )
    goto cleanup;

after the check for pbSignature == NULL above, and unconditionally pass *pcbSignature here.

ScosslCommon/src/scossl_rsa.c
```diff
@@ -0,0 +1,959 @@
...
+    cbResult = SymCryptRsakeySizeofModulus(keyCtx->key);
+    if (pcbSignature != NULL)
+    {
+        *pcbSignature = cbResult;
+    }
+
+    if (pbSignature == NULL)
+    {
+        ret = SCOSSL_SUCCESS;
+        goto cleanup; // Not error - this can be called with NULL parameter for siglen
```
samuel-lee-msft, line 350: >goto cleanup; // Not error - this can be called with NULL parameter for siglen [](http://example.com/codeflow?start=8&length=78)

Also I wonder if this should be an error when pcbSignature is NULL - i.e. should this move into the if( pcbSignature != NULL ) block above?

ScosslCommon/src/scossl_rsa.c
```diff
@@ -0,0 +1,959 @@
...
+    {
+    case RSA_PSS_SALTLEN_DIGEST:
+        cbSalt = cbDigest;
+        break;
+    case RSA_PSS_SALTLEN_MAX:
+        cbSalt = cbSaltMax;
+        break;
+    case RSA_PSS_SALTLEN_AUTO:
+#ifdef RSA_PSS_SALTLEN_AUTO_DIGEST_MAX
+    // Added in 3.1, should only be used in provider. Unsupported auto salt len for verify
```
samuel-lee-msft, line 427: >should only be used in provider [](http://example.com/codeflow?start=21&length=31)

We can build SCOSSL engine for 3.1, right?

ScosslCommon/src/scossl_rsa.c
```diff
@@ -0,0 +1,959 @@
...
+    {
+        if (scError != SYMCRYPT_SIGNATURE_VERIFICATION_FAILURE)
+        {
+            SCOSSL_LOG_SYMCRYPT_ERROR(SCOSSL_ERR_F_RSAPSS_VERIFY, SCOSSL_ERR_R_SYMCRYPT_FAILURE,
+                                      "SymCryptRsaPssVerify returned unexpected error", scError);
+        }
+        goto cleanup;
+    }
+
+    ret = SCOSSL_SUCCESS;
```
samuel-lee-msft, line 495: The phrasing in scossl_rsa_pkcs1_verify seems a little cleaner.

ScosslCommon/src/scossl_rsa.c
```diff
@@ -0,0 +1,959 @@
...
+                             "Unknown type: %d.", mdnid);
+            goto cleanup;
+        }
+
+        scError = SymCryptRsaOaepDecrypt(
+            keyCtx->key,
+            pbSrc,
+            cbSrc,
+            SYMCRYPT_NUMBER_FORMAT_MSB_FIRST,
+            SymCryptSha1Algorithm,
```
samuel-lee-msft, line 691: >SymCryptSha1Algorithm [](http://example.com/codeflow?start=12&length=21)

scossl_mac_algo

samuel-lee-msft: Actually, all variables in this file called `scossl_mac_algo` should probably be renamed to `scossl_hash_alg` - they are pointers to hash algorithms, not mac algorithms

ScosslCommon/src/scossl_rsa.c
```diff
@@ -0,0 +1,959 @@
...
+                BN_clear_free(rsaParams->privateParams->d);
+            }
+            OPENSSL_free(rsaParams->privateParams);
+        }
+        OPENSSL_free(rsaParams);
+    }
+}
+
+_Use_decl_annotations_
+SCOSSL_STATUS scossl_rsa_export_key(PSYMCRYPT_RSAKEY key, SCOSSL_RSA_EXPORT_PARAMS *rsaParams)
```
samuel-lee-msft, line 802: >PSYMCRYPT_RSAKEY [](http://example.com/codeflow?start=36&length=16)

PCSYMCRYPT_RSAKEY

SymCryptEngine/src/e_scossl_rsa.c
```diff
@@ -833,25 +275,16 @@ SCOSSL_STATUS e_scossl_rsa_keygen(_Out_ RSA* rsa, int bits, _In_ BIGNUM* e,
...
-        BN_clear_free(rsa_d);
     }
 
-    if( pbData )
-    {
-        OPENSSL_clear_free( pbData, cbData );
-    }
+    // Only free the set params in failure case,
+    // since OpenSSL directly copies the pointers
+    scossl_rsa_free_export_params(rsaParams, !ret);
```
samuel-lee-msft, line 282: >!ret [](http://example.com/codeflow?start=45&length=4)

Not a huge fan of this as I need to look up what the values of ret can be - maybe better to use (ret != SCOSSL_SUCCESS)?

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -125,8 +124,7 @@ extern const OSSL_DISPATCH p_scossl_ecc_keymgmt_functions[];
 
 static const OSSL_ALGORITHM p_scossl_keymgmt[] = {
     // ALG("DH:dhKeyAgreement:1.2.840.113549.1.3.1", p_scossl_dh_keymgmt_functions),
-    // ALG("RSA:rsaEncryption:1.2.840.113549.1.1.1:", p_scossl_rsa_keymgmt_functions),
-    // ALG("RSA-PSS:RSASSA-PSS:1.2.840.113549.1.1.10", p_scossl_rsa_keymgmt_functions),
+    ALG("RSA:rsaEncryption:1.2.840.113549.1.1.1:", p_scossl_rsa_keymgmt_functions),
```
samuel-lee-msft, line 127: This looks like you aren't intercepting RSA-PSS?

SymCryptProvider/src/asymcipher/p_scossl_rsa_cipher.c
```diff
@@ -0,0 +1,373 @@
...
+
+SCOSSL_STATUS p_scossl_rsa_cipher_encrypt(_In_ SCOSSL_RSA_CIPHER_CTX *ctx,
+                                          _Out_writes_bytes_(*outlen) unsigned char *out, _Out_ size_t *outlen, size_t outsize,
+                                          _In_reads_bytes_(inlen) const unsigned char *in, size_t inlen)
+{
+    INT32 cbResult;
+    SCOSSL_STATUS ret;
+
+    // Default to SHA1 for OAEP. Update md in context so this is
+    // reflected in getparam
```
samuel-lee-msft, line 104: Should this also update the OSSL_ASYM_CIPHER_PARAM_MGF1_DIGEST to SHA1?

ScosslCommon/src/scossl_rsa.c
```diff
@@ -0,0 +1,959 @@
...
+        SCOSSL_LOG_SYMCRYPT_ERROR(SCOSSL_ERR_F_RSA_VERIFY, SCOSSL_ERR_R_SYMCRYPT_FAILURE,
+                                  "SymCryptRsaPkcs1verify returned unexpected error", scError);
+    }
+
+cleanup:
+    return ret;
+}
+
+_Use_decl_annotations_
+SCOSSL_STATUS scossl_rsapss_sign(SCOSSL_RSA_KEY_CTX *keyCtx, const EVP_MD *md, int cbSalt,
```
samuel-lee-msft, line 305: >const EVP_MD *md, [](http://example.com/codeflow?start=61&length=17)

Why do we pass EVP_MD* for rsa_pss but use mdnid for PKCS1 / OAEP?

When looking at the provider code it looks like we might use the EVP_MD for something (calling back into libcrypto to perform a hash), but we are essentially just using it for the NID.

SymCryptProvider/src/p_scossl_base.c
mamckee: We still are. The default provider needs to register PSS as a separate algorithm for encoding/decoding. Their implementation returns "RSA" for rsa_query_operation_name though, so those algorithms are selected when the key is used for signature/cipher, and the default provider will export to this interface instead of what was configured as "RSA-PSS".

ScosslCommon/src/scossl_rsa.c
mamckee: I think it would make more sense to put the check before pbSignature == NULL. The requirements aren't defined, but if both pbSignature and pcbSignature are NULL I'd consider that a failure case.

ScosslCommon/src/scossl_rsa.c
mamckee: Yes, only new code should be using this, in which case it shouldn't use the engine. I'll remove the note about the provider though to avoid future assumptions that the engine shouldn't handle this.

ScosslCommon/src/scossl_rsa.c
mamckee: Going to switch to camelCase to be in line with the other variables too.

ScosslCommon/src/scossl_rsa.c
```diff
@@ -0,0 +1,961 @@
...
+        SymcryptRsaParam.nPubExp = 1;
+        copy_ctx->key = SymCryptRsakeyAllocate(&SymcryptRsaParam, 0);
+        if (copy_ctx->key == NULL)
+        {
+            scossl_rsa_free_key_ctx(copy_ctx);
+            return NULL;
+        }
+
+        // TODO: Enable in SymCrypt
+        SymCryptRsakeyCopy((PCSYMCRYPT_RSAKEY)keyCtx->key, copy_ctx->key);
```
samuel-lee-msft, line 146: >SymCryptRsakeyCopy((PCSYMCRYPT_RSAKEY)keyCtx->key, copy_ctx->key); [](http://example.com/codeflow?start=8&length=66)

I agree this is the medium term solution, but I wonder whether we can get away with doing this manually in SCOSSL for this first iteration (can we do something ugly but simple like export then import?).

I am aware that we will want to have folks move to the SCOSSL provider, and if they need to move to a later SymCrypt version to use the provider they need to have a separate FIPS certification.

It would be nice to be able to separate the concerns;
i.e. enable upgrade to OpenSSL 3.x + SCOSSL provider with existing libsymcrypt.so, with some missing features (at least parity with the existing SCOSSL engine, but not all the features we want to support (i.e. AES-XTS)), and then sometime next year folks can update SymCrypt and SCOSSL to have full features.

SymCryptProvider/src/asymcipher/p_scossl_rsa_cipher.c
mamckee: The default provider doesn't but I'm not sure why. If no mgf1md is set they default to using whatever was configured for the md. I'm trying to stay consistent with their behavior as much as possible.

To be honest the only reason we need the mgf1MdInfo is to reject distinct Md/Mgf1Md

ScosslCommon/src/scossl_rsa.c
mamckee: It's also needed for calculating the saltlen but that can be pulled out into a separate function.

mamckee: Actually we get this value from scossl_get_expected_hash_length, so we should just use that.

ScosslCommon/src/scossl_rsa.c
mamckee: It won't be pretty but yes. I can just copy the code from scossl_rsa_export_key and pass that into SymCryptRsakeySetValue.

ScosslCommon/src/scossl_rsa.c
```diff
@@ -0,0 +1,946 @@
...
+
+_Use_decl_annotations_
+SCOSSL_STATUS scossl_rsapss_verify(SCOSSL_RSA_KEY_CTX *keyCtx, int mdnid, int cbSalt,
+                                   PCBYTE pbHashValue, SIZE_T cbHashValue,
+                                   PCBYTE pbSignature, SIZE_T pcbSignature)
+{
+    int ret = SCOSSL_FAILURE;
+    SYMCRYPT_ERROR scError = SYMCRYPT_NO_ERROR;
+    int cbSaltMax;
+    PCSYMCRYPT_HASH scosslHashAlgo = scossl_get_symcrypt_hash_algorithm(mdnid);;
```
samuel-lee-msft, line 411: >;; [](http://example.com/codeflow?start=78&length=2)

NP: ;;

ScosslCommon/src/scossl_rsa.c
```diff
@@ -0,0 +1,946 @@
...
+        }
+        scError = SymCryptRsaRawEncrypt(
+            keyCtx->key,
+            pbSrc,
+            cbSrc,
+            SYMCRYPT_NUMBER_FORMAT_MSB_FIRST,
+            0,
+            pbDst,
+            cbModulus);
+        cbResult = cbDst;
```
samuel-lee-msft, line 591: >cbDst [](http://example.com/codeflow?start=19&length=5)

Should this also be cbModulus?

ScosslCommon/src/scossl_rsa.c
```diff
@@ -0,0 +1,946 @@
...
+
+    cbModulus = SymCryptRsakeySizeofModulus(keyCtx->key);
+
+    if (pbDst == NULL)
+    {
+        *pcbDst = (INT32)cbModulus;
+        goto cleanup;
+    }
+
+    if (cbDst > INT_MAX)
```
samuel-lee-msft, line 632: >if (cbDst > INT_MAX) [](http://example.com/codeflow?start=4&length=20)

NP:
if (cbDst == (SIZE_T)-1)
is used in scossl_rsa_encrypt, should these be the same?

ScosslCommon/inc/scossl_rsa.h
```diff
@@ -0,0 +1,68 @@
...
+void scossl_rsa_free_key_ctx(_In_ SCOSSL_RSA_KEY_CTX *keyCtx);
+
+SCOSSL_STATUS scossl_rsa_pkcs1_sign(_In_ SCOSSL_RSA_KEY_CTX *keyCtx, int mdnid, 
+                                    _In_reads_bytes_(cbHashValue) PCBYTE pbHashValue, SIZE_T cbHashValue,
+                                    _Out_writes_bytes_(*pcbSignature) PBYTE pbSignature, _Out_ SIZE_T* pcbSignature);
+SCOSSL_STATUS scossl_rsa_pkcs1_verify(_In_ SCOSSL_RSA_KEY_CTX *keyCtx, int mdnid, 
+                                      _In_reads_bytes_(cbHashValue) PCBYTE pbHashValue, SIZE_T cbHashValue,
+                                      _In_reads_bytes_(pcbSignature) PCBYTE pbSignature, SIZE_T pcbSignature);
+
+SCOSSL_STATUS scossl_rsapss_sign(_In_ SCOSSL_RSA_KEY_CTX *keyCtx, _In_ int mdnid, int cbSalt,
```
samuel-lee-msft, line 45: >_In_ [](http://example.com/codeflow?start=66&length=4)

NP: _In_ is redundant for non-pointer type

SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
```diff
@@ -0,0 +1,983 @@
...
+
+static SCOSSL_STATUS p_scossl_rsa_keymgmt_import(_Inout_ SCOSSL_RSA_KEY_CTX *keyCtx, int selection, const _In_ OSSL_PARAM params[])
+{
+    BOOL include_private = (selection & OSSL_KEYMGMT_SELECT_PRIVATE_KEY) != 0;
+    const OSSL_PARAM *p;
+    SYMCRYPT_ERROR scError = SYMCRYPT_NO_ERROR;
+    SCOSSL_STATUS ret = SCOSSL_FAILURE;
+    UINT64 pubExp64;
+    PBYTE pbModulus = NULL;
+    SIZE_T cbModulus = 2048;
```
samuel-lee-msft, line 776: >SIZE_T cbModulus = 2048; [](http://example.com/codeflow?start=4&length=24)

Interesting - 2048-byte modulus is very big as a default size; operations on a 16384-bit RSA key are _slow_.

Where did this value come from, and are you sure you didn't mean 2048-bits?

samuel-lee-msft (dismissed):
Looks good to me, subject to the nitpicks :)

ScosslCommon/src/scossl_rsa.c
mamckee: Yes

SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
mamckee: Yes, correcting to use the same default bits macro as keygen for consistency too.

cmake-toolchain/LinuxUserMode-AMD64.cmake
```diff
@@ -7,7 +7,7 @@ set(CMAKE_SYSTEM_PROCESSOR AMD64)
 
 # Define _AMD64_ to set up the correct SymCrypt macros, e.g. SYMCRYPT_CPU_AMD64
 add_compile_options(-D_AMD64_)
-add_compile_options(-O3)
+# add_compile_options(-O3)
```
samuel-lee-msft, line 10: Why remove this?

mamckee: Thanks for the catch. Didn't mean to commit this

samuel-lee-msft (approved):

PR #64, opened 2023-07-19
Provider ECDH and ECDSA

mamckee:
- Added ECC key management interface to the provider
- Added ECDSA signature interface to the provider
- Added ECDH key exchange interface to the provider
- Refactored existing engine ECDH and ECDSA into common code

Some of the changes to the provider base in this PR, namely accepting libctx in provider creation, are also present in PR https://github.com/microsoft/SymCrypt-OpenSSL/pull/63.


SymCryptProvider/src/p_scossl_ecc.h
```diff
@@ -0,0 +1,20 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+typedef struct {
+    OSSL_LIB_CTX *libctx;
+    BOOL initialized;
+    int includePublic;
+    PSYMCRYPT_ECKEY key;
+    EC_GROUP* ecGroup;
```
samuel-lee-msft, line 15: >EC_GROUP* ecGroup; [](http://example.com/codeflow?start=4&length=18)

Why track OpenSSL's EC_GROUP rather than PCSYMCRYPT_ECURVE here?

SymCryptProvider/src/keyexch/p_scossl_ecdh.c
```diff
@@ -0,0 +1,174 @@
...
+    {
+        ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+        goto cleanup;
+    }
+
+    if (EC_GROUP_cmp(ctx->keyCtx->ecGroup, peerKeyCtx->ecGroup, bnCtx) != 0)
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_MISMATCHING_DOMAIN_PARAMETERS);
+        goto cleanup;
+    }
```
samuel-lee-msft, line 93: If tracking PCSYMCRYPT_ECURVE in SCOSSL_ECC_KEY_CTX, this would be replaced by SymCryptEcurveIsSame

SymCryptProvider/src/p_scossl_ecc.h
mamckee: I believe there was some issue I encountered when originally refactoring the engine code. Looking at where this is referenced, I don't think there's any issue with using PCSYMCRYPT_ECURVE.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.h
```diff
@@ -0,0 +1,164 @@
...
+    p_scossl_ecc_impexp_types_0x07,
+    p_scossl_ecc_impexp_types_0x80,
+    p_scossl_ecc_impexp_types_0x81,
+    p_scossl_ecc_impexp_types_0x82,
+    p_scossl_ecc_impexp_types_0x83,
+    p_scossl_ecc_impexp_types_0x84,
+    p_scossl_ecc_impexp_types_0x85,
+    p_scossl_ecc_impexp_types_0x86,
+    p_scossl_ecc_impexp_types_0x87};
+
```
samuel-lee-msft, line 161: Curious why these are 0x80-0x87, rather than 0x08-0x0f?

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -0,0 +1,712 @@
...
+            !OSSL_PARAM_BLD_push_octet_string(bld, OSSL_PKEY_PARAM_PUB_KEY, pbPublicKey, cbPublicKey))
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_SET_PARAMETER);
+            goto cleanup;
+        }
+    }
+
+    if ((selection & OSSL_KEYMGMT_SELECT_OTHER_PARAMETERS) != 0)
+    {
+        if (!OSSL_PARAM_BLD_push_int(bld, OSSL_PKEY_PARAM_EC_INCLUDE_PUBLIC, keyCtx->includePublic))
```
samuel-lee-msft, line 649: >OSSL_PKEY_PARAM_EC_INCLUDE_PUBLIC [](http://example.com/codeflow?start=42&length=33)

Curious about this OSSL_PKEY_PARAM_EC_INCLUDE_PUBLIC; it seems like on export we should be exporting the public key with the private key, even in the absense of OSSL_KEYMGMT_SELECT_PUBLIC_KEY?
It doesn't seem like we're doing anything like this right now, but maybe this is handled in libcrypto code?

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
```diff
@@ -0,0 +1,411 @@
...
+{
+    SIZE_T cbResult;
+
+    if (siglen == NULL)
+    {
+        ERR_raise(ERR_LIB_PROV, ERR_R_PASSED_NULL_PARAMETER);
+        return SCOSSL_FAILURE;
+    }
+
+    cbResult = 2*SymCryptEcurveSizeofFieldElement(ctx->keyCtx->key->pCurve);
```
samuel-lee-msft, line 140: >ctx->keyCtx->key->pCurve [](http://example.com/codeflow?start=50&length=24)

NP:
ctx->keyCtx->key->pCurve
I think could be just
ctx->keyCtx->curve

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
```diff
@@ -0,0 +1,411 @@
...
+    {
+        return ret;
+    }
+
+    // If sig is NULL, this is a size fetch, and the digest does not need to be computed
+    if (sig == NULL || EVP_DigestFinal(ctx->mdctx, digest, &cbDigest))
+    {
+        ctx->allowMdUpdates = sig != NULL;
+        ret = p_scossl_ecdsa_sign(ctx, sig, siglen, sigsize, digest, cbDigest);
+    }
```
samuel-lee-msft, line 243: This was a bit confusing to me, and compared to verify, allowMdUpdates is not reset in the same way if there is an error in the Digest computation.

Does the following modification make sense to you?

// If sig is NULL, this is a size fetch, and the digest does not need to be computed
if (sig != NULL)
{
    ctx->allowMdUpdates = TRUE;
    if (!EVP_DigestFinal(ctx->mdctx, digest, &cbDigest))
    {
        return ret;
    }    
}

ret = p_scossl_ecdsa_sign(ctx, sig, siglen, sigsize, digest, cbDigest);

mamckee: Yeah that makes sense. We should unconditionally allow digest updates after this function is called with actual data. Whether or not digest computation fails, the final call indicates the operation is complete.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -0,0 +1,712 @@
...
+        ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+        return NULL;
+    }
+
+    keyCtx->libctx = provctx->libctx;
+
+    return keyCtx;
+}
+
+static SCOSSL_ECC_KEY_CTX *p_scossl_ecc_keymgmt_dup_ctx(_In_ const SCOSSL_ECC_KEY_CTX *keyCtx)
```
samuel-lee-msft, line 60: Shouldn't `p_scossl_ecc_keymgmt_dup_ctx` take a second `int selection` parameter?

mamckee: Yes. I'm guessing this passed tests because the cast created undefined behavior that didn't lead to an error.

samuel-lee-msft: I was looking at this on the RSA front, when testing RSA-PSS improvements. I think it might make sense to just export/import in SCOSSL provider unconditionally for asymmetric keypairs (RSA and ECC) to support copying only the public portion.

mamckee: That's a fair point. I'll make the necessary changes.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -0,0 +1,782 @@
...
+        copyCtx->libctx = keyCtx->libctx;
+    }
+
+cleanup:
+    if (pbData != NULL)
+    {
+        OPENSSL_secure_clear_free(pbData, cbData);
+    }
+
+    return copyCtx;
```
samuel-lee-msft, line 163: >return copyCtx; [](http://example.com/codeflow?start=4&length=15)

In this function, on error, shouldn't we free copyCtx and return NULL?

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -0,0 +1,782 @@
...
+{
+    PBYTE pbData = NULL;
+    PBYTE pbPrivateKey = NULL;
+    PBYTE pbPublicKey = NULL;
+    SIZE_T cbData = 0;
+    SIZE_T cbPublicKey = 0;
+    SIZE_T cbPrivateKey = 0;
+    SYMCRYPT_ERROR scError = SYMCRYPT_NO_ERROR;
+
+    SCOSSL_ECC_KEY_CTX *copyCtx = OPENSSL_malloc(sizeof(SCOSSL_ECC_KEY_CTX));
```
samuel-lee-msft, line 71: >OPENSSL_malloc [](http://example.com/codeflow?start=34&length=14)

Safer to OPENSSL_zalloc here, as we may only set partial data?

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -0,0 +1,782 @@
...
+
+    keyCtx->key = SymCryptEckeyAllocate(keyCtx->curve);
+    if (keyCtx->key == NULL)
+    {
+        ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+        goto cleanup;
+    }
+
+    // We don't know whether this key will be used for ECDSA or ECDH
+    scError = SymCryptEckeySetRandom(SYMCRYPT_FLAG_ECKEY_ECDSA | SYMCRYPT_FLAG_ECKEY_ECDH, keyCtx->key);
```
samuel-lee-msft, line 254: >SYMCRYPT_FLAG_ECKEY_ECDSA [](http://example.com/codeflow?start=37&length=25)

This will unduly hamper performance; I think in the engine we assume ECDH, and then extend key usage to ECDSA iff the key is used in ECDSA.

This allows lower overhead in ECDHE key generation (we don't need to perform needless ECDSA FIPS key tests). ECDSA key generation is generally less performance critical.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -0,0 +1,782 @@
...
+    if (!keyCtx->initialized)
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_GET_PARAMETER);
+        return SCOSSL_FAILURE;
+    }
+
+    OSSL_PARAM *p;
+
+    if ((p = OSSL_PARAM_locate(params, OSSL_PKEY_PARAM_MAX_SIZE)) != NULL &&
+        !OSSL_PARAM_set_uint32(p, 2 * SymCryptEcurveSizeofScalarMultiplier(keyCtx->key->pCurve)))
```
samuel-lee-msft, line 284: >2 * SymCryptEcurveSizeofScalarMultiplier(keyCtx->key->pCurve) [](http://example.com/codeflow?start=34&length=61)

I think with DER encoding ECDSA signature can be a few (7) bytes larger than this?

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -0,0 +1,782 @@
...
+    // SCOSSL only allows named curves, so these is never true
+    if ((p = OSSL_PARAM_locate(params, OSSL_PKEY_PARAM_EC_DECODED_FROM_EXPLICIT_PARAMS)) != NULL &&
+        !OSSL_PARAM_set_int(p, 0))
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_GET_PARAMETER);
+        return SCOSSL_FAILURE;
+    }
+
+
+    if ((p = OSSL_PARAM_locate(params, OSSL_PKEY_PARAM_USE_COFACTOR_ECDH)) != NULL &&
```
samuel-lee-msft, line 330: >OSSL_PKEY_PARAM_USE_COFACTOR_ECDH [](http://example.com/codeflow?start=39&length=33)

Worth noting that if you add support for X25519, SymCrypt only supports multiplying by the cofactor. For NIST prime curves this value doesn't matter.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -0,0 +1,782 @@
...
+        hasSelection = hasSelection && keyCtx->initialized;
+    }
+    if ((selection & OSSL_KEYMGMT_SELECT_PRIVATE_KEY) != 0)
+    {
+        hasSelection = hasSelection && SymCryptEckeyHasPrivateKey(keyCtx->key);
+    }
+    if ((selection & OSSL_KEYMGMT_SELECT_DOMAIN_PARAMETERS) != 0)
+    {
+        hasSelection = hasSelection &&
+                       (keyCtx->key != NULL && keyCtx->key->pCurve != NULL);
```
samuel-lee-msft, line 365: >keyCtx->key != NULL && keyCtx->key->pCurve != NULL [](http://example.com/codeflow?start=24&length=50)

We shouldn't be reaching into the SymCrypt Eckey here. Just check that keyCtx->curve != NULL?

samuel-lee-msft: Actually, the same applies to anywhere you have `keyCtx->key->pCurve`

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -0,0 +1,782 @@
...
+                goto cleanup;
+            }
+        }
+
+        scError = SymCryptEckeySetValue(
+            pbPrivateKey, cbPrivateKey,
+            pbPublicKey, cbPublicKey,
+            SYMCRYPT_NUMBER_FORMAT_MSB_FIRST,
+            SYMCRYPT_ECPOINT_FORMAT_XY,
+            SYMCRYPT_FLAG_ECKEY_ECDSA | SYMCRYPT_FLAG_ECKEY_ECDH,
```
samuel-lee-msft, line 599: >SYMCRYPT_FLAG_ECKEY_ECDSA [](http://example.com/codeflow?start=12&length=25)

Same comment here - restrict to ECDH in import and extend to ECDSA on first use.

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
```diff
@@ -0,0 +1,414 @@
...
+        return SCOSSL_SUCCESS;
+    }
+
+    if (sigsize < cbResult)
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_OUTPUT_BUFFER_TOO_SMALL);
+        return SCOSSL_FAILURE;
+    }
+
+    if ((tbslen | ctx->mdSize) != tbslen)
```
samuel-lee-msft, line 153: >(tbslen | ctx->mdSize) [](http://example.com/codeflow?start=8&length=22)

Why are you bitwise or-ing here?
Shouldn't this just be `ctx->mdSize != tbslen`?

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
```diff
@@ -0,0 +1,414 @@
...
+{
+    SIZE_T cbResult;
+
+    if (siglen == NULL)
+    {
+        ERR_raise(ERR_LIB_PROV, ERR_R_PASSED_NULL_PARAMETER);
+        return SCOSSL_FAILURE;
+    }
+
+    cbResult = 2*SymCryptEcurveSizeofFieldElement(ctx->keyCtx->curve);
```
samuel-lee-msft, line 140: >cbResult = 2*SymCryptEcurveSizeofFieldElement(ctx->keyCtx->curve); [](http://example.com/codeflow?start=4&length=66)

This size is without the DER encoding. Probably you should factor this out into a little function which essentially returns this value + 7, and use it here and in the keymgmt OSSL_PKEY_PARAM_MAX_SIZE code.

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
```diff
@@ -0,0 +1,414 @@
...
+{
+    if (!p_scossl_ecdsa_signverify_init(ctx, keyCtx, params))
+    {
+        return SCOSSL_FAILURE;
+    }
+
+    if (mdname != NULL &&
+        (mdname[0] == '\0' || !EVP_MD_is_a(ctx->md, mdname)))
+    {
+        // Different digest specified than what was previously set by paramters.
```
samuel-lee-msft, line 180: >paramters [](http://example.com/codeflow?start=70&length=9)

NP: parameters

samuel-lee-msft (changes requested):
LGTM modulo the changes in comments

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
mamckee: We may only partially set data but the default behavior is all wrapped behind the selection check. The default values in case of partial setting are in the else block. Also libctx should always be set from keyCtx, and includePrivate should default to 1. I realize that I dropped setting libctx and initialized when I moved to copy by export, so I'll address that.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
mamckee: In the case of duplication, is our only option for getting this flag from the source key directly accessing the value (keyCtx->key->fAlgorithmInfo)? Just want to make sure I'm not missing a function I should use instead.

We should copy the flags in duplication.  It would be best to only call SymCryptEckeyExtendKeyUsage in sign/verify init  , but there's a case where an EVP_CTX (and by proxy the EVP_PKEY) are copied after sign/verify init.

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
mamckee: mdSize may not necessarily be set. I can't exactly remember why I did this instead of ctx->mdSize != 0 && tbslen != ctx->mdsize but the second is clearer.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -0,0 +1,793 @@
...
+                goto cleanup;
+            }
+
+            // Default ECDH only. If the key is used for ECDSA then we call SymCryptEckeyExtendKeyUsage
+            scError = SymCryptEckeySetValue(
+                pbPrivateKey, cbPrivateKey,
+                pbPublicKey, cbPublicKey,
+                SYMCRYPT_NUMBER_FORMAT_MSB_FIRST,
+                SYMCRYPT_ECPOINT_FORMAT_XY,
+                keyCtx->key->fAlgorithmInfo,
```
samuel-lee-msft, line 153: >keyCtx->key->fAlgorithmInfo [](http://example.com/codeflow?start=16&length=27)

This doesn't work generally - the internal algorithmInfo will track self-test run which are not valid flags to pass in here. We could expose a SymCrypt API to get the KeyUsage, but for now I would just reset to ECDH only here - SymCrypt needs to re-run FIPS tests on the new key before first use/after import anyway.

samuel-lee-msft: An alternative would be to track the key usage in the provider struct, similar to how we track the curve; but I don't think it's worth it.

samuel-lee-msft: I think we already do extend key usage on every sign/verify in the common code? It is not a costly operation.

samuel-lee-msft: Ahh - the solution in the old engine code was a bit more subtle.
On _generate_, set only ECDH by default, and ECDSA before first sign.
On _import_, set ECDH / ECDSA unconditionally (this covers ECDSA verify and importing a peer's ECDH public key).

I am a bit loathe to have SCOSSL dig into SymCrypt internals if we can avoid it.

mamckee: I agree with that statement, so if it's fine for performance to keep it in the common sign/verify operation I can just do that.

PR #65, opened 2023-07-25
Provider KDF

mamckee:
- Added provider support for the following KDF algorithms
  - HKDF
  - SSHKDF
  - TLS1-PRF
 - Refactored engine HKDF and TLS1-PRF to use common implementation
 - Minor cleanup in above algorithms in the engine to make style more consistent


ScosslCommon/inc/scossl_helpers.h
```diff
@@ -239,6 +239,15 @@ void _scossl_log_SYMCRYPT_ERROR(
 #define SCOSSL_LOG_SYMCRYPT_ERROR(func_code, reason_code, description, scError) \
     _scossl_log_SYMCRYPT_ERROR(SCOSSL_LOG_LEVEL_ERROR, func_code, reason_code, __FILE__, __LINE__, description, scError)
 
+//
+// Common helper functions
+//
+
+// Functions for converting OpenSSL types to their SymCrypt equivalent
+BOOL scossl_is_md_supported(int mdnid);
+PCSYMCRYPT_MAC scossl_get_symcrypt_mac_algorithm(int mdnid);
```
samuel-lee-msft, line 248: >scossl_get_symcrypt_mac_algorithm [](http://example.com/codeflow?start=15&length=33)

I think this should probably be `scossl_get_symcrypt_hmac_algorithm` - there are other constructions for forming a MAC from a hash (MD), this function translates mdnid -> SymCrypt HMAC

ScosslCommon/src/scossl_helpers.c
```diff
@@ -394,6 +394,74 @@ void _scossl_log_SYMCRYPT_ERROR(
...
+        return SymCryptHmacSha512Algorithm;
+    case NID_sha3_256:
+        return SymCryptHmacSha3_256Algorithm;
+    case NID_sha3_384:
+        return SymCryptHmacSha3_384Algorithm;
+    case NID_sha3_512:
+        return SymCryptHmacSha3_512Algorithm;
+    }
+    SCOSSL_LOG_ERROR(SCOSSL_ERR_F_GET_SYMCRYPT_MAC_ALGORITHM, SCOSSL_ERR_R_NOT_IMPLEMENTED,
+        "SymCrypt-OpenSSL does not support Mac algorithm %d", mdnid);
```
samuel-lee-msft, line 436: >Mac [](http://example.com/codeflow?start=43&length=3)

hmac

ScosslCommon/src/scossl_hkdf.c
```diff
@@ -0,0 +1,172 @@
...
+            return NULL;
+        }
+
+        if (ctx->pbKey == NULL)
+        {
+            copyCtx->pbKey = NULL;
+        }
+        else if ((copyCtx->pbKey = OPENSSL_memdup(ctx->pbKey, ctx->cbKey)) == NULL)
+        {
+            scossl_hkdf_freectx(copyCtx);
```
samuel-lee-msft, line 38: >scossl_hkdf_freectx(copyCtx); [](http://example.com/codeflow?start=12&length=29)

We may leak pbSalt here; refactor into cleanup section?

ScosslCommon/src/scossl_hkdf.c
```diff
@@ -0,0 +1,172 @@
...
+    OPENSSL_clear_free(ctx->pbKey, ctx->cbKey);
+    OPENSSL_cleanse(ctx, sizeof(*ctx));
+    return SCOSSL_SUCCESS;
+}
+
+_Use_decl_annotations_
+SCOSSL_STATUS scossl_hkdf_append_info(SCOSSL_HKDF_CTX *ctx,
+                                      PCBYTE pbInfo, SIZE_T cbInfo)
+{
+    if (cbInfo + ctx->cbInfo > HKDF_MAXBUF)
```
samuel-lee-msft, line 78: >cbInfo + ctx->cbInfo > HKDF_MAXBUF [](http://example.com/codeflow?start=8&length=34)

`cbInfo > HKDF_MAXBUF - ctx->cbInfo`
this way if cbInfo is (SIZE_T)-1, we don't overflow and memcpy a huge buffer

ScosslCommon/inc/scossl_helpers.h
mamckee: Is the error logging in the RSA code correct? e.g. "Using Mac algorithm MD5 which is not FIPS compliant"

samuel-lee-msft: Good catch - yeah, that is inaccurate and those logs would be better without the "Mac algorithm" part.

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -297,7 +297,21 @@ SCOSSL_STATUS OSSL_provider_init(_In_ const OSSL_CORE_HANDLE *handle,
         scossl_prov_initialized = 1;
     }
 
-    scossl_setup_logging();
+        if (core_get_libctx == NULL)
```
samuel-lee-msft, line 300: >     [](http://example.com/codeflow?start=4&length=4)

NP: indentation

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -297,7 +297,21 @@ SCOSSL_STATUS OSSL_provider_init(_In_ const OSSL_CORE_HANDLE *handle,
...
+    p_ctx = OPENSSL_malloc(sizeof(SCOSSL_PROVCTX));
+    if (p_ctx == NULL)
+    {
+        ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+        return SCOSSL_FAILURE;
+    }
+
+    p_ctx->handle = handle;
+    p_ctx->libctx = (OSSL_LIB_CTX *)core_get_libctx(handle);
+    *provctx = p_ctx;
```
samuel-lee-msft, line 315: Actually, this whole thing looks like a merge error - it's duplicated just below?

ScosslCommon/src/scossl_sshkdf.c
```diff
@@ -0,0 +1,119 @@
...
+        else if ((copyCtx->pbKey = OPENSSL_memdup(ctx->pbKey, ctx->cbKey)) == NULL)
+        {
+            scossl_sshkdf_freectx(copyCtx);
+            return NULL;
+        }
+
+        copyCtx->pHash = ctx->pHash;
+        copyCtx->cbKey = ctx->cbKey;
+        memcpy(copyCtx->hashValue, ctx->hashValue, ctx->cbHashValue);
+        memcpy(copyCtx->sessionId, ctx->sessionId, ctx->cbSessionId);
```
samuel-lee-msft, line 37: These are fine and strictly correct, but copying only the valid bytes from hashValue prevents the compiler from making this a simple fixed-size memcpy of the whole struct.

I wonder whether a better pattern for this and similar functions could be:
if( copyCtx != NULL )
{
    // Shallow copy the struct
    copyCtx = ctx;
    // Remove any copied pointers
    copyCtx->pbKey = NULL;
    // Copy data referenced by pointers
    if ( (ctx->pbKey != NULL) &&
        ((copyCtx->pbKey = OPENSSL_memdup(ctx->pbKey, ctx->cbKey)) == NULL) )
    {
        goto cleanup;
    }
}

(with cleanup freeing any non-NULL pointers in an error case)

SymCryptProvider/src/p_scossl_base.c
mamckee: Yes looks like it. And a duplicate allocation above too...

ScosslCommon/src/scossl_sshkdf.c
mamckee: Good point. I'll try this but the suggested change looks good to me.

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -315,6 +304,8 @@ SCOSSL_STATUS OSSL_provider_init(_In_ const OSSL_CORE_HANDLE *handle,
     p_ctx->libctx = (OSSL_LIB_CTX *)core_get_libctx(handle);
     *provctx = p_ctx;
 
+    *out = p_scossl_base_dispatch;
```
samuel-lee-msft, line 307: This is also set at the top of the function?

ScosslCommon/src/scossl_hkdf.c
samuel-lee-msft: This is still outstanding; refactor this and scossl_tls1prf_dupctx?

ScosslCommon/src/scossl_tls1prf.c
```diff
@@ -0,0 +1,130 @@
...
+    OPENSSL_clear_free(ctx->pbSecret, ctx->cbSecret);
+    OPENSSL_cleanse(ctx, sizeof(*ctx));
+    return SCOSSL_SUCCESS;
+}
+
+_Use_decl_annotations_
+SCOSSL_STATUS scossl_tls1prf_append_seed(SCOSSL_TLS1_PRF_CTX *ctx,
+                                         PCBYTE pbSeed, SIZE_T cbSeed)
+{
+    if (cbSeed + ctx->cbSeed > TLS1_PRF_MAXBUF)
```
samuel-lee-msft, line 65: >cbSeed + ctx->cbSeed [](http://example.com/codeflow?start=8&length=20)

Same comment on overflow as scossl_hkdf_append_info -

SymCryptEngine/src/e_scossl_tls1prf.h
```diff
@@ -28,8 +28,8 @@ SCOSSL_STATUS e_scossl_tls1prf_derive_init(_Inout_ EVP_PKEY_CTX *ctx);
 // is written to the keylen parameter. If key is not NULL, then keylen should contain the length of
 // the key buffer, the shared secret is written to key and the amount of data written to keylen.
```
samuel-lee-msft, line 29: Is this comment is wrong, given the change in SAL annotation?

I think the comment and SAL annotations are wrong for HKDF / SSHKDF too?

samuel-lee-msft: Ah, SSHKDF has a different engine function API (EVP_KDF). 

And HKDF does actually return a keylen for extract-only mode.
Perhaps this tls1prf derive just happens to always be called with a set keylen, hence the annotation change?
Weird :/

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -0,0 +1,353 @@
...
+
+static SCOSSL_PROV_HKDF_CTX *p_scossl_hkdf_dupctx(_In_ SCOSSL_PROV_HKDF_CTX *ctx)
+{
+    SCOSSL_PROV_HKDF_CTX *copyCtx = OPENSSL_malloc(sizeof(SCOSSL_PROV_HKDF_CTX));
+    if (copyCtx != NULL)
+    {
+        if ((copyCtx->hkdfCtx = scossl_hkdf_dupctx(ctx->hkdfCtx)) == NULL ||
+            (ctx->hkdfCtx->md != NULL && !EVP_MD_up_ref(ctx->hkdfCtx->md)))
+        {
+            OPENSSL_free(copyCtx);
```
samuel-lee-msft, line 85: > OPENSSL_free(copyCtx); [](http://example.com/codeflow?start=11&length=23)

I think we can leak `copyCtx->hkdfCtx` here?

SymCryptProvider/src/kdf/p_scossl_sshkdf.c
```diff
@@ -0,0 +1,316 @@
...
+            return SCOSSL_FAILURE;
+        }
+
+        if ((md = EVP_MD_fetch(ctx->libctx, mdName, mdProps)) == NULL)
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_DIGEST);
+            return SCOSSL_FAILURE;
+        }
+
+        mdName = EVP_MD_get0_name(md);
```
samuel-lee-msft, line 204: >mdName [](http://example.com/codeflow?start=8&length=6)

It's unclear to me how safe it is to use this mdName pointer after md is freed two lines below?

Why not keep an EVP_MD pointer like in HKDF?

SymCryptProvider/src/kdf/p_scossl_tls1prf.c
```diff
@@ -0,0 +1,255 @@
...
+                return SCOSSL_FAILURE;
+            }
+
+            if ((md = EVP_MD_fetch(ctx->libctx, mdName, mdProps)) == NULL)
+            {
+                ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_DIGEST);
+                return SCOSSL_FAILURE;
+            }
+
+            mdName = EVP_MD_get0_name(md);
```
samuel-lee-msft, line 183: >mdName [](http://example.com/codeflow?start=12&length=6)

Same comment here as for sshkdf - is it safe to persist this pointer to a string after md is freed?

SymCryptProvider/src/kdf/p_scossl_sshkdf.c
samuel-lee-msft: (or copy the string and own the memory for it)

SymCryptProvider/src/p_scossl_base.c
mamckee: Another merge issue. Looks like the last one in this file.

samuel-lee-msft:
LGTM modulo comments

SymCryptEngine/src/e_scossl_tls1prf.h
mamckee: The original annotation and comment were wrong, but I neglected to update the comment. If that's the expected behavior in the engine we've been expecting a fixed length, but I guess that behavior hasn't caused issues yet?

SymCryptProvider/src/kdf/p_scossl_hkdf.c
mamckee: Yes, in the case EVP_MD_up_ref fails.

SymCryptProvider/src/kdf/p_scossl_sshkdf.c
mamckee: Good catch. Looks like the behavior changed and it's no longer safe. 1.1.1's EVP_MD_name was a macro for OBJ_nid2sn, which would return the matching, global sn_ALGNAME string. In 3.0, this will normally be allocated during algorithm fetching, and will be freed by EVP_MD_free.

ScosslCommon/src/scossl_hkdf.c
mamckee: Sorry, my reply didn't post for some reason... scossl_hkdf_freectx frees copyCtx->pbSalt. Although cbSalt needs to be set before this section to be cleared properly.

samuel-lee-msft (dismissed):
Just need to fix up the conflicts and LGTM

samuel-lee-msft (approved):

PR #66, opened 2023-08-09
Provider MAC

mamckee:
- Added provider functionality for the following algorithms
  - HMAC
    - SHA1
    - SHA256
    - SHA384
    - SHA512
  - CMAC
    - AES-128
    - AES-192
    - AES-256
   - KMAC
     - 128
     - 256
- Refactored existing engine HMAC code into common implementation for HMAC and CMAC

### DIFF
--- a/ScosslCommon/inc/scossl_helpers.h
+++ b/ScosslCommon/inc/scossl_helpers.h
@@ -74,12 +74,12 @@ C_ASSERT( SYMCRYPT_ALIGN_VALUE < 256 );
     {                                                               \
         PBYTE scossl_aligned = (PBYTE) ptr;                         \
         PBYTE scossl_alloc = scossl_aligned - *(scossl_aligned-1);  \
-        deallocator(scossl_alloc, size);                            \
+        deallocator(scossl_alloc, size + SYMCRYPT_ALIGN_VALUE);     \
         ptr = NULL;                                                 \
     }
 
 #define SCOSSL_COMMON_ALIGNED_FREE(ptr, deallocator, typename)      \
-    SCOSSL_COMMON_ALIGNED_FREE_EX(ptr, deallocator, SCOSSL_ALIGNED_SIZEOF(typename))
+    SCOSSL_COMMON_ALIGNED_FREE_EX(ptr, deallocator, sizeof(typename))
 
 void SCOSSL_set_trace_level(int trace_level, int ossl_ERR_level);
 void SCOSSL_set_trace_log_filename(const char *filename);
--- a/ScosslCommon/inc/scossl_helpers.h
+++ b/ScosslCommon/inc/scossl_helpers.h
@@ -54,10 +54,10 @@ typedef _Return_type_success_(return >= 0) int SCOSSL_RETURNLENGTH; // For funct
 // We need to be able to represent the offset into our allocation in a single byte
 C_ASSERT( SYMCRYPT_ALIGN_VALUE < 256 );
 
-#define SCOSSL_COMMON_ALIGNED_ALLOC(ptr, allocator, typename)               \
+#define SCOSSL_COMMON_ALIGNED_ALLOC_EX(ptr, allocator, typename, size)      \
     typename *ptr;                                                          \
     {                                                                       \
-        PBYTE scossl_alloc = allocator(SCOSSL_ALIGNED_SIZEOF(typename));    \
+        PBYTE scossl_alloc = allocator(size + SYMCRYPT_ALIGN_VALUE);        \
         PBYTE scossl_aligned = NULL;                                        \
         if (scossl_alloc)                                                   \
         {                                                                   \
@@ -67,13 +67,19 @@ C_ASSERT( SYMCRYPT_ALIGN_VALUE < 256 );
         ptr = (typename *) scossl_aligned;                                  \
     }
 
-#define SCOSSL_COMMON_ALIGNED_FREE(ptr, deallocator, typename)      \
+#define SCOSSL_COMMON_ALIGNED_ALLOC(ptr, allocator, typename) \
+    SCOSSL_COMMON_ALIGNED_ALLOC_EX(ptr, allocator, typename, sizeof(typename))
+
+#define SCOSSL_COMMON_ALIGNED_FREE_EX(ptr, deallocator, size)       \
     {                                                               \
         PBYTE scossl_aligned = (PBYTE) ptr;                         \
         PBYTE scossl_alloc = scossl_aligned - *(scossl_aligned-1);  \
-        deallocator(scossl_alloc, SCOSSL_ALIGNED_SIZEOF(typename)); \
+        deallocator(scossl_alloc, size);                            \
         ptr = NULL;                                                 \
     }
+
+#define SCOSSL_COMMON_ALIGNED_FREE(ptr, deallocator, typename)      \
+    SCOSSL_COMMON_ALIGNED_FREE_EX(ptr, deallocator, SCOSSL_ALIGNED_SIZEOF(typename))
 
 void SCOSSL_set_trace_level(int trace_level, int ossl_ERR_level);
 void SCOSSL_set_trace_log_filename(const char *filename);
--- a/ScosslCommon/inc/scossl_mac.h
+++ b/ScosslCommon/inc/scossl_mac.h
@@ -8,8 +8,8 @@
 extern "C" {
 #endif
 
-typedef PVOID PSCOSSL_MAC_EXPANDED_KEY;
-typedef PVOID PSCOSSL_MAC_STATE;
+typedef VOID SCOSSL_MAC_EXPANDED_KEY, *PSCOSSL_MAC_EXPANDED_KEY;
+typedef VOID SCOSSL_MAC_STATE, *PSCOSSL_MAC_STATE;
 
 typedef VOID (SYMCRYPT_CALL * PSYMCRYPT_MAC_KEY_COPY) (PSCOSSL_MAC_EXPANDED_KEY pSrc, PSCOSSL_MAC_EXPANDED_KEY pDst);
 typedef VOID (SYMCRYPT_CALL * PSYMCRYPT_MAC_STATE_COPY) (PSCOSSL_MAC_STATE pSrc, PSCOSSL_MAC_EXPANDED_KEY pExpandedKey, PSCOSSL_MAC_STATE pDst);
@@ -31,24 +31,20 @@ typedef struct
     const SCOSSL_MAC_EX *pMacEx;
     PBYTE pbKey;
     SIZE_T cbKey;
+
+    // Provider specific fields
+    PVOID libctx;
+    // Purely informational
+    char* mdName;
 } SCOSSL_MAC_CTX;
 
-// The MAC context contains SymCrypt structures that must be properly
-// aligned. Because of this, these functions use the pointer to the aligned
-// memory rather than the SCOSSL_MAC_CTX directly. Callers can use
-// SCOSSL_ALIGN_UP if they need to access the SCOSSL_MAC_CTX directly.
-
-SCOSSL_MAC_CTX *scossl_mac_newctx();
 SCOSSL_MAC_CTX *scossl_mac_dupctx(_In_ SCOSSL_MAC_CTX *ctx);
 void scossl_mac_freectx(_Inout_ SCOSSL_MAC_CTX *ctx);
 
-SCOSSL_STATUS scossl_mac_set_md(_Inout_ SCOSSL_MAC_CTX *ctx, _In_ const EVP_MD *md);
-SCOSSL_STATUS scossl_mac_set_cipher(_Inout_ SCOSSL_MAC_CTX *ctx, _In_ const EVP_CIPHER *cipher);
+SCOSSL_STATUS scossl_mac_set_hmac_md(_Inout_ SCOSSL_MAC_CTX *ctx, _In_ const EVP_MD *md);
+SCOSSL_STATUS scossl_mac_set_cmac_cipher(_Inout_ SCOSSL_MAC_CTX *ctx, _In_ const EVP_CIPHER *cipher);
 SCOSSL_STATUS scossl_mac_set_mac_key(_Inout_ SCOSSL_MAC_CTX *ctx,
                                      _In_reads_bytes_(cbMacKey) PCBYTE pbMacKey, SIZE_T cbMacKey);
-
-SIZE_T scossl_mac_get_result_size(_In_ SCOSSL_MAC_CTX *ctx);
-SIZE_T scossl_mac_get_block_size(_In_ SCOSSL_MAC_CTX *ctx);
 
 SCOSSL_STATUS scossl_mac_init(_Inout_ SCOSSL_MAC_CTX *ctx,
                               _In_reads_bytes_(cbKey) PCBYTE pbKey, SIZE_T cbKey);
--- a/ScosslCommon/inc/scossl_mac.h
+++ b/ScosslCommon/inc/scossl_mac.h
@@ -8,16 +8,24 @@
 extern "C" {
 #endif
 
-typedef VOID (SYMCRYPT_CALL * PSYMCRYPT_MAC_KEY_COPY) (PVOID pSrc, PVOID pDst);
-typedef VOID (SYMCRYPT_CALL * PSYMCRYPT_MAC_STATE_COPY) (PVOID pSrc, PVOID pExpandedKey, PVOID pDst);
+typedef VOID (SYMCRYPT_CALL * PSYMCRYPT_MAC_KEY_COPY) (SYMCRYPT_MAC_EXPANDED_KEY *pSrc, SYMCRYPT_MAC_EXPANDED_KEY *pDst);
+typedef VOID (SYMCRYPT_CALL * PSYMCRYPT_MAC_STATE_COPY) (SYMCRYPT_MAC_STATE *pSrc, SYMCRYPT_MAC_EXPANDED_KEY *pExpandedKey, SYMCRYPT_MAC_STATE *pDst);
+
+// Structure for holding additional constant MAC functions and
+// properties needed by SCOSSL
+typedef struct
+{
+    PSYMCRYPT_MAC_KEY_COPY keyCopyFunc;
+    PSYMCRYPT_MAC_STATE_COPY stateCopyFunc;
+    SIZE_T blockSize;
+} SCOSSL_MAC_EX;
 
 typedef struct
 {
     SYMCRYPT_MAC_EXPANDED_KEY expandedKey;
     SYMCRYPT_MAC_STATE macState;
     PCSYMCRYPT_MAC pMac;
-    PSYMCRYPT_MAC_KEY_COPY keyCopyFunc;
-    PSYMCRYPT_MAC_STATE_COPY stateCopyFunc;
+    const SCOSSL_MAC_EX *pMacEx;
     PBYTE pbKey;
     SIZE_T cbKey;
 } SCOSSL_MAC_CTX;
@@ -25,7 +33,7 @@ typedef struct
 // The MAC context contains SymCrypt structures that must be properly
 // aligned. Because of this, these functions use the pointer to the aligned
 // memory rather than the SCOSSL_MAC_CTX directly. Callers can use
-// SCOSSL_ALIGN_UP if they need to access the SCOSSL_MAC_CTX.
+// SCOSSL_ALIGN_UP if they need to access the SCOSSL_MAC_CTX directly.
 typedef PVOID PSCOSSL_MAC_ALIGNED_CTX;
 
 PSCOSSL_MAC_ALIGNED_CTX scossl_mac_newctx();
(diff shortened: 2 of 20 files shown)


SymCryptProvider/src/mac/p_scossl_kmac.c
```diff
@@ -0,0 +1,355 @@
...
+    if (out != NULL)
+    {
+        if (outsize < ctx->cbOutput)
+        {
+            return SCOSSL_FAILURE;
+        }
+
+        if (ctx->xofMode)
+        {
+            ctx->pMacEx->extractFunc(&ctx->macState, out, ctx->cbOutput, TRUE);
```
ccalik, line 211: Are successive extractions from XOF-mode not allowed?

SymCryptProvider/src/mac/p_scossl_kmac.c
```diff
@@ -0,0 +1,355 @@
...
+
+    copyCtx->pMac = ctx->pMac;
+    copyCtx->pMacEx = ctx->pMacEx;
+
+    ctx->pMacEx->keyCopyFunc(&ctx->expandedKey, &copyCtx->expandedKey);
+    ctx->pMacEx->stateCopyFunc(&ctx->macState, &copyCtx->macState);
+
+    if (ctx->cbCustomizationString > 0)
+    {
+        memcpy(copyCtx->customizationString, ctx->customizationString, ctx->cbCustomizationString);
```
ccalik, line 150: It may be useful to ensure that the number of copied bytes does not exceed `KMAC_MAX_CUSTOM`.

SymCryptProvider/src/mac/p_scossl_kmac.c
```diff
@@ -0,0 +1,355 @@
...
+                                         ctx->customizationString, ctx->cbCustomizationString) != SYMCRYPT_NO_ERROR)
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_GET_PARAMETER);
+            return SCOSSL_FAILURE;
+        }
+
+        ctx->pMac->initFunc(&ctx->macState, &ctx->expandedKey);
+    }
+
+    if ((p = OSSL_PARAM_locate_const(params, OSSL_MAC_PARAM_CUSTOM)) != NULL)
```
ccalik, line 302: Is it clear to the callers that setting the customization string must be performed before setting the key?

SymCryptProvider/src/mac/p_scossl_kmac.c
mamckee: I'm not against adding the check but I'm unclear on the scenario we'd be handling. This value is controlled by our provider and the context is opaque to the caller so it should only be zero or a value checked in p_scossl_kmac_set_ctx_params. The only reason it would be higher than KMAC_MAX_CUSTOM is if the caller was trying to intentionally mess with the context or memory corruption right?

SymCryptProvider/src/mac/p_scossl_kmac.c
mamckee: No, they aren't tested by the openssl tests though so that's a gap we need to be aware of.

SymCryptProvider/src/mac/p_scossl_kmac.c
mamckee: Yes, documented here: https://www.openssl.org/docs/man3.0/man7/EVP_MAC-KMAC.html

We set the parameters before setting the key so if the customization string is passed to init we set this first.

SymCryptProvider/src/mac/p_scossl_kmac.c
ccalik: Right, it should be within bounds under normal circumstances.

SymCryptProvider/src/mac/p_scossl_kmac.c
```diff
@@ -208,7 +208,7 @@ static SCOSSL_STATUS p_scossl_kmac_final(_Inout_ PSCOSSL_KMAC_ALIGNED_CTX aligne
 
         if (ctx->xofMode)
         {
-            ctx->pMacEx->extractFunc(&ctx->macState, out, ctx->cbOutput, TRUE);
+            ctx->pMacEx->extractFunc(&ctx->macState, out, ctx->cbOutput, FALSE);
```
ccalik, line 211: On a second thought, the previous version seems to be better; it makes sense to get one output out of KMACXOF. We also leave the state unwiped here.

SymCryptProvider/src/mac/p_scossl_kmac.c
```diff
@@ -0,0 +1,357 @@
...
+
+static PSCOSSL_KMAC_ALIGNED_CTX *p_scossl_kmac128_newctx(ossl_unused void *provctx)
+{
+    PSCOSSL_KMAC_ALIGNED_CTX alignedCtx = OPENSSL_zalloc(SCOSSL_ALIGNED_SIZEOF(SCOSSL_KMAC_CTX));
+    if (alignedCtx != NULL)
+    {
+        SCOSSL_KMAC_CTX *ctx = (SCOSSL_KMAC_CTX *)SCOSSL_ALIGN_UP(alignedCtx);
+        ctx->pMac = SymCryptKmac128Algorithm;
+        ctx->pMacEx = &SymCryptKmac128AlgorithmEx;
+        ctx->cbOutput = ctx->pMac->resultSize;
```
ccalik, line 103: Note that SymCrypt's default KMAC output sizes are larger than OpenSSL defaults.

ScosslCommon/src/scossl_mac.c
```diff
@@ -0,0 +1,241 @@
...
+        ctx->pMac = SymCryptHmacSha384Algorithm;
+        ctx->pMacEx = &SymCryptHmacSha384Ex;
+        return SCOSSL_SUCCESS;
+    case NID_sha512:
+        ctx->pMac = SymCryptHmacSha512Algorithm;
+        ctx->pMacEx = &SymCryptHmacSha512Ex;
+        return SCOSSL_SUCCESS;
+    default:
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_GET_SYMCRYPT_HASH_ALGORITHM, SCOSSL_ERR_R_NOT_IMPLEMENTED,
+                         "SymCrypt engine does not support hash algorithm for MAC %d", type);
```
samuel-lee-msft, line 123: >SymCrypt engine [](http://example.com/codeflow?start=26&length=15)

This is common code - replace "SymCrypt engine" with "SCOSSL"?

ScosslCommon/src/scossl_mac.c
```diff
@@ -0,0 +1,241 @@
...
+    (PSYMCRYPT_MAC_KEY_COPY)SymCryptHmacSha512KeyCopy,
+    (PSYMCRYPT_MAC_STATE_COPY)SymCryptHmacSha512StateCopy,
+    SYMCRYPT_HMAC_SHA512_INPUT_BLOCK_SIZE
+};
+
+static const SCOSSL_MAC_EX SymCryptAesCmacEx = {
+    (PSYMCRYPT_MAC_KEY_COPY)SymCryptAesCmacKeyCopy,
+    (PSYMCRYPT_MAC_STATE_COPY)SymCryptAesCmacStateCopy,
+    SYMCRYPT_AES_CMAC_INPUT_BLOCK_SIZE
+};
```
samuel-lee-msft, line 39: We also expose Hmac SHA3 from latest SymCrypt - might be worth hooking this up as well unless @ccalik thinks otherwise?

SymCryptProvider/src/mac/p_scossl_hmac.c
```diff
@@ -0,0 +1,218 @@
...
+            return SCOSSL_FAILURE;
+        }
+
+        if ((md = EVP_MD_fetch(ctx->libctx, mdName, mdProps)) == NULL)
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_DIGEST);
+            return SCOSSL_FAILURE;
+        }
+
+        mdName = EVP_MD_get0_name(md);
```
samuel-lee-msft, line 175: >mdName = EVP_MD_get0_name(md); [](http://example.com/codeflow?start=8&length=30)

Same question as for the KDF PR - is this safe?

SymCryptProvider/src/mac/p_scossl_cmac.c
```diff
@@ -0,0 +1,198 @@
...
+    OSSL_PARAM_utf8_string(OSSL_MAC_PARAM_CIPHER, NULL, 0),
+    OSSL_PARAM_utf8_string(OSSL_MAC_PARAM_PROPERTIES, NULL, 0),
+    OSSL_PARAM_octet_string(OSSL_MAC_PARAM_KEY, NULL, 0),
+    OSSL_PARAM_END};
+
+static SCOSSL_STATUS p_scossl_cmac_set_ctx_params(_Inout_ SCOSSL_PROV_CMAC_CTX *ctx, _In_ const OSSL_PARAM params[]);
+
+static SCOSSL_PROV_CMAC_CTX *p_scossl_cmac_newctx(_In_ SCOSSL_PROVCTX *provctx)
+{
+    SCOSSL_PROV_CMAC_CTX *ctx = OPENSSL_zalloc(SCOSSL_ALIGNED_SIZEOF(SCOSSL_PROV_CMAC_CTX));
```
samuel-lee-msft, line 37: >SCOSSL_ALIGNED_SIZEOF [](http://example.com/codeflow?start=47&length=21)

not required here?

ScosslCommon/src/scossl_mac.c
mamckee: Pattern from existing common code is SymCrypt for OpenSSL, but SCOSSL is probably better here.

ScosslCommon/src/scossl_mac.c
mamckee: I'll go ahead and add SHA3 now that it's available. Worst case we can comment it out if issues arise.

SymCryptProvider/src/mac/p_scossl_cmac.c
mamckee: Nope, handled by scossl_mac_newctx. Neither is zalloc since all members are explicitly set.

SymCryptProvider/src/mac/p_scossl_kmac.c
mamckee: Output size is entirely provider controlled here. Applications should be getting the size from the provider instead of the OpenSSL defaults. If the provided buffer is too small we'll fail.

ScosslCommon/inc/scossl_helpers.h
```diff
@@ -67,14 +67,20 @@ C_ASSERT( SYMCRYPT_ALIGN_VALUE < 256 );
...
-#define SCOSSL_COMMON_ALIGNED_FREE(ptr, deallocator, typename)      \
+#define SCOSSL_COMMON_ALIGNED_ALLOC(ptr, allocator, typename) \
+    SCOSSL_COMMON_ALIGNED_ALLOC_EX(ptr, allocator, typename, sizeof(typename))
+
+#define SCOSSL_COMMON_ALIGNED_FREE_EX(ptr, deallocator, size)       \
     {                                                               \
         PBYTE scossl_aligned = (PBYTE) ptr;                         \
         PBYTE scossl_alloc = scossl_aligned - *(scossl_aligned-1);  \
-        deallocator(scossl_alloc, SCOSSL_ALIGNED_SIZEOF(typename)); \
+        deallocator(scossl_alloc, size);                            \
```
samuel-lee-msft, line 77: >size [](http://example.com/codeflow?start=34&length=4)

There's an asymmetry here with `SCOSSL_COMMON_ALIGNED_ALLOC_EX`

If you `SCOSSL_COMMON_ALIGNED_ALLOC_EX` with size 8, we will allocate 8+SYMCRYPT_ALIGN_VALUE
When we SCOSSL_COMMON_ALIGNED_FREE_EX with size 8 we only free with size 8 - that means if we're using clear-free as the deallocator, we'd only wipe the first 8 bytes.

I think this is fixed by specifying `size + SYMCRYPT_ALIGN_VALUE` here, and sizeof(typename) in `SCOSSL_COMMON_ALIGNED_FREE`?

ScosslCommon/inc/scossl_mac.h
```diff
@@ -0,0 +1,62 @@
...
+    const SCOSSL_MAC_EX *pMacEx;
+    PBYTE pbKey;
+    SIZE_T cbKey;
+} SCOSSL_MAC_CTX;
+
+// The MAC context contains SymCrypt structures that must be properly
+// aligned. Because of this, these functions use the pointer to the aligned
+// memory rather than the SCOSSL_MAC_CTX directly. Callers can use
+// SCOSSL_ALIGN_UP if they need to access the SCOSSL_MAC_CTX directly.
+
```
samuel-lee-msft, line 40: This comment seems to be incorrect; there are only pointers to other structs in `SCOSSL_MAC_CTX`?

samuel-lee-msft: I think it's just out of date looking at earlier revisions :)

ScosslCommon/src/scossl_mac.c
```diff
@@ -0,0 +1,328 @@
...
+        }
+
+        copyCtx->pMac = ctx->pMac;
+        copyCtx->pMacEx = ctx->pMacEx;
+
+        if (ctx->pMacEx != NULL)
+        {
+            if (ctx->expandedKey != NULL)
+            {
+                SCOSSL_COMMON_ALIGNED_ALLOC_EX(expandedKey, OPENSSL_malloc, PSCOSSL_MAC_EXPANDED_KEY, ctx->pMac->expandedKeySize);
```
samuel-lee-msft, line 90: >PSCOSSL_MAC_EXPANDED_KEY [](http://example.com/codeflow?start=76&length=24)

I think logically this should be `SCOSSL_MAC_EXPANDED_KEY`; you want to allocate an expanded key and you will get a pointer to it.

It works at the moment because `expandedKey` is a void* will be happy being set to a void**, but it's a bit dubious.

Perhaps define `SCOSSL_MAC_EXPANDED_KEY` and `SCOSSL_MAC_STATE` as `VOID`, and use them here?

ScosslCommon/src/scossl_mac.c
```diff
@@ -0,0 +1,328 @@
...
+                    goto cleanup;
+                }
+
+                copyCtx->expandedKey = expandedKey;
+                ctx->pMacEx->keyCopyFunc(ctx->expandedKey, copyCtx->expandedKey);
+            }
+
+            if (ctx->macState != NULL)
+            {
+                SCOSSL_COMMON_ALIGNED_ALLOC_EX(macState, OPENSSL_malloc, PSCOSSL_MAC_STATE, ctx->pMac->stateSize);
```
samuel-lee-msft, line 102: >PSCOSSL_MAC_STATE [](http://example.com/codeflow?start=73&length=17)

See previous comment

ScosslCommon/src/scossl_mac.c
```diff
@@ -0,0 +1,328 @@
...
+        ctx->pMacEx = &SymCryptHmacSha3_384Ex;
+        break;
+    case NID_sha3_512:
+        ctx->pMac = SymCryptHmacSha3_512Algorithm;
+        ctx->pMacEx = &SymCryptHmacSha3_512Ex;
+        break;
+    default:
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_GET_SYMCRYPT_HASH_ALGORITHM, SCOSSL_ERR_R_NOT_IMPLEMENTED,
+                         "SCOSSL does not support hash algorithm for MAC %d", type);
+        ret = SCOSSL_FAILURE;
```
samuel-lee-msft, line 187: >ret = SCOSSL_FAILURE; [](http://example.com/codeflow?start=8&length=21)

return here? Otherwise we fall into allocating something before we return.

ScosslCommon/src/scossl_mac.c
```diff
@@ -0,0 +1,328 @@
...
+    if (macState == NULL)
+    {
+        ret = SCOSSL_FAILURE;
+    }
+    else
+    {
+        ctx->macState = macState;
+    }
+
+    return ret;
```
samuel-lee-msft, line 205: Make consistent with `scossl_mac_set_cipher`?
    ctx->macState = macState;

    return  ctx->macState != NULL;

ScosslCommon/inc/scossl_mac.h
```diff
@@ -0,0 +1,62 @@
...
+SCOSSL_MAC_CTX *scossl_mac_newctx();
+SCOSSL_MAC_CTX *scossl_mac_dupctx(_In_ SCOSSL_MAC_CTX *ctx);
+void scossl_mac_freectx(_Inout_ SCOSSL_MAC_CTX *ctx);
+
+SCOSSL_STATUS scossl_mac_set_md(_Inout_ SCOSSL_MAC_CTX *ctx, _In_ const EVP_MD *md);
+SCOSSL_STATUS scossl_mac_set_cipher(_Inout_ SCOSSL_MAC_CTX *ctx, _In_ const EVP_CIPHER *cipher);
```
samuel-lee-msft, line 46: Maybe name these as `scossl_mac_set_hmac_md`, and `scossl_mac_set_cmac_cipher` respectively?

ScosslCommon/src/scossl_mac.c
```diff
@@ -0,0 +1,328 @@
...
+        break;
+    default:
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_GET_SYMCRYPT_HASH_ALGORITHM, SCOSSL_ERR_R_NOT_IMPLEMENTED,
+                         "SCOSSL does not support hash algorithm for MAC %d", type);
+        ret = SCOSSL_FAILURE;
+    }
+
+    if (ctx->macState != NULL)
+    {
+        SCOSSL_COMMON_ALIGNED_FREE_EX(ctx->macState, OPENSSL_clear_free, ctx->pMac->stateSize);
```
samuel-lee-msft, line 192: >ctx->pMac->stateSize [](http://example.com/codeflow?start=73&length=20)

This is the new state size, we should clear using the old state size I think?

Also, if ctx->expandedKey != NULL, we should also clear/free that based on the old expandedKeySize.

ScosslCommon/src/scossl_mac.c
```diff
@@ -0,0 +1,328 @@
...
+
+_Use_decl_annotations_
+SCOSSL_STATUS scossl_mac_init(SCOSSL_MAC_CTX *ctx,
+                              PCBYTE pbKey, SIZE_T cbKey)
+{
+    if (pbKey != NULL)
+    {
+        if (ctx->expandedKey != NULL)
+        {
+            SCOSSL_COMMON_ALIGNED_FREE_EX(ctx->expandedKey, OPENSSL_clear_free, ctx->pMac->expandedKeySize);
```
samuel-lee-msft, line 280: >SCOSSL_COMMON_ALIGNED_FREE_EX(ctx->expandedKey, OPENSSL_clear_free, ctx->pMac->expandedKeySize); [](http://example.com/codeflow?start=12&length=96)

If this is an expanded key for this key type then, why free and reallocate - just wipe and reuse?

If this is an expanded key for a different key type, then we're the wiping the wrong size; we need to free the expanded key when we change key type.

SymCryptProvider/src/mac/p_scossl_cmac.c
```diff
@@ -0,0 +1,197 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct
+{
+    // Needed for fetching cipher
+    OSSL_LIB_CTX *libctx;
+
+    SCOSSL_MAC_CTX *macCtx;
```
samuel-lee-msft, line 19: >SCOSSL_MAC_CTX *macCtx; [](http://example.com/codeflow?start=4&length=23)

It's probably not a huge deal if the performance is still OK, but the number of indirections is getting larger than necessary here:
The SCOSSL provider CTX keeps a pointer to the SCOSSL common MAC CTX which keeps pointers to the SymCrypt states.

It might be better to have a `SCOSSL_MAC_CTX` directly in the provider structs, especially as `scossl_mac_newctx` doesn't do anything other than allocate a zero buffer.

samuel-lee-msft: I guess this would also imply re-jigging the `scossl_mac_dupctx`

SymCryptProvider/src/mac/p_scossl_hmac.c
```diff
@@ -0,0 +1,221 @@
...
+    OSSL_PARAM_utf8_string(OSSL_MAC_PARAM_DIGEST, NULL, 0),
+    OSSL_PARAM_utf8_string(OSSL_MAC_PARAM_PROPERTIES, NULL, 0),
+    OSSL_PARAM_octet_string(OSSL_MAC_PARAM_KEY, NULL, 0),
+    OSSL_PARAM_END};
+
+static SCOSSL_STATUS p_scossl_hmac_set_ctx_params(_Inout_ SCOSSL_PROV_HMAC_CTX *ctx, _In_ const OSSL_PARAM params[]);
+
+static SCOSSL_PROV_HMAC_CTX *p_scossl_hmac_newctx(_In_ SCOSSL_PROVCTX *provctx)
+{
+    SCOSSL_PROV_HMAC_CTX *ctx = OPENSSL_zalloc(sizeof(SCOSSL_PROV_HMAC_CTX));
```
samuel-lee-msft, line 41: >OPENSSL_zalloc [](http://example.com/codeflow?start=32&length=14)

NP: `OPENSSL_malloc`?

SymCryptProvider/src/mac/p_scossl_kmac.c
```diff
@@ -0,0 +1,357 @@
...
+{
+    SCOSSL_KMAC_CTX *ctx, *copyCtx;
+    PSCOSSL_KMAC_ALIGNED_CTX alignedCopy;
+
+    if ((alignedCopy = OPENSSL_zalloc(SCOSSL_ALIGNED_SIZEOF(SCOSSL_KMAC_CTX))) == NULL)
+    {
+        return NULL;
+    }
+
+    ctx = (SCOSSL_KMAC_CTX *)SCOSSL_ALIGN_UP(alignedCtx);
```
samuel-lee-msft, line 139: >SCOSSL_ALIGN_UP [](http://example.com/codeflow?start=29&length=15)

This seems like it should be redundant if the in pointer is indeed an alignedCtx?

samuel-lee-msft: Hmm, actually I am confused about the use of alignment in this KMAC code.
Elsewhere in the provider we use: SCOSSL_COMMON_ALIGNED_ALLOC which generates an aligned allocation and SCOSSL_COMMON_ALIGNED_FREE which takes frees the aligned allocation by checking what the larger allocation address actually is.

Here we have a mix?

samuel-lee-msft (changes requested):
I think this mostly looks good; but there are a few issues to address before merging.

ScosslCommon/inc/scossl_mac.h
mamckee: Yep, hanging comment

ScosslCommon/src/scossl_mac.c
mamckee: Yep, otherwise we'll keep the old macState pointer if it existed...

ScosslCommon/src/scossl_mac.c
mamckee: We should probably just do the clear and re-alloc at the top of the function, so we don't unexpectedly persist the old state.

ScosslCommon/src/scossl_mac.c
mamckee: Do we even need to wipe? Shouldn't the expand key function overwrite the previous data?

SymCryptProvider/src/mac/p_scossl_cmac.c
mamckee: The problem is tracking the libctx. We can just store it as a void pointer in the common implementation though.

ScosslCommon/src/scossl_mac.c
samuel-lee-msft: You're right, the wipe is unnecessary.

SymCryptProvider/src/mac/p_scossl_hmac.c
mamckee: Converting this to directly use SCOSSL_MAC_CTX

SymCryptProvider/src/mac/p_scossl_kmac.c
mamckee: I think I got confused on the usage when I originally wrote this. Also we can use SCOSSL_COMMON_ALIGNED_ALLOC_EX, so I just need to go through and make this consistent.

samuel-lee-msft (dismissed):
LGTM

PR #70, opened 2023-11-22
Provider X25519

mamckee:
This PR adds X25519 support to the provider. Most of the changes are in the ECC key management interface to account for differences in key formats between X25519 and the existing supported curves. SymCrypt supports X25519 through its ECDH functions, so the ECDH key exchange interface is re-used and registered for X25519 with OpenSSL.


SymCryptProvider/src/p_scossl_ecc.h
```diff
@@ -12,6 +12,17 @@ typedef struct {
...
     PCSYMCRYPT_ECURVE curve;
+    BOOL isX25519;
+    // RFC7748 dictates that to decode the x25519 private key, we need to
+    // 1. Set the three least significant bits of the MSB to 0
+    // 2. Set the most significant bit of the LSB to 0
+    // 3. Set the second most significant bit of the LSB to 1
+    // SymCrypt expects the operation to be applied to the private
+    // key before import. In order to preserve the original key data
+    // for export, the transformed bits are stored here. The position
+    // of the midified bits in the MSB and LSB differ, so we can use
```
samuel-lee-msft, line 23: >midified [](http://example.com/codeflow?start=14&length=8)

NP: modified

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -12,16 +13,87 @@
...
+    OSSL_PARAM_END}
+
 static int scossl_prov_initialized = 0;
 
 static OSSL_FUNC_CRYPTO_malloc_fn *c_CRYPTO_malloc;
 static OSSL_FUNC_CRYPTO_zalloc_fn *c_CRYPTO_zalloc;
 static OSSL_FUNC_CRYPTO_free_fn *c_CRYPTO_free;
 static OSSL_FUNC_CRYPTO_clear_free_fn *c_CRYPTO_clear_free;
 
+static const OSSL_PARAM p_scossl_supported_group_list[][10] = {
```
samuel-lee-msft, line 84: >10 [](http://example.com/codeflow?start=56&length=2)

Do you need to have a value here?
It looks like the number of entries is 11

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -36,6 +37,14 @@ static const OSSL_PARAM p_scossl_ecc_keymgmt_gettable_param_types[] = {
     OSSL_PARAM_int(OSSL_PKEY_PARAM_USE_COFACTOR_ECDH, NULL),
     OSSL_PARAM_END};
 
+// We don't need to support setting the group for X25519 import/export
+// Note: OpenSSL passes the private key for X25519 as an octet string
+// instead of a BN in the general ECC case.
```
samuel-lee-msft, line 42: It looks like you've changed the general ECC case to octet string too?

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -312,21 +378,43 @@ static SCOSSL_STATUS p_scossl_ecc_keymgmt_get_params(_In_ SCOSSL_ECC_KEY_CTX *ke
         return SCOSSL_FAILURE;
     }
 
+    // b p_scossl_ecc_keymgmt.c:384
```
samuel-lee-msft, line 381: >// b p_scossl_ecc_keymgmt.c:384 [](http://example.com/codeflow?start=4&length=31)

remove this?

SymCryptProvider/src/p_scossl_base.c
mamckee: We do need the value, but it does need to be updated.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
mamckee: General case is still BN for the private key. The public key was also BN but should have been octet string. These are just informational for the caller.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -756,6 +850,185 @@ static SCOSSL_STATUS p_scossl_ecc_keymgmt_export(_In_ SCOSSL_ECC_KEY_CTX *keyCtx
...
+
+    return ret;
+}
+
+static SCOSSL_STATUS p_scossl_x25519_keymgmt_export(_In_ SCOSSL_ECC_KEY_CTX *keyCtx, int selection,
+                                                    _In_ OSSL_CALLBACK *param_cb, _In_ void *cbarg)
+{
+    SYMCRYPT_ERROR scError = SYMCRYPT_NO_ERROR;
+    SCOSSL_STATUS ret = SCOSSL_FAILURE;
+    OSSL_PARAM_BLD *bld = NULL;\
```
samuel-lee-msft, line 947: >\ [](http://example.com/codeflow?start=31&length=1)

typo

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -756,6 +850,185 @@ static SCOSSL_STATUS p_scossl_ecc_keymgmt_export(_In_ SCOSSL_ECC_KEY_CTX *keyCtx
     return ret;
 }
 
+static const OSSL_PARAM *p_scossl_x25519_keymgmt_impexp_types(int selection)
+{
+    return (selection & OSSL_KEYMGMT_SELECT_KEYPAIR) == 0 ? NULL : p_scossl_x25519_keymgmt_impexp_param_types;
```
samuel-lee-msft, line 855: >return (selection & OSSL_KEYMGMT_SELECT_KEYPAIR) == 0 ? NULL : p_scossl_x25519_keymgmt_impexp_param_types; [](http://example.com/codeflow?start=4&length=106)

This is subtly different to what the general ECC case returns.

This will return `p_scossl_x25519_keymgmt_impexp_param_types` if selection is any of:
`OSSL_KEYMGMT_SELECT_PRIVATE_KEY`,
`OSSL_KEYMGMT_SELECT_PUBLIC_KEY`,
`OSSL_KEYMGMT_SELECT_KEYPAIR` (== `OSSL_KEYMGMT_SELECT_PRIVATE_KEY|OSSL_KEYMGMT_SELECT_PUBLIC_KEY`)

For general ECC we would return 4 distinct parameter arrays for these cases, rather than 2. Not sure if it matters.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -756,6 +850,185 @@ static SCOSSL_STATUS p_scossl_ecc_keymgmt_export(_In_ SCOSSL_ECC_KEY_CTX *keyCtx
...
+            goto cleanup;
+        }
+
+        keyCtx->initialized = TRUE;
+    }
+
+    ret = SCOSSL_SUCCESS;
+cleanup:
+    if (pbPrivateKey != NULL)
+    {
```
samuel-lee-msft, line 933: Unnecessary check?

samuel-lee-msft (approved):
Looks good to me, modulo small fixes.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -756,6 +850,185 @@ static SCOSSL_STATUS p_scossl_ecc_keymgmt_export(_In_ SCOSSL_ECC_KEY_CTX *keyCtx
...
+
+    if ((params = OSSL_PARAM_BLD_to_param(bld)) == NULL)
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_SET_PARAMETER);
+        goto cleanup;
+    }
+
+    ret = param_cb(params, cbarg);
+
+cleanup:
```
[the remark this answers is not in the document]
mamckee: Yes. It looks like this was missed in the RSA implementation too so I'll go back and fix there as well.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -727,16 +958,193 @@ static SCOSSL_STATUS p_scossl_ecc_keymgmt_export(_In_ SCOSSL_ECC_KEY_CTX *keyCtx
 
     if ((selection & OSSL_KEYMGMT_SELECT_OTHER_PARAMETERS) != 0)
     {
-        if (!OSSL_PARAM_BLD_push_int(bld, OSSL_PKEY_PARAM_EC_INCLUDE_PUBLIC, keyCtx->includePublic))
+        if (!OSSL_PARAM_BLD_push_int(bld, OSSL_PKEY_PARAM_EC_INCLUDE_PUBLIC, keyCtx->includePublic) ||
+            !OSSL_PARAM_BLD_push_int(bld, OSSL_PKEY_PARAM_USE_COFACTOR_ECDH, 0))
```
samuel-lee-msft, line 962: >0 [](http://example.com/codeflow?start=77&length=1)

This should probably also be 1 for X25519.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -283,101 +364,240 @@ static SCOSSL_ECC_KEY_CTX *p_scossl_ecc_keygen(_In_ SCOSSL_ECC_KEYGEN_CTX *genCt
...
+                NULL, 0,
+                pbPublicKey, cbPublicKey,
+                numFormat,
+                pointFormat,
+                0);
+
+        if (scError != SYMCRYPT_NO_ERROR)
+        {
+            ERR_raise(ERR_LIB_PROV, ERR_R_INTERNAL_ERROR);
+            goto cleanup;
```
samuel-lee-msft, line 454: >goto cleanup; [](http://example.com/codeflow?start=12&length=13)

If we hit this, we try to free the wrong value of `pbPublicKey` - maybe better to have a separate variables tracking a pointer to the allocation and the start of the buffer set by SymCryptEckeyGetValue?

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -283,101 +364,240 @@ static SCOSSL_ECC_KEY_CTX *p_scossl_ecc_keygen(_In_ SCOSSL_ECC_KEYGEN_CTX *genCt
...
+    SIZE_T cbPublicKey = 0;
+    BN_CTX *bnCtx = NULL;
+    EC_POINT *ecPoint = NULL;
+    SCOSSL_STATUS ret = SCOSSL_FAILURE;
+    SYMCRYPT_NUMBER_FORMAT numFormat = keyCtx->isX25519 ? SYMCRYPT_NUMBER_FORMAT_LSB_FIRST : SYMCRYPT_NUMBER_FORMAT_MSB_FIRST;
+    SYMCRYPT_ECPOINT_FORMAT pointFormat = keyCtx->isX25519 ? SYMCRYPT_ECPOINT_FORMAT_X : SYMCRYPT_ECPOINT_FORMAT_XY;
+    const OSSL_PARAM *p;
 
-    if ((selection & SCOSSL_ECC_POSSIBLE_SELECTIONS) == 0)
+    if ((p = OSSL_PARAM_locate_const(params, OSSL_PKEY_PARAM_ENCODED_PUBLIC_KEY)) != NULL)
```
samuel-lee-msft, line 495: >OSSL_PKEY_PARAM_ENCODED_PUBLIC_KEY [](http://example.com/codeflow?start=45&length=34)

Just worth double checking what the expected behavior is for setting this on a key which already has a private key-pair.

Currently we wipe the existing keypair and keep only the decoded public key, but

samuel-lee-msft (dismissed):

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -283,101 +364,240 @@ static SCOSSL_ECC_KEY_CTX *p_scossl_ecc_keygen(_In_ SCOSSL_ECC_KEYGEN_CTX *genCt
...
         !OSSL_PARAM_set_int(p, 0))
     {
         ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_GET_PARAMETER);
-        return SCOSSL_FAILURE;
+        goto cleanup;
     }
 
 
     if ((p = OSSL_PARAM_locate(params, OSSL_PKEY_PARAM_USE_COFACTOR_ECDH)) != NULL &&
         !OSSL_PARAM_set_int(p, 0))
```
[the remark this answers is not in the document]
mamckee: OSSL_PKEY_PARAM_EC_DECODED_FROM_EXPLICIT_PARAMS and OSSL_PKEY_PARAM_USE_COFACTOR_ECDH actually aren't needed for x25519. I'll hide these behind a check for x25519.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
mamckee: Same as above. Unused for x25519

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
mamckee: It looks like in the general EC case, the private key is preserved, but in the x25519 case, the private key is wiped. I think we should wipe in either case. This gets used by EVP_PKEY_set1_encoded_public_key, which is typically used to create the peer key. The typical flow is to create an empty EVP_PKEY, set the parameters, then import the public key via this API rather than importing like we would with the public key.

The caller should either import the private and public key at the same time, or just the public key. Importing the private key, then setting the public key later through this parameter would be ill defined behavior.

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -283,101 +371,251 @@ static SCOSSL_ECC_KEY_CTX *p_scossl_ecc_keygen(_In_ SCOSSL_ECC_KEYGEN_CTX *genCt
...
+            goto cleanup;
+        }
+
+        pbPublicKeyTmp = pbPublicKey;
+
+        if (!keyCtx->isX25519)
+        {
+            pbPublicKeyTmp[0] = 0x04;
+            pbPublicKeyTmp++;
+            pbPublicKeyTmp--;
```
samuel-lee-msft, line 433: >pbPublicKeyTmp [](http://example.com/codeflow?start=12&length=14)

cbPublicKey

samuel-lee-msft (approved):

PR #71, opened 2023-12-05
Provider DH

mamckee:
- Add Diffie-Hellman key exchange to the provider
    - Added DH key management interface
        - Supports import by group parameters for app compat, but a warning is logged
    - Added DH key exchange interface
        - Supports X9.42 for app compat. This follows the default provider model of creating a new EVP_KDF and feeding it the result of the derivation. Import of X9.42 formatted keys is offloaded to the default provider
- Refactored common code from engine to share with provider
    - Derive function remains separate due to differences in the requirements between the provider and engine
    - Added support for added NID_modp_* in OpenSSL 3
    - Added support for import by group parameters due to requirement in the provider

### DIFF
--- a/ScosslCommon/src/scossl_dh.c
+++ b/ScosslCommon/src/scossl_dh.c
@@ -39,8 +39,8 @@ void scossl_dh_free_key_ctx(SCOSSL_DH_KEY_CTX *ctx)
 SCOSSL_DH_KEY_CTX *scossl_dh_dup_key_ctx(SCOSSL_DH_KEY_CTX *ctx, BOOL copyGroup)
 {
     SCOSSL_DH_KEY_CTX *copyCtx = OPENSSL_malloc(sizeof(SCOSSL_DH_KEY_CTX));
+    PSYMCRYPT_DLGROUP pDlgroupCopy = NULL;
     PCSYMCRYPT_DLGROUP pDlgroup;
-    PSYMCRYPT_DLGROUP pDlgroupCopy;
 
     if (copyCtx != NULL)
     {
--- a/ScosslCommon/src/scossl_dh.c
+++ b/ScosslCommon/src/scossl_dh.c
@@ -407,6 +407,7 @@ SCOSSL_STATUS scossl_dh_get_group_by_nid(int dlGroupNid, const BIGNUM* p,
     case NID_ffdhe4096:
         *ppDlGroup = _hidden_dlgroup_ffdhe4096;
         break;
+#if OPENSSL_VERSION_MAJOR >= 3
     case NID_modp_2048:
         *ppDlGroup = _hidden_dlgroup_modp2048;
         break;
@@ -416,6 +417,7 @@ SCOSSL_STATUS scossl_dh_get_group_by_nid(int dlGroupNid, const BIGNUM* p,
     case NID_modp_4096:
         *ppDlGroup = _hidden_dlgroup_modp4096;
         break;
+#endif // OPENSSL_VERSION_MAJOR >= 3
     default:
         // Not one of the supported ffdhe groups, but may still be a supported MODP group
         // Given we know the generator is 2, we can now check whether P corresponds to a MODP group
@@ -470,6 +472,7 @@ int scossl_dh_get_group_nid(PCSYMCRYPT_DLGROUP pDlGroup)
     {
         dlGroupNid = NID_ffdhe4096;
     }
+#if OPENSSL_VERSION_MAJOR >= 3
     else if (pDlGroup == _hidden_dlgroup_modp2048)
     {
         dlGroupNid = NID_modp_2048;
@@ -482,6 +485,7 @@ int scossl_dh_get_group_nid(PCSYMCRYPT_DLGROUP pDlGroup)
     {
         dlGroupNid = NID_modp_4096;
     }
+#endif // OPENSSL_VERSION_MAJOR >= 3
 
     return dlGroupNid;
 }
--- a/ScosslCommon/src/scossl_dh.c
+++ b/ScosslCommon/src/scossl_dh.c
@@ -11,9 +11,11 @@ extern "C" {
 static PSYMCRYPT_DLGROUP _hidden_dlgroup_ffdhe2048 = NULL;
 static PSYMCRYPT_DLGROUP _hidden_dlgroup_ffdhe3072 = NULL;
 static PSYMCRYPT_DLGROUP _hidden_dlgroup_ffdhe4096 = NULL;
+#if OPENSSL_VERSION_MAJOR >= 3
 static PSYMCRYPT_DLGROUP _hidden_dlgroup_modp2048 = NULL;
 static PSYMCRYPT_DLGROUP _hidden_dlgroup_modp3072 = NULL;
 static PSYMCRYPT_DLGROUP _hidden_dlgroup_modp4096 = NULL;
+#endif // OPENSSL_VERSION_MAJOR >= 3
 static BIGNUM* _hidden_bignum_modp2048 = NULL;
 static BIGNUM* _hidden_bignum_modp3072 = NULL;
 static BIGNUM* _hidden_bignum_modp4096 = NULL;
@@ -38,6 +40,7 @@ void scossl_dh_free_key_ctx(SCOSSL_DH_KEY_CTX *ctx)
 
 SCOSSL_DH_KEY_CTX *scossl_dh_dup_key_ctx(SCOSSL_DH_KEY_CTX *ctx, BOOL copyGroup)
 {
+    SCOSSL_STATUS success = SCOSSL_FAILURE;
     SCOSSL_DH_KEY_CTX *copyCtx = OPENSSL_malloc(sizeof(SCOSSL_DH_KEY_CTX));
     PSYMCRYPT_DLGROUP pDlgroupCopy = NULL;
     PCSYMCRYPT_DLGROUP pDlgroup;
@@ -64,20 +67,19 @@ SCOSSL_DH_KEY_CTX *scossl_dh_dup_key_ctx(SCOSSL_DH_KEY_CTX *ctx, BOOL copyGroup)
                     NULL,
                     NULL);
 
-                if ((pDlgroupCopy = SymCryptDlgroupAllocate(pcbPrimeP, pcbPrimeQ)) != NULL)
+                if ((pDlgroupCopy = SymCryptDlgroupAllocate(pcbPrimeP, pcbPrimeQ)) == NULL)
                 {
-                    SymCryptDlgroupCopy(pDlgroup, pDlgroupCopy);
+                    goto cleanup;
                 }
 
+                SymCryptDlgroupCopy(pDlgroup, pDlgroupCopy);
                 pDlgroup = pDlgroupCopy;
             }
 
             if (pDlgroup == NULL ||
                 (copyCtx->dlkey = SymCryptDlkeyAllocate(pDlgroup)) == NULL)
             {
-                SymCryptDlgroupFree(pDlgroupCopy);
-                OPENSSL_free(copyCtx);
-                copyCtx = NULL;
+                goto cleanup;
             }
             else
             {
@@ -88,6 +90,15 @@ SCOSSL_DH_KEY_CTX *scossl_dh_dup_key_ctx(SCOSSL_DH_KEY_CTX *ctx, BOOL copyGroup)
         {
             copyCtx->dlkey = NULL;
         }
+    }
+
+    success = SCOSSL_SUCCESS;
+cleanup:
+    if (!success)
+    {
+        OPENSSL_free(pDlgroupCopy);
+        scossl_dh_free_key_ctx(copyCtx);
+        copyCtx = NULL;
     }
 
     return copyCtx;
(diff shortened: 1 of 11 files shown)


ScosslCommon/src/scossl_dh.c
```diff
@@ -0,0 +1,495 @@
...
+        {
+            copyCtx->dlkey = NULL;
+        }
+    }
+
+    success = SCOSSL_SUCCESS;
+cleanup:
+    if (!success)
+    {
+        OPENSSL_free(pDlgroupCopy);
```
samuel-lee-msft, line 97: >OPENSSL_free [](http://example.com/codeflow?start=8&length=12)

SymCryptDlgroupFree

ScosslCommon/src/scossl_dh.c
```diff
@@ -0,0 +1,495 @@
...
+    {
+        cbPublicKey = 0;
+    }
+
+    // The SymCrypt provider must support non-FIPS groups since it cannot
+    // fallback to the default implementation like the engine.
+    if (skipGroupValidation)
+    {
+        SCOSSL_LOG_INFO(SCOSSL_ERR_F_DH_IMPORT_KEYPAIR, SCOSSL_ERR_R_NOT_FIPS_ALGORITHM,
+                        "Importing non-FIPS DH group.");
```
samuel-lee-msft, line 197: >Importing non-FIPS DH group [](http://example.com/codeflow?start=25&length=27)

NP, maybe: "Importing DH key in non-FIPS group."?

We already created/imported the group before entry to this function.

ScosslCommon/src/scossl_dh.c
```diff
@@ -0,0 +1,495 @@
...
+    }
+
+    ctx->initialized = TRUE;
+    ret = SCOSSL_SUCCESS;
+
+cleanup:
+    if (!ret)
+    {
+        ctx->initialized = FALSE;
+        SymCryptDlkeyFree(ctx->dlkey);
```
samuel-lee-msft, line 221: >SymCryptDlkeyFree [](http://example.com/codeflow?start=8&length=17)

I think this has to be guarded by a check for NULL, unlike the OpenSSL free functions; SymCrypt will unconditionally attempt to wipe the struct.

ScosslCommon/src/scossl_dh.c
```diff
@@ -0,0 +1,495 @@
...
```
samuel-lee-msft, line 270: >SymCryptDlkeyFree(ctx->dlkey); [](http://example.com/codeflow?start=8&length=30)

ditto

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
```diff
@@ -0,0 +1,1337 @@
...
+            }
+
+            if (!OSSL_PARAM_get_BN(p, &bnPublicKey))
+            {
+                ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_GET_PARAMETER);
+                goto cleanup;
+            }
+        }
+
+        if (!scossl_dh_import_keypair(ctx->keyCtx, nBitsPriv, pDlGroup, groupSetByParams, bnPrivateKey, bnPublicKey))
```
samuel-lee-msft, line 1036: >ctx->keyCtx [](http://example.com/codeflow?start=38&length=11)

Should we free any existing `ctx->keyCtx` before importing a new one here?

ScosslCommon/src/scossl_dh.c
```diff
@@ -0,0 +1,495 @@
...
+            else if (BN_cmp(p, _hidden_bignum_modp4096) == 0)
+            {
+                *ppDlGroup = _hidden_dlgroup_modp4096;
+            }
+        }
+
+        if (*ppDlGroup == NULL)
+        {
+            SCOSSL_LOG_INFO(SCOSSL_ERR_F_GET_DH_CONTEXT_EX, SCOSSL_ERR_R_OPENSSL_FALLBACK,
+                "SymCrypt engine does not support this DH dlgroup - falling back to OpenSSL.");
```
samuel-lee-msft, line 443: >SymCrypt engine does not support this DH dlgroup - falling back to OpenSSL. [](http://example.com/codeflow?start=17&length=75)

This message is incorrect for the provider, right?

SymCryptEngine/src/e_scossl_dh.c
```diff
@@ -148,22 +99,23 @@ SCOSSL_STATUS e_scossl_dh_generate_keypair(
...
     {
         OPENSSL_clear_free(pbData, cbData);
+
+        if( res != SCOSSL_SUCCESS )
+        {
+            // On error free the partially constructed key context
+            pKeyCtx->initialized = FALSE;
+            SymCryptDlkeyFree(pKeyCtx->dlkey);
+            pKeyCtx->dlkey = NULL;
+        }
```
samuel-lee-msft, line 118: Why is this only when `pbData != NULL`?

Also need to check `pKeyCtx->dlkey` is not NULL before `SymCryptDlkeyFree`.

SymCryptEngine/src/e_scossl_dh.c
```diff
@@ -398,56 +245,16 @@ SCOSSL_STATUS e_scossl_get_dh_context_ex(_Inout_ DH* dh, _Out_ PSCOSSL_DH_KEY_CO
 
     // OpenSSL is a bit inconsistent with how it handles different named safe-prime groups
     // We can get OpenSSL to return a nid for ffdhe groups we support
```
samuel-lee-msft, line 247: This comment is no longer relevant here; we've hidden these details in the common function.

SymCryptProvider/src/keyexch/p_scossl_dh.c
```diff
@@ -0,0 +1,497 @@
...
+    UINT pad;
+
+    // X9.42 parameters
+    enum scossl_kdf_type kdfType;
+    char *kdfMdName;
+    char *kdfMdProps;
+    char *kdfCekAlg;
+    unsigned char *kdfUkm;
+    SIZE_T kdfUkmlen;
+    SIZE_T kdfOutlen;
```
samuel-lee-msft, line 40: When `kdfType == SCOSSL_KDF_TYPE_X9_42`, should these fields be set to some reasonable defaults, or is it expected that the caller will set _all_ of these properties.

It's fine if we have the same behavior as the default provider :)

SymCryptProvider/src/keyexch/p_scossl_dh.c
```diff
@@ -0,0 +1,497 @@
...
+        params[3] = OSSL_PARAM_construct_octet_string(OSSL_KDF_PARAM_KEY, pbAgreedSecret, cbAgreedSecret);
+
+        if (ctx->kdfUkm != NULL)
+        {
+            params[4] = OSSL_PARAM_construct_octet_string(OSSL_KDF_PARAM_UKM, ctx->kdfUkm, ctx->kdfUkmlen);
+        }
+
+        params[5] = OSSL_PARAM_construct_end();
+
+        if (!EVP_KDF_derive(kdfCtx, secret, outlen, params))
```
samuel-lee-msft, line 217: >outlen [](http://example.com/codeflow?start=44&length=6)

Should this actually be `ctx->kdfOutlen`, and we should error if `outlen < ctx->kdfOutlen`?

SymCryptProvider/src/keyexch/p_scossl_dh.c
```diff
@@ -0,0 +1,497 @@
...
+
+        params[0] = OSSL_PARAM_construct_utf8_string(OSSL_KDF_PARAM_DIGEST, ctx->kdfMdName, 0);
+        params[1] = OSSL_PARAM_construct_utf8_string(OSSL_KDF_PARAM_PROPERTIES, ctx->kdfMdProps, 0);
+        params[2] = OSSL_PARAM_construct_utf8_string(OSSL_KDF_PARAM_CEK_ALG, ctx->kdfCekAlg, 0);
+        params[3] = OSSL_PARAM_construct_octet_string(OSSL_KDF_PARAM_KEY, pbAgreedSecret, cbAgreedSecret);
+
+        if (ctx->kdfUkm != NULL)
+        {
+            params[4] = OSSL_PARAM_construct_octet_string(OSSL_KDF_PARAM_UKM, ctx->kdfUkm, ctx->kdfUkmlen);
+        }
```
samuel-lee-msft, line 213: Is it well defined what happens here if params[4] is not set?

Should we set params[4] to `OSSL_PARAM_construct_end()` in this case?

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
```diff
@@ -0,0 +1,1337 @@
...
+            OPENSSL_free(copyCtx);
+            return NULL;
+        }
+
+        // If we also copied a custom group make sure it gets
+        // set in the caller so it can be properly freed.
+        if (copyCtx->keyCtx->initialized &&
+            ctx->pDlGroup != NULL)
+        {
+            copyCtx->pDlGroup = (PSYMCRYPT_DLGROUP) SymCryptDlkeyGetGroup(ctx->keyCtx->dlkey);
```
samuel-lee-msft, line 132: >copyCtx->pDlGroup = (PSYMCRYPT_DLGROUP) SymCryptDlkeyGetGroup(ctx->keyCtx->dlkey); [](http://example.com/codeflow?start=12&length=82)

This just copies the pointer to the Dlgroup.

There is only 1 Dlgroup object that multiple `SCOSSL_PROV_DH_KEY_CTX` have a pointer to, and may try to free multiple times?

samuel-lee-msft: Ohh, I see - `scossl_dh_dup_key_ctx` does the deep copy, but we need to track it at this level to know to free it. Looks OK

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
```diff
@@ -0,0 +1,1337 @@
...
+    scossl_dh_free_key_ctx(ctx->keyCtx);
+
+    if (ctx->pDlGroup != NULL)
+    {
+        SymCryptDlgroupFree(ctx->pDlGroup);
+    }
+}
+
+// Returns true if this is a known group. If this returns false, then the caller
+// is responsible for freeing the group. On error *ppDlGroup is set to NULL.
```
samuel-lee-msft, line 155: I think the code is the opposite of this comment?
Also might be better to talk about the out parameters in the comment rather than "returns".

samuel-lee-msft (dismissed):
LGTM modulo comments.

ScosslCommon/src/scossl_dh.c
mamckee: Yes, we're unconditionally freeing in the engine, so I'll clean that up too.

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
mamckee: scossl_dh_import_keypair frees the key. We can reuse the keyCtx.

ScosslCommon/src/scossl_dh.c
mamckee: Yes, I'll move this to the engine code and log an error in the provider instead.

SymCryptProvider/src/keyexch/p_scossl_dh.c
mamckee: We just need to pass 0/NULL/empty string for anything not set by the caller. This is what the default provider does.

SymCryptProvider/src/keyexch/p_scossl_dh.c
mamckee: We should to be safe. we can follow what the default provider does and just use an incrementing pointer, and set wherever we end to `OSSL_PARAM_construct_end()`

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
mamckee: I'll update the comment to make this clearer

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
mamckee: Forgot to update this after changing the function signature, so this comment is totally off.

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
samuel-lee-msft: Looking at this again, I think the intent is that you call:
`SymCryptDlkeyGetGroup(copyCtx->keyCtx->dlkey)` here?

We have done the deep copy, but need to extract the _copied_ dlgroup from the copied dlkey to track it for freeing?

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
```diff
@@ -0,0 +1,1346 @@
...
+    return ret;
+}
+
+static SCOSSL_STATUS p_scossl_dh_keymgmt_get_key_params(_In_ SCOSSL_DH_KEY_CTX *keyCtx, _Inout_ OSSL_PARAM params[])
+{
+    PBYTE pbPrivateKey;
+    PBYTE pbPublicKey;
+    PBYTE pbData = NULL;
+    SIZE_T cbPrivateKey;
+    SIZE_T cbPublicKey;
```
samuel-lee-msft, line 667: Should these be initialized to 0?

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
mamckee: Yep. It may be worth revisiting this to refcount the imported groups instead of copying every time but we can go back and do that if necessary.

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
mamckee: Yes

samuel-lee-msft (approved):

PR #74, opened 2024-02-24
Initial KeysInUse integration

mamckee:
This PR adds optional KeysInUse logging following the original formatting defined for https://github.com/microsoft/KeysInUse-OpenSSL

- The KeysInUse feature is hidden behind a compile time flag and must be enabled at runtime either by config or environment
- Logging is handled by a dedicated logging thread to minimize overhead on crypto
- All KeysInUse errors are logged to a separate file to avoid polluting the error stack
- Supported private key operations for logging:
    - RSA sign
    - RSA decrypt
    - ECDSA sign


SymCryptProvider/src/p_scossl_ecc.c
```diff
@@ -0,0 +1,90 @@
...
+        cbEncodedKey++;
+    }
+
+    if ((pbEncodedKey = OPENSSL_malloc(cbEncodedKey)) == NULL)
+    {
+        ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+        goto cleanup;
+    }
+
+    // pbPublicKeyTmp = *ppbEncodedPublicKey;
```
samuel-lee-msft, line 43: NP: this comment does not make much sense with latest code - remove?

SymCryptProvider/src/p_scossl_ecc.c
```diff
@@ -0,0 +1,90 @@
...
+        goto cleanup;
+    }
+
+    *ppbEncodedKey = pbEncodedKey;
+    *pcbEncodedKey = cbEncodedKey;
+
+    ret = SCOSSL_SUCCESS;
+
+cleanup:
+    if (!ret)
```
samuel-lee-msft, line 78: >!ret [](http://example.com/codeflow?start=8&length=4)

ret != SCOSSL_SUCCESS

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -428,6 +522,14 @@ SCOSSL_STATUS OSSL_provider_init(_In_ const OSSL_CORE_HANDLE *handle,
 
     *out = p_scossl_base_dispatch;
 
+#ifdef KEYSINUSE_ENABLED
+    // Start keysinuse if configured
+    if (core_get_params != NULL)
```
samuel-lee-msft, line 527: >core_get_params != NULL [](http://example.com/codeflow?start=8&length=23)

When does `core_get_params == NULL` here?

Should there be an error condition rather than just starting with KeysInUse disabled?

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -352,6 +363,86 @@ static SCOSSL_STATUS p_scossl_get_capabilities(ossl_unused void *provctx, _In_ c
...
+    if (keysinuseEnabled)
+    {
+        if (confMaxFileSize != NULL)
+        {
+            // Convert file size to off_t in bytes.
+            // This is the same behavior as atol but also handles MB, KB, and GB suffixes.
+            off_t maxFileSizeBytes = 0;
+            int i;
+            for (i = 0;
+                 confMaxFileSize[i] != '\0' && confMaxFileSize[i] < '0' && '9' < confMaxFileSize[i];
```
samuel-lee-msft, line 404: >confMaxFileSize[i] < '0' && '9' < confMaxFileSize[i] [](http://example.com/codeflow?start=47&length=52)

This always evaluates to false I think?

I think you may have meant:
`(confMaxFileSize[i] < '0' || '9' < confMaxFileSize[i])`
If we are trimming off the non-numeric prefix?

Feel like we could probably just only allow strings which are immediately numeric and do away with this for loop entirely?

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -352,6 +363,86 @@ static SCOSSL_STATUS p_scossl_get_capabilities(ossl_unused void *provctx, _In_ c
...
+                    maxFileSizeBytes <<= 20;
+                    break;
+                case 'G':
+                case 'g':
+                    maxFileSizeBytes <<= 30;
+                    break;
+                }
+            }
+
+            p_scossl_keysinuse_set_max_file_size(maxFileSizeBytes);
```
samuel-lee-msft, line 433: >maxFileSizeBytes [](http://example.com/codeflow?start=49&length=16)

Should there be a compile time limit on this below the maximum value of off_t?

Should there be an error on overflow (i.e. if `off_t` is 32-bits, and I specify I want max file size to be "4GB", I think this will be set to 0, and no logging will occur?)

SymCryptProvider/symcrypt_prov.cnf
```diff
@@ -6,4 +6,10 @@
...
 activate = 1
-identity = symcryptprovider
\ No newline at end of file
+identity = symcryptprovider
+keysinuse = keysinuse_sect
+
+[keysinuse_sect]
+enabled = 1
+max_file_size = 5KB
+logging_delay = 3600
```
samuel-lee-msft, line 15: >logging_delay [](http://example.com/codeflow?start=0&length=13)

Maybe this should explicitly have "seconds" in the name?

SymCryptProvider/src/p_scossl_base.c
samuel-lee-msft, line 433: >p_scossl_keysinuse_set_max_file_size [](http://example.com/codeflow?start=12&length=36)

We are ignoring return value here? Should we set keysinuseEnabled to false on error?

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -352,6 +363,86 @@ static SCOSSL_STATUS p_scossl_get_capabilities(ossl_unused void *provctx, _In_ c
...
+        }
+
+        if (confLoggingDelay != NULL)
+        {
+            p_scossl_keysinuse_set_logging_delay(atol(confLoggingDelay));
```
samuel-lee-msft, line 438: >p_scossl_keysinuse_set_logging_delay [](http://example.com/codeflow?start=12&length=36)

We are ignoring return value here? Should we set keysinuseEnabled to FALSE on error?

SymCryptProvider/src/p_scossl_rsa.c
```diff
@@ -169,6 +171,49 @@ SCOSSL_STATUS p_scossl_rsa_pss_restrictions_to_params(const SCOSSL_RSA_PSS_RESTR
...
+    SCOSSL_STATUS  ret = SCOSSL_FAILURE;
+
+    rsaParams = scossl_rsa_new_export_params(FALSE);
+    if (rsaParams == NULL ||
+        !scossl_rsa_export_key(key, rsaParams))
+    {
+        goto cleanup;
+    }
+
+    if ((cbEncodedKey = i2d_SymcryptRsaPublicKey(rsaParams, &pbEncodedKey)) < 0)
```
samuel-lee-msft, line 200: >i2d_SymcryptRsaPublicKey [](http://example.com/codeflow?start=24&length=24)

Where is this defined?

samuel-lee-msft: Oh, I see `IMPLEMENT_ASN1_FUNCTIONS_name` above..

I can believe this does the right thing but I don't know how to review this - `IMPLEMENT_ASN1_FUNCTIONS_name` is not documented! :p

Is it not feasible to to call more explicitly supported functions to encode n and e separately and concatenate them appropriately?

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,773 @@
...
+
+_Use_decl_annotations_
+SCOSSL_STATUS p_scossl_keysinuse_upref(SCOSSL_PROV_KEYSINUSE_INFO *keysinuseInfo, INT32 *refOut)
+{
+    if (keysinuseInfo == NULL)
+        return SCOSSL_FAILURE;
+
+    INT32 ref = 0;
+
+    if (CRYPTO_atomic_add(&keysinuseInfo->refCount, 1, &ref, keysinuseInfo->lock))
```
samuel-lee-msft, line 298: `!CRYPTO_atomic_add` here?

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,773 @@
...
+SCOSSL_STATUS p_scossl_keysinuse_downref(SCOSSL_PROV_KEYSINUSE_INFO *keysinuseInfo, INT32 *refOut)
+{
+    if (keysinuseInfo == NULL)
+        return SCOSSL_FAILURE;
+
+    INT32 ref = 0;
+
+    if (!CRYPTO_atomic_add(&keysinuseInfo->refCount, -1, &ref, keysinuseInfo->lock))
+    {
+        p_scossl_keysinuse_log_error("p_scossl_keysinuse_upref failed,OPENSSL_%d", ERR_get_error());
```
samuel-lee-msft, line 321: >up [](http://example.com/codeflow?start=57&length=2)

down

SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
```diff
@@ -247,6 +250,14 @@ static SCOSSL_PROV_RSA_KEY_CTX *p_scossl_rsa_keymgmt_dup_ctx(_In_ const SCOSSL_P
         copyCtx = NULL;
     }
 
+#ifdef KEYSINUSE_ENABLED
+    if (keyCtx->keysinuseInfo != NULL &&
```
samuel-lee-msft, line 254: check copyCtx is not NULL?

samuel-lee-msft:
>         (copyCtx->pssRestrictions = OPENSSL_memdup(keyCtx->pssRestrictions, sizeof(SCOSSL_RSA_PSS_RESTRICTIONS))) == NULL)

Probably follow pattern in ecc with jump to cleanup

---
In reply to: [2128187962](https://github.com/microsoft/SymCrypt-OpenSSL/pull/74#issuecomment-2128187962) [](http://example.com/codeflow?ancestors=2128187962)

---
Refers to: SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c:246 in ca235cd. [](commit_id = ca235cd95e89ee45466ac5e92c0372e06fc483e1, deletion_comment = False)

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,773 @@
...
+        goto cleanup;
+    }
+
+    if ((keysinuseInfo = OPENSSL_zalloc(sizeof(SCOSSL_PROV_KEYSINUSE_INFO))) != NULL)
+    {
+        if ((pbHash = OPENSSL_malloc(SYMCRYPT_SHA256_RESULT_SIZE)) == NULL ||
+            (keysinuseInfo->lock = CRYPTO_THREAD_lock_new()) == NULL)
+        {
+            p_scossl_keysinuse_log_error("malloc failure in p_scossl_keysinuse_info_new,OPENSSL_%d", ERR_R_MALLOC_FAILURE);
+            p_scossl_keysinuse_info_free(keysinuseInfo);
```
samuel-lee-msft, line 253: >p_scossl_keysinuse_info_free(keysinuseInfo); [](http://example.com/codeflow?start=12&length=44)

I think we leak `keysinuseInfo` here because `keysinuseInfo->refCount` is 0 (not 1)

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
```diff
@@ -105,9 +110,31 @@ static SCOSSL_STATUS p_scossl_ecdsa_signverify_init(_Inout_ SCOSSL_ECDSA_CTX *ct
...
+    ctx->operation = operation;
+
     if (keyCtx != NULL)
     {
         ctx->keyCtx = keyCtx;
+#ifdef KEYSINUSE_ENABLED
+        if (p_scossl_keysinuse_running() &&
+            operation == EVP_PKEY_OP_SIGN &&
+            keyCtx->isImported &&
+            keyCtx->keysinuseInfo == NULL)
```
samuel-lee-msft, line 122: >keyCtx [](http://example.com/codeflow?start=12&length=6)

Is it definitely true that keyCtxs only contain a single key for their whole lifetime?

It looks like we would have a bunch of memory leaks if a keyCtx is initialized multiple times (i.e. import/generate an already-initialized keyCtx), so I guess this is beyond the scope of KeysInUse.
If we do need to support multiple-initializations of a keyCtx, we need to downref and set keysinuseInfo to NULL at the point of changing the key material the keyCtx contains.

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
```diff
@@ -156,20 +189,36 @@ static SCOSSL_STATUS p_scossl_ecdsa_sign(_In_ SCOSSL_ECDSA_CTX *ctx,
...
     }
 
-    return scossl_ecdsa_sign(ctx->keyCtx->key, ctx->keyCtx->curve, tbs, tbslen, sig, (unsigned int *)siglen);
+    if (!scossl_ecdsa_sign(ctx->keyCtx->key, ctx->keyCtx->curve, tbs, tbslen, sig, (unsigned int *)siglen))
+    {
+        return SCOSSL_FAILURE;
+    }
+
+#ifdef KEYSINUSE_ENABLED
+    p_scossl_keysinuse_on_sign(ctx->keyCtx->keysinuseInfo);
```
samuel-lee-msft, line 198: >p_scossl_keysinuse_on_sign(ctx->keyCtx->keysinuseInfo); [](http://example.com/codeflow?start=4&length=55)

Probably out of scope for this PR, but it seems like it might be useful to track key usage, along with failure or success - maybe it could indicate some bad key which always fails but is used a lot.

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,773 @@
...
+            keysinuseInfo->decryptCounter++;
+        }
+
+        // Add to pending usage if not already in stack
+        if (!keysinuseInfo->logPending)
+        {
+            if (CRYPTO_THREAD_write_lock(sk_keysinuse_info_lock))
+            {
+                keysinuseInfo->logPending = TRUE;
+                keysinuseInfo->refCount++;
```
samuel-lee-msft, line 362: >keysinuseInfo->refCount++; [](http://example.com/codeflow?start=16&length=26)

I think this still needs to be an atomic add, right?

Other modifications of refcount are not guarded by the write lock.

SymCryptProvider/src/p_scossl_base.c
mamckee: This is a sanity check, so it should never happen, and it is an error case. We don't want any errors related to keysinuse to interfere with normal provider operation though, so the behavior is the same to the caller with the feature enabled or disabled. This is also why keysinuse errors are logged separately rather than calling ERR_raise.

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
```diff
@@ -105,9 +110,31 @@ static SCOSSL_STATUS p_scossl_ecdsa_signverify_init(_Inout_ SCOSSL_ECDSA_CTX *ct
...
+            keyCtx->keysinuseInfo == NULL)
+        {
+            // Initialize keysinuse for private keys. Generated keys are
+            // ignored to avoid noise from ephemeral keys.
+            PBYTE pbPublicKey;
+            SIZE_T cbPublicKey;
+
+            if (p_scossl_ecc_get_encoded_public_key(keyCtx, &pbPublicKey, &cbPublicKey))
+            {
+                keyCtx->keysinuseInfo = p_scossl_keysinuse_info_new(pbPublicKey, cbPublicKey);
```
samuel-lee-msft, line 131: >keyCtx->keysinuseInfo = p_scossl_keysinuse_info_new(pbPublicKey, cbPublicKey); [](http://example.com/codeflow?start=16&length=78)

This should probably be a compare-exchange to prevent two threads setting `keyCtx->keysinuseInfo` at the same time and leaking one of them. (Same for RSA)

SymCryptProvider/src/p_scossl_base.c
mamckee: This was mostly for stripping whitespace, but it looks like OpenSSL already does that for us so that sounds good. We can just check if maxFileSizeBytes <= 0 in p_scossl_keysinuse_set_max_file_size instead of < 0.

SymCryptProvider/src/p_scossl_base.c
mamckee: p_scossl_keysinuse_set_max_file_size checks for overflow but we don't raise an error for the reason listed above. If this fails we just fall back to the defaults.

mamckee: We're matching the behavior of the keysinuse engine here. There's two error cases here. Either the max size is too large and we overflow, or the number is negative. In case of overflow, we _could_ have a max size, but falling back to the default is safer. In case of negative it's a misconfiguration and we should ignore as if this was some nonsense value.

mamckee: I can make this a void return though since we aren't checking.

SymCryptProvider/src/p_scossl_rsa.c
mamckee: It's feasible, but would be alot more code and far less performant :/ The responsibility falls on the providers for this now (same reason why we had to add OSSL_SIGNATURE_PARAM_ALGORITHM_ID to the provider)

The recommended path is to use a provider with the encoder interface to replace i2d_RSAPublicKey, but that would involve exporting the key to the default provider, then using the default provider's encoder. https://www.openssl.org/docs/man3.3/man3/i2d_RSAPublicKey.html.

SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
mamckee: I'll just return NULL in the failure cases above since there's no cleanup.

SymCryptProvider/src/p_scossl_base.c
samuel-lee-msft: `p_scossl_keysinuse_set_max_file_size` will be passed 0 if I write 4GB in the config file and off_t is 32-bits, so there is no ability to check for overflow there, the overflow happens in the parsing in this function.

It would make more logical sense to me to clamp values to some compile-time maximum which fits in an off_t; but it probably doesn't matter a great deal given the config has to be trusted input (just would be nice to make configuring KeysInUse less error-prone)

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: Yeah, I'll set refcount at the begining of this block.

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
mamckee: It would be weird for the keyCtx to be used for multiple keys, but not unsupported.

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: We've acquired keysinuseInfo->lock at this point which is used for the refcount operations.

samuel-lee-msft: https://www.openssl.org/docs/manmaster/man3/CRYPTO_atomic_add.html#:~:text=CRYPTO_atomic_add()%20atomically%20adds,function%20will%20fail.

Atomic add will typically not use the lock

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,773 @@
...
+}
+
+//
+// KeysInUse info management
+//
+_Use_decl_annotations_
+SCOSSL_PROV_KEYSINUSE_INFO *p_scossl_keysinuse_info_new(PBYTE pbPublicKey, SIZE_T cbPublicKey)
+{
+    SCOSSL_PROV_KEYSINUSE_INFO *keysinuseInfo = NULL;
+    PBYTE pbHash = NULL;
```
samuel-lee-msft, line 240: >PBYTE pbHash = NULL; [](http://example.com/codeflow?start=4&length=20)

Seems like this may as well be on the stack as:
`BYTE aHash[32] = { 0 };`

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,773 @@
...
+                {
+                    first_use_counter++;
+                    wakeLoggingThread = TRUE;
+                }
+            }
+            else
+            {
+                p_scossl_keysinuse_log_error("Failed to lock keysinuse info stack,OPENSSL_%d", ERR_get_error());
+            }
+            CRYPTO_THREAD_unlock(sk_keysinuse_info_lock);
```
samuel-lee-msft, line 376: >CRYPTO_THREAD_unlock(sk_keysinuse_info_lock); [](http://example.com/codeflow?start=12&length=45)

Hrmm, we might get here having failed to lock `sk_keysinuse_info_lock`, and then incorrectly unlock the lock which another thread has locked.

i.e.
thread A fails to take lock
thread B takes lock
thread A unlocks lock
thread C takes lock
thread B and C corrupt sk_keysinuse_info


Seems like a bug in the OpenSSL documentation :/: [https://www.openssl.org/docs/manmaster/man3/CRYPTO_THREAD_write_lock.html](https://www.openssl.org/docs/manmaster/man3/CRYPTO_THREAD_write_lock.html)

SymCryptProvider/src/p_scossl_keysinuse.c
samuel-lee-msft: Interestingly, I think if CRYPTO_atomic_add did need to take the lock, then it would fail here (as the thread is already locked), so I think the code would need to look something like:

```
INT32 ref = 0;

...

if (!CRYPTO_atomic_add(&keysinuseInfo->refCount, 1, &ref, keysinuseInfo->lock))
{
    // we get here only when atomics not supported on this platform
    // and we failed to take lock because we already own it.
    // simply increment refCount
    keysinuseInfo->refCount++;
}
```

Alternatively, you could have a separate lock specially protecting `keysinuseInfo->refCount`, or not use CRYPTO_atomic_add in up/downref.

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,773 @@
...
+    {
+        p_scossl_keysinuse_log_error("Failed to lock keysinuse info,OPENSSL_%d", ERR_get_error());
+        return;
+    }
+    CRYPTO_THREAD_unlock(keysinuseInfo->lock);
+
+    if (wakeLoggingThread)
+    {
+        if ((pthreadErr = pthread_mutex_lock(&logging_thread_mutex)) == 0 ||
+            pthreadErr == EDEADLK) // Already have the lock
```
samuel-lee-msft, line 389: >pthreadErr == EDEADLK) // Already have the lock [](http://example.com/codeflow?start=12&length=47)

I don't see how we should ever get this return from `pthread_mutex_lock` here?

SymCryptProvider/src/signature/p_scossl_rsa_signature.c
```diff
@@ -211,35 +229,60 @@ static SCOSSL_STATUS p_scossl_rsa_sign(_In_ SCOSSL_RSA_SIGN_CTX *ctx,
...
     default:
         ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_PADDING_MODE);
+        goto err;
     }
 
-    return SCOSSL_FAILURE;
+#ifdef KEYSINUSE_ENABLED
+    if (ret && sig != NULL)
+    {
+        p_scossl_keysinuse_on_sign(ctx->keyCtx->keysinuseInfo);
```
samuel-lee-msft, line 268: >p_scossl_keysinuse_on_sign [](http://example.com/codeflow?start=8&length=26)

Another thing that would be nice is to track what the padding mode was for RSA - would be great to surface PKCS1 usage and try and configure folks to use PSS/OAEP instead.

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,773 @@
...
+        if (!keysinuseInfo->logPending)
+        {
+            if (CRYPTO_THREAD_write_lock(sk_keysinuse_info_lock))
+            {
+                keysinuseInfo->logPending = TRUE;
+                keysinuseInfo->refCount++;
+                sk_SCOSSL_PROV_KEYSINUSE_INFO_push(sk_keysinuse_info, keysinuseInfo);
+
+                // First use of this key, wake the logging thread
+                if (keysinuseInfo->firstUse == 0)
```
samuel-lee-msft, line 366: >if (keysinuseInfo->firstUse == 0) [](http://example.com/codeflow?start=16&length=33)

Kind of feels like this should be a check of:
```
if ((keysinuseInfo->signCounter + keysinuseInfo->decryptCounter) == 1)
```
and maybe keysinuseInfo->firstUse should be set to now here, rather than in the logging thread sometime later.

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,773 @@
...
+            if ((pthreadErr = pthread_cond_signal(&logging_thread_cond_wake_early)) != 0)
+            {
+                p_scossl_keysinuse_log_error("Failed to signal logging thread,SYS_%d", pthreadErr);
+            }
+        }
+        else
+        {
+            p_scossl_keysinuse_log_error("Add use failed to accquire mutex,SYS_%d", pthreadErr);
+        }
+        pthread_mutex_unlock(&logging_thread_mutex);
```
samuel-lee-msft, line 400: Actually, here I don't think the signaling thread needs to lock/unlock `logging_thread_mutex` at all - we should be able to call:
```
if ((pthreadErr = pthread_cond_signal(&logging_thread_cond_wake_early)) != 0)
{
    p_scossl_keysinuse_log_error("Failed to signal logging thread,SYS_%d", pthreadErr);
}
```

The mutex is not protecting writing any data here.

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,773 @@
...
+    SCOSSL_PROV_KEYSINUSE_INFO *pKeysinuseInfo;
+    SCOSSL_PROV_KEYSINUSE_INFO keysinuseInfoTmp;
+
+    do
+    {
+        // This is the exit point for the logging thread. In that case, is_logging should be FALSE,
+        // and logging_thread_cond_wake_early should be signalled. All pending events are logged
+        // before the thread exits.
+        if ((pthreadErr = pthread_mutex_lock(&logging_thread_mutex)) == 0 ||
+            pthreadErr == EDEADLK) // Already have the lock
```
samuel-lee-msft, line 645: >pthreadErr == EDEADLK) // Already have the lock [](http://example.com/codeflow?start=12&length=47)

Why this, we will error out below with pthreadErr != 0 -> Logging thread failed to acquire mutex anyway?

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,773 @@
...
+        {
+            p_scossl_keysinuse_log_error("Logging thread failed to accquire mutex,SYS_%d", pthreadErr);
+            goto cleanup;
+        }
+
+        if (waitStatus != 0 && waitStatus != ETIMEDOUT)
+        {
+            p_scossl_keysinuse_log_error("Logging thread woken up with unexpected status, SYS_%d", waitStatus);
+            goto cleanup;
+        }
```
samuel-lee-msft, line 697: Should we cleanup / explicitly turn off KeysInUse in these cases, rather than just stopping the logging thread?

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,773 @@
...
+            first_use_counter = 0;
+
+            for (int i = 0; i < numInfoPending; i++)
+            {
+                pKeysinuseInfo = sk_SCOSSL_PROV_KEYSINUSE_INFO_pop(sk_keysinuse_info);
+                if (pKeysinuseInfo != NULL)
+                {
+                    sk_SCOSSL_PROV_KEYSINUSE_INFO_push(sk_keysinuse_info_pending, pKeysinuseInfo);
+                }
+            }
```
samuel-lee-msft, line 723: I'm not quite sure I follow this bit.

It seems to be the assumption that the last `first_use_counter` infos are the ones which have just been added to the stack on their first use, but I don't see how this is enforced.
Probably most of the time you will get 1 new info which is first use, and this use will wake the logging thread and the last info on the stack will be the new info.

However, there's nothing stopping another thread with a key which has been used before, but not since the last logging timeout, being pushed on the stack after the new info.

It also looks like if we're on a machine using a new key every 15 minutes, and a logging timeout of 1 hour, that we'll only ever log new keys, and never get around to logging the rest of the stack (at least until we're unloaded)?

Is it prohibitively expensive to just do the dumb thing of logging everything in the stack on first use?

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,773 @@
...
+
+        if (waitStatus != 0 && waitStatus != ETIMEDOUT)
+        {
+            p_scossl_keysinuse_log_error("Logging thread woken up with unexpected status, SYS_%d", waitStatus);
+            goto cleanup;
+        }
+
+        if (CRYPTO_THREAD_write_lock(sk_keysinuse_info_lock))
+        {
+            now = time(NULL);
```
samuel-lee-msft, line 701: >now = time(NULL); [](http://example.com/codeflow?start=12&length=17)

Seems like this should not be in a critical section ideally

SymCryptProvider/src/p_scossl_keysinuse.h
```diff
@@ -0,0 +1,52 @@
...
+// This structure is refcounted. The key object this is associated with may
+// be freed, but logging events may be pending according to the logging backoff.
+// This is created on the first sign/decrypt init using the key, and destroyed
+// with the key context in the key management interface.
+typedef struct
+{
+    time_t firstUse;
+    time_t lastUse;
+    UINT signCounter;
+    UINT decryptCounter;
```
samuel-lee-msft, line 20: NP: Any reason these aren't UINT32 / UINT64?

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,773 @@
...
+        }
+        CRYPTO_THREAD_unlock(sk_keysinuse_info_lock);
+
+        while (sk_SCOSSL_PROV_KEYSINUSE_INFO_num(sk_keysinuse_info_pending) > 0)
+        {
+            pKeysinuseInfo = sk_SCOSSL_PROV_KEYSINUSE_INFO_pop(sk_keysinuse_info_pending);
+            if (CRYPTO_THREAD_write_lock(pKeysinuseInfo->lock))
+            {
+                pKeysinuseInfo->firstUse = pKeysinuseInfo->lastUse == 0 ? now : pKeysinuseInfo->firstUse;
+                pKeysinuseInfo->lastUse = now;
```
samuel-lee-msft, line 737: >pKeysinuseInfo->lastUse = now; [](http://example.com/codeflow?start=16&length=30)

It's a little odd to me to set lastUse here, which is more like, lastLogged. There could be a long time between the last call to `p_scossl_keysinuse_add_use` for this info and logging time.

If it's too expensive to get the time in `p_scossl_keysinuse_add_use` (I would expect it shouldn't be relevant compared to cost of asymmetric key operations), then at least should rename the field to lastLogged.

samuel-lee-msft (changes requested):
I think the overall structure looks good - just many nitpicks.

I have looked at all of the code now and don't have any more comments at the moment. I only skimmed through the interactions with the file system in `p_scossl_keysinuse_log_common`, but I guess that code is unchanged from the engine.

SymCryptProvider/src/p_scossl_base.c
mamckee: I see, you're right I misunderstood. I agree, it's unlikely anyone will set this file to something so large but clamping to a max value would be the behavior I'd expect as a consumer.

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: I like the first suggestion the best, since we don't have to manage an extra lock, and don't take extra locks on a platform with atomics.

I think the lock shouldn't be passed though. If atomics aren't supported and the lock is NULL, CRYPTO_atomic_add will fail. If the lock is passed, CRYPTO_atomic_add will also try to acquire the lock first. There's potential for deadlock if the implementation changes similar to https://github.com/openssl/openssl/issues/24517.

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
mamckee: OpenSSL doesn't provide a compatibility layer for compare-exchange atomic functions. I think the only way to do this in a safe, platform-agnostic way is to put this block behind a lock in the keyCtx.

If thread A and B enter this section at the same time, and thread A starts initializing, thread B has to wait for that operation before advancing anyways.

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: I think I was just trying to be safe when I wrote this. None of the functions that acquire this lock call into each other though so we don't need this check.

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: `signCounter` and `decryptCounter` should have been reset before when logged since this is a sliding window.

The names `firstUse` and `lastUse` were carried over from the engine, but they're a bit misleading. I think something like `lastLogTime` and `logTime` would be more accurate.

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: There's a race condition if we don't acquire this lock.

The logging thread acquires `logging_thread_mutex` and performs some initialization before `pthread_cond_timedwait`, which releases the lock. If we signal while the logging thread during this time, the signal is lost and the logging thread won't wake up early.

https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_cond_signal.html
> The _pthread_cond_signal()_ and _pthread_cond_broadcast()_ functions have no effect if there are no threads currently blocked on _cond_.

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: Probably, the cleanup functions will need to be split into something like `p_scossl_keysinuse_cleanup` that can be called from the logging thread and `p_scossl_keysinuse_teardown` that can be called from the main thread and perform the logging thread join too.

mamckee: Actually if we call `p_scossl_keysinuse_cleanup` from the end of the logging thread then `p_scossl_keysinuse_teardown` would just need to join and cleanup the thread

SymCryptProvider/src/p_scossl_keysinuse.c
samuel-lee-msft: I suppose that's true, but I don't see the race as being much different to when the signal is sent just before the logging thread took the mutex for the initialization (i.e. was just processing its last logging).
The signal doesn't do anything unless the logging thread is waiting for the signal.

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: It wouldn't be prohibitively expensive and that would simplify so we might as well do that. It doesn't match the behavior of the engine but the important thing is that the usage is logged. It might be a little noisier and fill the log file faster, but it would be less error prone.

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: Probably a bit overkill to have it in the loop as well. The time doens't need to be _that_ precise

SymCryptProvider/src/p_scossl_keysinuse.h
mamckee: Not really. Might as well make them UINT32.

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: I'll rename these based on the above comment

SymCryptProvider/src/p_scossl_keysinuse.c
samuel-lee-msft: Nice, yeah, I think I like passing NULL as the lock argument here - make sure reasoning is in a comment.

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: The logging thread checks `first_use_counter` before sleeping, so any first use events that occurred during the last loop are processed before trying to sleep again.

The race would be in these lines, after `first_use_counter` has been checked and the mutex released, but before the thread has gone to sleep:
https://github.com/microsoft/SymCrypt-OpenSSL/blob/ba1d198901745b2df8ec92364d43e16656006c83/SymCryptProvider/src/p_scossl_keysinuse.c#L658-L666

As discussed below, we shouldn't bother with first use count, but `firstUseCounter` currently acts as the predicate for the condition. It should be replaced with a boolean, and we should set it under logging_thread_mutex instead of sk_keysinuse_info_lock.

SymCryptProvider/src/p_scossl_keysinuse.c
samuel-lee-msft: Oh, I don't see that `signCounter` / `decryptCounter` are ever reset, my understanding was that they were cumulative across the lifetime of the info object (so would only sum to 0 when the object has just been created with no use yet).

SymCryptProvider/src/p_scossl_base.c
samuel-lee-msft: I see that now on overflow we just keep the default value. That's certainly better than setting the size to 0, but still a little bit unexpected.

I would still probably add a compile time constant in here `SCOSSL_MAX_CONFIGURABLE_FILE_SIZE` and clamp `maxFileSizeBytes` to that.

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
samuel-lee-msft: Generally prefer to go compare-exchange as we scale up number of threads - but lock should be fine here

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,783 @@
...
+    {
+        if ((pthreadErr = pthread_mutex_lock(&logging_thread_mutex)) == 0)
+        {
+            first_use_pending = TRUE;
+
+            if ((pthreadErr = pthread_cond_signal(&logging_thread_cond_wake_early)) != 0)
+            {
+                p_scossl_keysinuse_log_error("Failed to signal logging thread,SYS_%d", pthreadErr);
+            }
+           pthread_mutex_unlock(&logging_thread_mutex);
```
samuel-lee-msft, line 414: >pthread_mutex_unlock [](http://example.com/codeflow?start=11&length=20)

NP: indent

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,783 @@
...
+
+        while (sk_SCOSSL_PROV_KEYSINUSE_INFO_num(sk_keysinuse_info_pending) > 0)
+        {
+            pKeysinuseInfo = sk_SCOSSL_PROV_KEYSINUSE_INFO_pop(sk_keysinuse_info_pending);
+            if (CRYPTO_THREAD_write_lock(pKeysinuseInfo->lock))
+            {
+                now = time(NULL);
+
+                pKeysinuseInfo->lastLogTime = pKeysinuseInfo->logTime == 0 ? now : pKeysinuseInfo->lastLogTime;
+                pKeysinuseInfo->logTime = now;
```
samuel-lee-msft, line 741: Seems like:
`lastLogTime` should be named `firstLogTime`, and `logTime` should be named `lastLogTime`?

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -0,0 +1,783 @@
...
+
+        if (waitStatus != 0 && waitStatus != ETIMEDOUT)
+        {
+            p_scossl_keysinuse_log_error("Logging thread woken up with unexpected status, SYS_%d", waitStatus);
+            goto cleanup;
+        }
+
+        if (CRYPTO_THREAD_write_lock(sk_keysinuse_info_lock))
+        {
+            first_use_pending = FALSE;
```
samuel-lee-msft, line 716: >first_use_pending = FALSE; [](http://example.com/codeflow?start=12&length=26)

Move this up to be under the `logging_thread_mutex`?

i.e.
```
if (first_use_pending)
{
    first_use_pending = FALSE;
    waitStatus = 0;
}
```

samuel-lee-msft (commented):
Very close to ready - just a few hanging comments

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: Sure, that sounds a little clearer

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: There's a potential race where the crypto thread has pushed the event to the stack, but hasn't set first_use_pending. In that case the event will be consumed by the logging thread, and first_use_pending will be TRUE at the beginning of the next loop.

That should be fine though, the worst case is that the logging thread makes an extra loop as a no-op.

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -350,6 +366,92 @@ static SCOSSL_STATUS p_scossl_get_capabilities(ossl_unused void *provctx, _In_ c
...
+    const char *confMaxFileSize = NULL;
+    const char *confLoggingDelay = NULL;
+    const char *envEnabled = NULL;
+
+    OSSL_PARAM keysinuseParams[] = {
+        OSSL_PARAM_utf8_ptr(CONF_KEYSINUSE_ENABLED, &confEnabled, 0),
+        OSSL_PARAM_utf8_ptr(CONF_KEYSINUSE_MAX_FILE_SIZE, &confMaxFileSize, 0),
+        OSSL_PARAM_utf8_ptr(CONF_KEYSINUSE_LOGGING_DELAY, &confLoggingDelay, 0),
+        OSSL_PARAM_END};
+
```
samuel-lee-msft, line 383: >maxFileSizeBytes = maxFileSizeBytes * 10 + (confMaxFileSize[i++] - '0'); [](http://example.com/codeflow?start=16&length=72)

Can also overflow here technically - a bit weird to not clamp on this loop too.

samuel-lee-msft (dismissed):
Just one nitpick w.r.t. clamping

samuel-lee-msft (approved):

PR #75, opened 2024-03-01
Properly support restricted PSS keys

mamckee:
The SymCrypt provider only supported PSS padded RSA if the key was imported through the RSA key management interface. RSASSA-PSS keys were not routed to the SymCrypt provider, and the keys' restrictions would not be properly handled. 

This PR addresses the gap by adding the RSASSA-PSS key management interface and updating the RSA signature interface to enforce the restrictions associated with RSASSA-PSS keys.

### DIFF
--- a/SymCryptProvider/src/p_scossl_rsa.c
+++ b/SymCryptProvider/src/p_scossl_rsa.c
@@ -15,8 +15,8 @@
 extern "C" {
 #endif
 
-#define SCOSSL_PROV_RSA_PSS_DEFAULT_MD 0
-#define SCOSSL_PROV_RSA_PSS_DEFAULT_SATLLEN_MIN 20
+#define SCOSSL_PROV_RSA_PSS_DEFAULT_MD (0) // Index of the default MD in p_scossl_rsa_supported_mds (SHA1)
+#define SCOSSL_PROV_RSA_PSS_DEFAULT_SALTLEN_MIN (20)
 
 static const OSSL_ITEM p_scossl_rsa_supported_mds[] = {
     {NID_sha1,     OSSL_DIGEST_NAME_SHA1}, // Default
@@ -113,7 +113,7 @@ SCOSSL_STATUS p_scossl_rsa_pss_restrictions_from_params(OSSL_LIB_CTX *libctx, co
         // as the default provider.
         pssRestrictions->mdInfo = &p_scossl_rsa_supported_mds[SCOSSL_PROV_RSA_PSS_DEFAULT_MD];
         pssRestrictions->mgf1MdInfo = &p_scossl_rsa_supported_mds[SCOSSL_PROV_RSA_PSS_DEFAULT_MD];
-        pssRestrictions->cbSaltMin = SCOSSL_PROV_RSA_PSS_DEFAULT_SATLLEN_MIN;
+        pssRestrictions->cbSaltMin = SCOSSL_PROV_RSA_PSS_DEFAULT_SALTLEN_MIN;
 
         *pPssRestrictions = pssRestrictions;
     }
--- a/SymCryptProvider/src/p_scossl_rsa.c
+++ b/SymCryptProvider/src/p_scossl_rsa.c
@@ -8,10 +8,15 @@
 
 #include <openssl/core_names.h>
 #include <openssl/evp.h>
+#include <openssl/param_build.h>
+#include <openssl/proverr.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+#define SCOSSL_PROV_RSA_PSS_DEFAULT_MD 0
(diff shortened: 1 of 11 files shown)


SymCryptProvider/src/p_scossl_rsa.c
```diff
@@ -8,11 +8,16 @@
...
 #include <openssl/evp.h>
+#include <openssl/param_build.h>
+#include <openssl/proverr.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#define SCOSSL_PROV_RSA_PSS_DEFAULT_MD 0
+#define SCOSSL_PROV_RSA_PSS_DEFAULT_SATLLEN_MIN 20
```
samuel-lee-msft, line 19: >SATLLEN [](http://example.com/codeflow?start=36&length=7)

typo: SALTLEN

SymCryptProvider/src/p_scossl_rsa.c
```diff
@@ -8,11 +8,16 @@
...
 #include <openssl/core_names.h>
 #include <openssl/evp.h>
+#include <openssl/param_build.h>
+#include <openssl/proverr.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#define SCOSSL_PROV_RSA_PSS_DEFAULT_MD 0
```
samuel-lee-msft, line 18: >SCOSSL_PROV_RSA_PSS_DEFAULT_MD [](http://example.com/codeflow?start=8&length=30)

comment that this mean SHA1 / name indicates this is an index into supported MDs (I thought it could be an NID or something)

SymCryptProvider/src/p_scossl_rsa.c
samuel-lee-msft, line 19: >20 [](http://example.com/codeflow?start=48&length=2)

NP: (braces around #define-d numbers)

SymCryptProvider/src/signature/p_scossl_rsa_signature.c
```diff
@@ -372,8 +414,21 @@ static SCOSSL_STATUS p_scossl_rsa_set_ctx_params(_Inout_ SCOSSL_RSA_SIGN_CTX *ct
...
+            EVP_MD_free(md);
+            if (mdInfo->id != ctx->mdInfo->id)
+            {
+                ERR_raise(ERR_LIB_PROV, PROV_R_DIGEST_NOT_ALLOWED);
+                return SCOSSL_FAILURE;
+            }
+        }
+        else
+        {
+            ctx->md = md;
```
samuel-lee-msft, line 429: >ctx->md = md; [](http://example.com/codeflow?start=12&length=13)

Is there anything preventing an application calling set params with OSSL_SIGNATURE_PARAM_DIGEST repeatedly on a non-PSS ctx? Do we need to free `ctx->md` before setting it here to prevent leaking the old md?

SymCryptProvider/src/signature/p_scossl_rsa_signature.c
```diff
@@ -153,19 +160,48 @@ static SCOSSL_STATUS p_scossl_rsa_signverify_init(_Inout_ SCOSSL_RSA_SIGN_CTX *c
     ctx->operation = operation;
     if (keyCtx != NULL)
     {
+        if (keyCtx->pssRestrictions != NULL)
+        {
+            EVP_MD *md = NULL;
+            // ScOSSL does not support distinct MD and MGF1 MD
+            if (keyCtx->pssRestrictions->mdInfo != keyCtx->pssRestrictions->mgf1MdInfo ||
```
samuel-lee-msft, line 167: >keyCtx->pssRestrictions->mdInfo != keyCtx->pssRestrictions->mgf1MdInfo [](http://example.com/codeflow?start=16&length=70)

I don't see why the caller would have necessarily set the restrictions for both mdInfo and mgf1MdIndfo?
Can 1 be NULL and the other be a supported digest?

mamckee: Neither should be NULL, so this would be an error case. [RFC 8017](https://datatracker.ietf.org/doc/html/rfc8017#appendix-A.2.3) specifies default values for both as SHA1. 

We could fail this in keygen, but it isn't really an issue until we try to use the key, so we don't need to break keygen or unused import/export with distinct md and mgf1Md.

samuel-lee-msft: I see; the defaults are set the first time a `SCOSSL_RSA_PSS_RESTRICTIONS` struct is allocated, and then in copies of this the info fields can never be set to NULL. Makes sense.

samuel-lee-msft (dismissed):
LGTM modulo comments

samuel-lee-msft (approved):

PR #76, opened 2024-03-06
Provider KBKDF (SP800-108)

mamckee:
This PR adds the KBKDF (SP800-108) interface to the SymCrypt provider. The following are supported:
Counter mode:
- HMAC-SHA1
- HMAC-SHA256
- HMAC-SHA384
- HMAC-SHA512
- HMAC-SHA3-256
- HMAC-SHA3-384
- HMAC-SHA3-512
- CMAC-AES128
- CMAC-AES192
- CMAC-AES256

KMAC:
- KMAC128
- KMAC256

### DIFF
--- a/SymCryptProvider/src/kdf/p_scossl_kbkdf.c
+++ b/SymCryptProvider/src/kdf/p_scossl_kbkdf.c
@@ -4,67 +4,425 @@
 
 #include "p_scossl_base.h"
 
+#include <openssl/proverr.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+// MAC algorithm and additional data needed to determine correct PCSYMCRYPT_MAC
+// (digest type for HMAC, cipher for CMAC) may be set in multiple calls to p_scossl_kbkdf_set_ctx_params.
+// MAC type, digest, and cipher need to be tracked independently. ctx->pMac is only set when MAC type
+// and required additional parameters have been set.
+#define SCOSSL_MAC_TYPE_HMAC 0
+#define SCOSSL_MAC_TYPE_CMAC 1
+#define SCOSSL_MAC_TYPE_KMAC 2
+
 typedef struct
 {
-    OSSL_LIB_CTX *libctx;
+    OSSL_LIB_CTX *libCtx;
 
+    PBYTE pbKey;
+    SIZE_T cbKey;
+    PBYTE pbLabel;
+    SIZE_T cbLabel;
+    PBYTE pbContext;
+    SIZE_T cbContext;
+    PCSYMCRYPT_MAC pMac;
+
+    UINT macType;
+    SIZE_T cbCmacKey;
 } SCOSSL_PROV_KBKDF_CTX;
 
 static const OSSL_PARAM p_scossl_kbkdf_gettable_ctx_param_types[] = {
+    OSSL_PARAM_size_t(OSSL_KDF_PARAM_SIZE, NULL),
     OSSL_PARAM_END};
 
 static const OSSL_PARAM p_scossl_kbkdf_settable_ctx_param_types[] = {
+    OSSL_PARAM_octet_string(OSSL_KDF_PARAM_KEY, NULL, 0),
+    OSSL_PARAM_octet_string(OSSL_KDF_PARAM_SALT, NULL, 0), // Label
+    OSSL_PARAM_octet_string(OSSL_KDF_PARAM_INFO, NULL, 0), // Context
+    OSSL_PARAM_utf8_string(OSSL_KDF_PARAM_MAC, NULL, 0),
+    OSSL_PARAM_utf8_string(OSSL_KDF_PARAM_DIGEST, NULL, 0),
+    OSSL_PARAM_utf8_string(OSSL_KDF_PARAM_CIPHER, NULL, 0),
+    OSSL_PARAM_utf8_string(OSSL_KDF_PARAM_PROPERTIES, NULL, 0),
+    OSSL_PARAM_utf8_string(OSSL_KDF_PARAM_MODE, NULL, 0),
+    // Below parameters aren't configurable. The provider will reject anything that
+    // does not match the fixed behavior of SymCrypt.
+    OSSL_PARAM_int(OSSL_KDF_PARAM_KBKDF_USE_L, NULL),
+    OSSL_PARAM_int(OSSL_KDF_PARAM_KBKDF_USE_SEPARATOR, NULL),
+    OSSL_PARAM_int(OSSL_KDF_PARAM_KBKDF_R, NULL),
     OSSL_PARAM_END};
+
+static SCOSSL_STATUS p_scossl_kbkdf_set_ctx_params(_Inout_ SCOSSL_PROV_KBKDF_CTX *ctx, const _In_ OSSL_PARAM params[]);
 
 static SCOSSL_PROV_KBKDF_CTX *p_scossl_kbkdf_newctx(_In_ SCOSSL_PROVCTX *provctx)
 {
+    SCOSSL_PROV_KBKDF_CTX *ctx = OPENSSL_zalloc(sizeof(SCOSSL_PROV_KBKDF_CTX));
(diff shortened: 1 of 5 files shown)


SymCryptProvider/src/kdf/p_scossl_kbkdf.c
```diff
@@ -0,0 +1,495 @@
...
+    OSSL_PARAM_utf8_string(OSSL_KDF_PARAM_PROPERTIES, NULL, 0),
+    OSSL_PARAM_utf8_string(OSSL_KDF_PARAM_MAC, NULL, 0),
+    OSSL_PARAM_utf8_string(OSSL_KDF_PARAM_DIGEST, NULL, 0),
+    OSSL_PARAM_utf8_string(OSSL_KDF_PARAM_CIPHER, NULL, 0),
+    // Below parameters aren't configurable. The provider will reject anything that
+    // does not match the fixed behavior of SymCrypt.
+    OSSL_PARAM_utf8_string(OSSL_KDF_PARAM_MODE, NULL, 0),
+    OSSL_PARAM_int(OSSL_KDF_PARAM_KBKDF_USE_L, NULL),
+    OSSL_PARAM_int(OSSL_KDF_PARAM_KBKDF_USE_SEPARATOR, NULL),
+    OSSL_PARAM_int(OSSL_KDF_PARAM_KBKDF_R, NULL),
```
samuel-lee-msft, line 55: Not sure this one is in OpenSSL 3.1 - might need an ifdef?

SymCryptProvider/src/kdf/p_scossl_kbkdf.c
```diff
@@ -0,0 +1,495 @@
...
+    SCOSSL_PROV_KBKDF_CTX *copyCtx = OPENSSL_malloc(sizeof(SCOSSL_PROV_KBKDF_CTX));
+    if (copyCtx != NULL)
+    {
+        *copyCtx = *ctx;
+
+        if ((ctx->pbKey != NULL     && (copyCtx->pbKey = OPENSSL_memdup(ctx->pbKey, ctx->cbKey)) == NULL) ||
+            (ctx->pbLabel != NULL   && (copyCtx->pbLabel = OPENSSL_memdup(ctx->pbLabel, ctx->cbLabel)) == NULL) ||
+            (ctx->pbContext != NULL && (copyCtx->pbContext = OPENSSL_memdup(ctx->pbContext, ctx->cbContext)) == NULL))
+        {
+            p_scossl_kbkdf_freectx(copyCtx);
```
samuel-lee-msft, line 93: hrmmm

If a `OPENSSL_memdup` fails for `copyCtx->pbKey`, we will free `copyCtx`, which will then `OPENSSL_clear_free` the copied `pbLabel` and `pbContext` from the source struct...
Not sure if we already have a pattern like this elsewhere in the provider, but probably we should set `copyCtx->pb*` to NULL before attempting OPENSSL_memdup, or something equivalent.

SymCryptProvider/src/kdf/p_scossl_kbkdf.c
```diff
@@ -0,0 +1,495 @@
...
+    OPENSSL_clear_free(ctx->pbLabel, ctx->cbLabel);
+    OPENSSL_clear_free(ctx->pbContext, ctx->cbContext);
+    OPENSSL_cleanse(ctx, sizeof(SCOSSL_PROV_KBKDF_CTX));
+
+    ctx->libCtx = libCtx;
+
+    return SCOSSL_SUCCESS;
+}
+
+// KMAC KBKDF is a special case. Pass the context as the input and lebel as the customization string.
```
samuel-lee-msft, line 116: NP: label

SymCryptProvider/src/kdf/p_scossl_kbkdf.c
mamckee: Yep, I'll wrap this and the parameter set code.

SymCryptProvider/src/kdf/p_scossl_kbkdf.c
mamckee: Right, we have something similar in DH that isn't prone to this failure. I'll copy what we have there.

SymCryptProvider/src/kdf/p_scossl_kbkdf.c
```diff
@@ -0,0 +1,503 @@
...
+    {
+        OPENSSL_clear_free(ctx->pbKey, ctx->cbKey);
+
+        if (!OSSL_PARAM_get_octet_string(p, (void **) &ctx->pbKey, 0, &ctx->cbKey))
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_GET_PARAMETER);
+            goto cleanup;
+        }
+
+        // Check that key size matches expected if cmac is intitialized
```
samuel-lee-msft, line 257: >intitialized [](http://example.com/codeflow?start=59&length=12)

NP: initialized

SymCryptProvider/src/kdf/p_scossl_kbkdf.c
```diff
@@ -0,0 +1,503 @@
...
+        if (ctx->pMac == SymCryptAesCmacAlgorithm &&
+            ctx->cbKey != ctx->cbCmacKey)
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_GET_PARAMETER);
```
samuel-lee-msft, line 261: >PROV_R_FAILED_TO_GET_PARAMETER [](http://example.com/codeflow?start=36&length=30)

Perhaps PROV_R_INVALID_KEY_LENGTH?

SymCryptProvider/src/kdf/p_scossl_kbkdf.c
```diff
@@ -0,0 +1,503 @@
...
+
+        if (!OSSL_PARAM_get_octet_string(p, (void **) &ctx->pbKey, 0, &ctx->cbKey))
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_GET_PARAMETER);
+            goto cleanup;
+        }
+
+        // Check that key size matches expected if cmac is intitialized
+        if (ctx->pMac == SymCryptAesCmacAlgorithm &&
+            ctx->cbKey != ctx->cbCmacKey)
```
samuel-lee-msft, line 259: >ctx->cbKey != ctx->cbCmacKey [](http://example.com/codeflow?start=12&length=28)

Technically it looks like this may be over-eager.

i.e. If I set Mac -> CMAC but don't define the key size yet (cbCmacKey == 0) in one set params call, and then I set key and CMAC cipher in a second call, the second call would fail?

Maybe just check for equality of cbKey and cbCmacKey when cbCmacKey is non-zero?

samuel-lee-msft (dismissed):
LGTM modulo comments

samuel-lee-msft:
Slight nit in the PR description; I think this also adds support for SP800-108 counter mode with HMAC-SHA3-256, HMAC-SHA3-384, and HMAC-SHA3-512

SymCryptProvider/src/kdf/p_scossl_kbkdf.c
```diff
@@ -0,0 +1,503 @@
...
+    EVP_CIPHER *cipher = NULL;
+    EVP_MAC *mac = NULL;
+    SCOSSL_STATUS ret = SCOSSL_FAILURE;
+    const OSSL_PARAM *p;
+
+    if ((p = OSSL_PARAM_locate_const(params, OSSL_KDF_PARAM_KEY)) != NULL)
+    {
+        OPENSSL_clear_free(ctx->pbKey, ctx->cbKey);
+
+        if (!OSSL_PARAM_get_octet_string(p, (void **) &ctx->pbKey, 0, &ctx->cbKey))
```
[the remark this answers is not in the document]
mamckee: Yes, we should be setting ctx->pbKey to NULL again here in case this is reset.

samuel-lee-msft (approved):

PR #77, opened 2024-04-09
Update tests for OpenSSL 3 and fix bugs found in test

mamckee:
This PR updates existing tests for OpenSSL 3, and fixes any bugs found in testing
- Added test cases to be run with the OpenSSL evp_test utility
  - Test cases for both OpenSSL 1.1.1 and OpenSSL 3.0
  - Most test cases are unmodified from the OpenSSL source. The test cases that are modified are due to known implementation differences with SymCrypt or unsupported configurations (e.g. omitting L/separator in SP800-108 or enforcing RSA parameter null termination)
  - Added a readme with instructions on how to test the SymCrypt provider and engine with these test files
- Remove engine EVP interface for OpenSSL 3 to prevent intercepting calls that should be routed to the provider. 
- Updated SSLPlay to pass in OpenSSL 3 with the engine and/or provider
 - Support variable-length IVs in the provider AES-GCM
   - A scenario exists in the engine that prevents this from working. EVP_CIPHER_CTX_iv_length calls into the engine, so the engine has no way of knowing the IV length for an IV passed to e_scossl_aes_gcm_init_key. The engine will continue to reject any IVs that aren't 12-bytes.
- Fixed an illegal cast i the engine RSA implementation
- Fixed expected behavior differences in the provider ECC key management, ECDSA, RSA sign and decrypt that caused test failures
- Fixed missing `OPENSSL_strcasecmp` preventing the provider from running with OpenSSL versions below 3.0.4, when this API was exported


mlindgren:
> 	    message(FATAL_ERROR "Provide symcrypt pkg-config or set SYMCRYPT_ROOT_DIR variable")

np: tabs

---
Refers to: CMakeLists.txt:19 in 55d107f. [](commit_id = 55d107fb4626a97dcb778ffeba4846c8f3c54615, deletion_comment = False)

ScosslCommon/inc/scossl_aes_aead.h
```diff
@@ -37,6 +37,7 @@ extern "C" {
 //  call to EVP_CTRL_GCM_SET_IV_FIXED)
 
 #define SCOSSL_GCM_IV_LENGTH (12)
```
mlindgren, line 39: This should no longer be necessary, right?

ScosslCommon/src/scossl_aes_aead.c
```diff
@@ -34,18 +35,17 @@ SCOSSL_STATUS scossl_aes_gcm_init_key(SCOSSL_CIPHER_GCM_CTX *ctx,
     SYMCRYPT_ERROR scError = SYMCRYPT_NO_ERROR;
 
     ctx->operationInProgress = 0;
-    if (iv)
+    if (iv != NULL)
     {
-        if (ivlen != SCOSSL_GCM_IV_LENGTH)
+        if (!scossl_aes_gcm_set_iv_len(ctx, ivlen) ||
+            (ctx->iv = OPENSSL_memdup(iv, ctx->ivlen)) == NULL)
```
mlindgren, line 41: This is probably just an OpenSSL quirk I'm unaware of, but why does the IV have to be set again here when it's also set in `scossl_aes_gcm_init_ctx`?

ScosslCommon/src/scossl_rsa.c
```diff
@@ -433,10 +433,6 @@ SCOSSL_STATUS scossl_rsa_encrypt(PSYMCRYPT_RSAKEY key, UINT padding,
         // cbDst is not caller supplied for engine
         cbDst = cbModulus;
     }
-    else if (cbDst < cbModulus)
-    {
-        goto cleanup;
-    }
```
mlindgren, line 439: Why is this no longer required?

SslPlay/SslPlay.cpp
```diff
@@ -987,6 +992,9 @@ int CreateKeys(int id, int modulus, uint32_t exponent, char* publicFileName, cha
         EVP_PKEY_CTX_free(pKeyContext);
     if (pKey)
         EVP_PKEY_free(pKey);
+#if OPENSSL_VERSION_MAJOR == 3
+    BN_free(exponent_bn);
+#endif
```
mlindgren, line 997: This is allocated for both versions. Is it freed with pKeyContext in 1.1.1 but not in 3.x?

SslPlay/SslPlay.cpp
```diff
@@ -1033,7 +1041,9 @@ void TestRsaEvp(int modulus, uint32_t exponent)
     //
     // Sign/Verify
     //
+#if OPENSSL_VERSION_MAJOR == 1
```
mlindgren, line 1044: Why only for 1.1.1? In general, it would be helpful if you could add comments to these cases where there are behavior differences between 1.1.1 and 3.x to indicate what they are and why the tests have to be have differently.

mlindgren:
>  A scenario exists in the engine that prevents this from working. EVP_CIPHER_CTX_iv_length calls into the engine, so the engine has no way of knowing the IV length for an IV passed to e_scossl_aes_gcm_init_key. The engine will continue to reject any IVs that aren't 12-bytes.

I don't follow this comment. Can't the engine save the IV length and use that for subsequent calls?

mlindgren (approved):
Left a few suggestions, but nothing blocking.

ScosslCommon/inc/scossl_aes_aead.h
mamckee: Yes, still needed for the engine but the name change makes sense.

ScosslCommon/src/scossl_aes_aead.c
mamckee: It's optional here. I think the original implementation of the engine was actually getting ahead of itself by setting the IV in EVP_CTRL_INIT. The IV should have been passed to e_scossl_aes_gcm_init_key too but there may have been cases where nothing was passed.

SslPlay/SslPlay.cpp
mamckee: Provider doesn't support MD5 for PKCS1 sign. I'll add a comment.

SslPlay/SslPlay.cpp
mamckee: In 1.1.1, the EVP_PKEY_CTX takes ownership of this bignum and will free it in EVP_PKEY_CTX_free. In 3.0, it's the caller's responsibility. I'll add that as a comment

ScosslCommon/src/scossl_rsa.c
mamckee: We should leave this check up to SymCrypt. This function wraps 3 different encrypt calls. SymCryptRsaRawEncrypt does support cbDst < cbModulus so some of the tests failed.

mamckee:
> > A scenario exists in the engine that prevents this from working. EVP_CIPHER_CTX_iv_length calls into the engine, so the engine has no way of knowing the IV length for an IV passed to e_scossl_aes_gcm_init_key. The engine will continue to reject any IVs that aren't 12-bytes.
> 
> I don't follow this comment. Can't the engine save the IV length and use that for subsequent calls?

It depends on EVP_CTRL_AEAD_SET_IVLEN being set before EVP_CTRL_INIT. We might be able to rework this (i.e. not setting the IV 
 until e_scossl_aes_gcm_init_key) but I don't want to spend a ton of time on adding new behavior to the engine. I'm sure there was a reason the IV was set in EVP_CTRL_INIT so it's something I'm hesitant to modify unless we need to.

mlindgren (approved):
:shipit:

PR #78, opened 2024-04-24
Bugfixes for TLS connections

mamckee:
Additional testing at the TLS layer uncovered a number of bugs that must be fixed before further rollout. These bugs were not surfaced when testing at the crypto layer.

- Fixed wrong pointer returned from kmac_dupctx
- Fixed incorrect return value check in dh_kmgmt_export 
- Support algorithm ID fetch in RSA and ECDSA signing interfaces
  - TLS layer will fail without this
- Support getting/setting OSSL_PKEY_PARAM_EC_ENCODING
  - Some TLS layer functions will explicitly set this and fail if the provider doesn't expose this parameter
  - Only OSSL_PKEY_EC_ENCODING_GROUP is allowed, so this doesn't change any behavior
- Allow DH group to be set without key data and support 
- Implement OSSL_FUNC_KEYMGMT_GEN_SET_TEMPLATE for DH
- Support setting OSSL_PKEY_PARAM_ENCODED_PUBLIC_KEY for DH
  - libssl imports the peer key group without the key data to a key object. The peer key object is passed to the OSSL_FUNC_KEYMGMT_GEN_SET_TEMPLATE function for keygen. The peer key data is set in the peer key object through the OSSL_PKEY_PARAM_ENCODED_PUBLIC_KEY parameter instead of import.
- Support TLS mode for AES block ciphers
  - OpenSSL 3 moved the responsibility of padding and MAC fetching to the providers. In TLS mode, each record is independently padded and encrypted/decrypted by a call to OSSL_FUNC_CIPHER_UPDATE. The provider must also remove and store the MAC if present. This is done in the provider using an implementation adapted from OpenSSL.


SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
```diff
@@ -15,6 +15,9 @@ extern "C" {
 #define SCOSSL_DH_KEYHEN_POSSIBLE_SELECTIONS (OSSL_KEYMGMT_SELECT_KEYPAIR | OSSL_KEYMGMT_SELECT_DOMAIN_PARAMETERS)
 
 #define SCOSSL_DH_PBITS_DEFAULT 2048
+// Private key length determined by group
+// Setting this to -1 matches the OpenSSL implementation for paramter fetching
```
mlindgren, line 19: >paramter [](http://example.com/codeflow?start=61&length=8)

np: parameter

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
```diff
@@ -28,8 +31,8 @@ typedef struct
 {
     SCOSSL_PROVCTX *provCtx;
     PCSYMCRYPT_DLGROUP pDlGroup;
-    SIZE_T pbits;
-    UINT32 nBitsPriv;
+    SIZE_T nBitsPub;
+    int nBitsPriv;
```
mlindgren, line 35: Kind of weird for these to be different types. Could they be the same type?

mamckee: I agree, but this is a design decision by OpenSSL. The OSSL_PKEY_PARAM_DH_PRIV_LEN is an int, while OSSL_PKEY_PARAM_FFC_PBITS is a SIZE_T. OpenSSL expects this to be -1 if unset in export as well, while pbits should be set to some default (2048). I renamed pbits to nBitsPub so the relationship was a little more clear.

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
```diff
@@ -146,12 +178,63 @@ static void p_scossl_dh_keymgmt_free_key_ctx(_In_ SCOSSL_PROV_DH_KEY_CTX *ctx)
...
+            NULL,
+            NULL);
+
+        return cbPrimeP * 8;
+    }
+
+    return -1;
+}
+
+static int p_scossl_dh_privkey_bits(SCOSSL_PROV_DH_KEY_CTX *ctx)
```
mlindgren, line 209: >p_scossl_dh_privkey_bits [](http://example.com/codeflow?start=11&length=24)

Having multiple places to get the key size from seems risky--there's a possibility that they could get out of sync. How much trouble would it be to not store the sizes in the context and just always use these functions to get them?

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
```diff
@@ -146,12 +178,63 @@ static void p_scossl_dh_keymgmt_free_key_ctx(_In_ SCOSSL_PROV_DH_KEY_CTX *ctx)
...
+        SIZE_T cbPrimeP;
+
+        SymCryptDlgroupGetSizes(
+            ctx->pDlGroup,
+            &cbPrimeP,
+            &cbPrimeQ,
+            NULL,
+            NULL);
+
+        return cbPrimeQ != 0 ? cbPrimeQ * 8 - 1  : cbPrimeP * 8 - 1;
```
mlindgren, line 232: >- 1 [](http://example.com/codeflow?start=44&length=3)

Why -1?

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
```diff
@@ -461,17 +536,45 @@ static SCOSSL_DH_KEYGEN_CTX *p_scossl_dh_keygen_init(_In_ SCOSSL_PROVCTX *provCt
...
+    if (genCtx == NULL ||
+        tmplCtx == NULL ||
+        tmplCtx->groupSetByParams)
+    {
+        return SCOSSL_FAILURE;
+    }
+
+    if (tmplCtx->pDlGroup != NULL)
+    {
+        genCtx->pDlGroup = tmplCtx->pDlGroup;
```
mlindgren, line 550: >genCtx->pDlGroup = tmplCtx->pDlGroup; [](http://example.com/codeflow?start=8&length=37)

Who owns this DlGroup? Do you need to copy it?

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
```diff
@@ -503,32 +606,92 @@ static SCOSSL_PROV_DH_KEY_CTX *p_scossl_dh_keygen(_In_ SCOSSL_DH_KEYGEN_CTX *gen
...
+        {
+            SymCryptDlkeyFree(ctx->keyCtx->dlkey);
+            ctx->keyCtx->initialized = FALSE;
+        }
+
+        if ((ctx->keyCtx->dlkey = SymCryptDlkeyAllocate(ctx->pDlGroup)) == NULL)
+        {
+            ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+            return SCOSSL_FAILURE;
+        }
```
mlindgren, line 651: What happens if you call this function with an encoded key from a different group than the one set in the context?

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
```diff
@@ -885,11 +1075,14 @@ static BOOL p_scossl_dh_keymgmt_match(_In_ SCOSSL_PROV_DH_KEY_CTX *ctx1, _In_ SC
...
+        // The below ensures the appropriate pointers are set for,
+        // SymCryptDlkeyGetValue and memcmp and the others are NULL.
+        pbPublicKey1 = cbPublicKey == 0 ? NULL : pbData;
         pbPrivateKey1 = cbPrivateKey == 0 ? NULL : pbData;
-        pbPrivateKey2 = cbPrivateKey == 0 ? NULL : pbData + cbPrivateKey;
 
-        pbPublicKey1 = cbPublicKey == 0 ? NULL : pbData + cbPrivateKey * 2;
-        pbPublicKey2 = cbPublicKey == 0 ? NULL : pbData + cbPrivateKey * 2 + cbPublicKey;
+        pbPublicKey2 = cbPublicKey == 0 ? NULL : pbData + cbPublicKey + cbPrivateKey;
+        pbPrivateKey2 = cbPrivateKey == 0 ? NULL : pbData + cbPublicKey + cbPrivateKey;;
```
mlindgren, line 1085: >; [](http://example.com/codeflow?start=87&length=1)

np: extra semicolon

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
```diff
@@ -1275,7 +1453,7 @@ static SCOSSL_STATUS p_scossl_dh_keymgmt_export(_In_ SCOSSL_PROV_DH_KEY_CTX *ctx
             }
 
             if (BN_bin2bn(pbPrivateKey, cbPrivateKey, bnPrivKey) == NULL ||
-                OSSL_PARAM_BLD_push_BN(bld, OSSL_PKEY_PARAM_PRIV_KEY, bnPrivKey))
+                !OSSL_PARAM_BLD_push_BN(bld, OSSL_PKEY_PARAM_PRIV_KEY, bnPrivKey))
```
mlindgren, line 1456: >!OSSL_PARAM_BLD_push_BN(bld, OSSL_PKEY_PARAM_PRIV_KEY, bnPrivKey)) [](http://example.com/codeflow?start=16&length=66)

Do we have test cases for this function? Seems like this wouldn't have worked previously

ScosslCommon/src/scossl_dh.c
```diff
@@ -81,7 +88,44 @@ SCOSSL_DH_KEY_CTX *scossl_dh_dup_key_ctx(SCOSSL_DH_KEY_CTX *ctx, BOOL copyGroup)
             }
             else
             {
```
mlindgren, line 90: np: This block is redundant since the above if results in `goto cleanup`. I think it's a bit easier to read without the extra block.

ScosslCommon/src/scossl_dh.c
```diff
@@ -81,7 +88,44 @@ SCOSSL_DH_KEY_CTX *scossl_dh_dup_key_ctx(SCOSSL_DH_KEY_CTX *ctx, BOOL copyGroup)
...
-                SymCryptDlkeyCopy(ctx->dlkey, copyCtx->dlkey);
+                // SymCryptDlkeyCopy expects the private key to be allocated
+                // The private key is only allocated with a call to SymCryptDlkeySetValue
+                // or SymCryptDlkeyGenerate, so instead we need to copy the key
```
mlindgren, line 93: What about `SymCryptDlkeyAllocate` followed by `SymCryptDlkeyCopy`?

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
mamckee: You're right, this should be copied of the group was set by params.

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
mamckee: The problem here is we need to track the bit length of the private key, which isn't available in the SymCrypt API. Ideally, we'd have an API in SymCrypt to fetch SYMCRYPT_DLKEY's nBitsPriv or SYMCRYPT_DLGROUP's nBitsOfQ.

SymCryptDlkeySizeofPrivateKey returns the number of bytes needed to contain the private key, but the OSSL_PKEY_PARAM_DH_PRIV_LEN parameter needs the maximum number of bits for Q, and those are only equal if the bit length is a multiple of 8. The default max bits of Q is bits of P-1 so this is usually not the case.

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
mamckee: This needs to be copied if it was generated.

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
mamckee: SymCryptDlkeySetValue should fail, and the key will be cleaned up.

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
mamckee: It didn't get exercised before. I found this when generating and saving DH keys to a file. That's really the only case when this would this function is used. Normal key exchange operations get the key material through p_scossl_dh_keymgmt_get_params.

ScosslCommon/src/scossl_dh.c
mamckee: The private key inside the SYMCRYPT_DLKEY is normally allocated in SymCryptDlkeyGenerate or SymCryptDlkeySetValue. SymCryptDlkeyCopy expects the private key to already by allocated so SymCryptDlkeyAllocate -> SymCryptDlkeyCopy causes a seg fault.

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -112,15 +131,129 @@ static SCOSSL_STATUS p_scossl_aes_generic_encrypt_init(_Inout_ SCOSSL_AES_CTX *c
...
+
+//
+// From include/internal/contant_time.h
+// Returns the value unmodified, but avoids optimizations.
+// The barriers prevent the compiler from narrowing down the
+// possible value range of the mask and ~mask in the select
+// statements, which avoids the recognition of the select
+// and turning it into a conditional load or branch.
+//
+static ossl_inline BYTE value_barrier_8(BYTE a)
```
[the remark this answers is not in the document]
mamckee: SYMCRYPT_FORCE_READ8 works so I'll use that. Better to use as much SymCrypt provided functionaliy as possible.

SymCryptProvider/src/keymgmt/p_scossl_dh_keymgmt.c
mlindgren: Would be good to add tests for this, but not blocking for this PR

mlindgren (approved):
:shipit:

PR #80, opened 2024-05-15
Provider and Engine bugfixes

### LINKED ISSUE #79: Git clone failing on AzureLinux 3.0 for https://git.kernel.org hosted repos with TLS IV length error
In Azure Linux 3.0 we recently (on 25th April 2024) [updated](https://github.com/microsoft/azurelinux/pull/8900) the SymCrypt-OpenSSL package to latest 1.4.2, since then we are having this issue where, git clone is failing on any repo hosted on `https://git.kernel.org`

Log:
```
$ git clone https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/
Cloning into 'linux-stable'...
[ERROR] error:41080106:SCOSSL::passed invalid argument:set_iv_fixed only works with TLS IV length at /usr/src/azl/BUILD/SymCrypt-OpenSSL-1.4.2/ScosslCommon/src/scossl_aes_aead.c, line 305
fatal: unable to access 'https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/': OpenSSL/3.3.0: error:41080106:SCOSSL::passed invalid argument
```
This doesn't occur when the clone url starts with `git://`
```
$ git clone git://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/
Cloning into 'linux-stable'...
remote: Enumerating objects: 12393993, done.
remote: Counting objects: 100% (64/64), done.
remote: Compressing objects: 100% (25/25), done.
^Cceiving objects:   4% (593148/12393993), 286.92 MiB | 27.45 MiB/s
```

After removing this package git clone also works with `https://` prefix.

Note: This doesn't affect clone operations from `https://github.com/`

mamckee:
- Various bugfixes for the provider TLS, mostly due to unsupported parameters
- Add variable length IV support for engine GCM
- Removed code to use alloc functions passed to the provider in init. In some cases, the CRYPTO_* functions defined in the provider were passed in, leading to infinite recursion
- Fixed memory leaks in DH and RSA key management

Fixes #79


SymCryptEngine/src/e_scossl_ciphers.c
```diff
@@ -254,6 +254,7 @@ static EVP_CIPHER *_hidden_aes_192_gcm = NULL;
 static const EVP_CIPHER *e_scossl_aes_192_gcm(void)
 {
     if( (_hidden_aes_192_gcm = EVP_CIPHER_meth_new(NID_aes_192_gcm, 1, AES_192_KEY_SIZE)) == NULL
+        || !EVP_CIPHER_meth_set_iv_length(_hidden_aes_192_gcm, SCOSSL_GCM_DEFAULT_IV_LENGTH)
```
samuel-lee-msft, line 257: >!EVP_CIPHER_meth_set_iv_length(_hidden_aes_192_gcm, SCOSSL_GCM_DEFAULT_IV_LENGTH) [](http://example.com/codeflow?start=11&length=81)

Also set this on `_hidden_aes_128_gcm` above?

SymCryptEngine/src/e_scossl_ciphers.c
```diff
@@ -743,19 +743,13 @@ static int e_scossl_aes_gcm_ctrl(_Inout_ EVP_CIPHER_CTX *ctx, int type, int arg,
...
     {
     case EVP_CTRL_INIT:
-        return scossl_aes_gcm_set_iv_len(cipherCtx, EVP_CIPHER_CTX_iv_length(ctx)) && scossl_aes_gcm_init_ctx(cipherCtx, EVP_CIPHER_CTX_iv(ctx));
+        return scossl_aes_gcm_init_ctx(cipherCtx, EVP_CIPHER_CTX_iv(ctx));
     case EVP_CTRL_GET_IVLEN:
-        *(int *)ptr = SCOSSL_GCM_DEFAULT_IV_LENGTH;
+        *(int *)ptr = cipherCtx->ivlen;
         break;
     case EVP_CTRL_AEAD_SET_IVLEN:
         // SymCrypt engine currently only supports SCOSSL_GCM_DEFAULT_IV_LENGTH
```
samuel-lee-msft, line 751: >// SymCrypt engine currently only supports SCOSSL_GCM_DEFAULT_IV_LENGTH [](http://example.com/codeflow?start=8&length=71)

Remove comment

SymCryptProvider/src/ciphers/p_scossl_aes_aead.c
```diff
@@ -560,7 +561,7 @@ static SCOSSL_STATUS p_scossl_aes_ccm_set_ctx_params(_Inout_ SCOSSL_CIPHER_CCM_C
     SCOSSL_STATUS p_scossl_aes_##kbits##_##lcmode##_get_params(_Inout_ OSSL_PARAM params[])                  \
     {                                                                                                        \
         return p_scossl_aes_generic_get_params(params, EVP_CIPH_##UCMODE##_MODE, kbits >> 3,                 \
-                                               ivlen, 1, SCOSSL_FLAG_AEAD | SCOSSL_FLAG_CUSTOM_IV);          \
+                                               defaultIvLen, 1, SCOSSL_FLAG_AEAD | SCOSSL_FLAG_CUSTOM_IV);   \
```
samuel-lee-msft, line 564: >defaultIvLen [](http://example.com/codeflow?start=47&length=12)

`ctx->ivlen` seems more correct here?

samuel-lee-msft: Oh I see, this is getting params for the cipher, not a specific context which may modify the default.

SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
```diff
@@ -531,187 +554,352 @@ static UINT16 p_scossl_rsa_get_security_bits(_In_ PSYMCRYPT_RSAKEY keydata)
...
+    {
+        ERR_raise(ERR_LIB_PROV, ERR_R_INTERNAL_ERROR);
+        goto cleanup;
+    }
+
+    // Convert buffers to BIGNUMs and set parameters
+    if (paramModulus != NULL)
+    {
+        if (BN_bin2bn(pbModulus, cbModulus, bnModulus) == NULL ||
+            !OSSL_PARAM_set_BN(paramModulus, bnModulus))
```
samuel-lee-msft, line 661: This is fine, but I'm kind of surprised you can't directly set this with something like `OSSL_PARAM_set_octet_string`, rather than converting through OpenSSL BNs

SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
```diff
@@ -531,187 +554,352 @@ static UINT16 p_scossl_rsa_get_security_bits(_In_ PSYMCRYPT_RSAKEY keydata)
...
+    // This parameter gets ignored for restricted PSS keys
+    if ((p = OSSL_PARAM_locate(params, OSSL_PKEY_PARAM_DEFAULT_DIGEST)) != NULL &&
+         keyCtx->pssRestrictions == NULL &&
+         !OSSL_PARAM_set_utf8_string(p, SCOSSL_RSA_DEFAULT_DIGEST))
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_SET_PARAMETER);
+        return SCOSSL_FAILURE;
+    }
+
+    return p_scossl_rsa_keymgmt_get_keydata(keyCtx, params) &&
```
samuel-lee-msft, line 899: >&& [](http://example.com/codeflow?start=60&length=2)

Not a huge fan of using && here, would be clearer and less error prone I think to have a local SCOSSL_STATUS variable, and something like:

```
SCOSSL_STATUS ret = SCOSSL_FAILURE;

...

ret = p_scossl_rsa_keymgmt_get_keydata(keyCtx, params);
if( ret != SCOSSL_SUCCESS )
{
  return ret;
}

return p_scossl_rsa_keymgmt_get_crt_keydata(keyCtx, params);
```

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.13.0)
 
 project(SymCrypt-OpenSSL
-    VERSION 1.4.1
+    VERSION 1.4.3
```
mlindgren, line 4: >VERSION 1.4.3 [](http://example.com/codeflow?start=4&length=13)

What happened to .2? 😛

ScosslCommon/src/scossl_aes_aead.c
```diff
@@ -14,15 +14,17 @@ extern "C" {
...
-        return SCOSSL_FAILURE;
-    }
     ctx->operationInProgress = 0;
     ctx->taglen = SCOSSL_GCM_MAX_TAG_LENGTH;
     ctx->tlsAadSet = 0;
     ctx->ivInvocation = 0;
     ctx->useInvocation = 0;
+    ctx->ivlen = SCOSSL_GCM_DEFAULT_IV_LENGTH;
+
+    if (iv != NULL && (ctx->iv = OPENSSL_memdup(iv, ctx->ivlen)) == NULL)
```
mlindgren, line 24: >OPENSSL_memdup(iv, ctx->ivlen) [](http://example.com/codeflow?start=33&length=30)

Is the IV passed in here guaranteed to be 12 bytes?

SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
```diff
@@ -531,187 +554,352 @@ static UINT16 p_scossl_rsa_get_security_bits(_In_ PSYMCRYPT_RSAKEY keydata)
...
-    PBYTE pbPrivateExponent = NULL;
-    PBYTE pbCrtCoefficient = NULL;
+    UINT32 cbModulus = 0;
+    BIGNUM *bnModulus = NULL;
+
+    OSSL_PARAM *paramPublicExponent;
+    UINT64 publicExponent;
+    PUINT64 pPublicExponent = NULL;
+    UINT32 nPublicExponent = 0;
+    BIGNUM *bnPublicExponent = NULL;
```
samuel-lee-msft, line 572: For public exponent it really makes sense to unconditionally get it from SymCrypt, and pass it to OSSL_PARAM with `OSSL_PARAM_set_uint64`.

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -370,45 +366,26 @@ SCOSSL_STATUS OSSL_provider_init(_In_ const OSSL_CORE_HANDLE *handle,
...
-        case OSSL_FUNC_CRYPTO_ZALLOC:
-            c_CRYPTO_zalloc = OSSL_FUNC_CRYPTO_zalloc(in);
-            break;
-        case OSSL_FUNC_CRYPTO_FREE:
-            c_CRYPTO_free = OSSL_FUNC_CRYPTO_free(in);
-            break;
-        case OSSL_FUNC_CRYPTO_CLEAR_FREE:
-            c_CRYPTO_clear_free = OSSL_FUNC_CRYPTO_clear_free(in);
-            break;
-        case OSSL_FUNC_CORE_GET_LIBCTX:
```
mlindgren, line 403: What were these used for before, and why were they removed?

SymCryptProvider/src/asymcipher/p_scossl_rsa_cipher.c
```diff
@@ -178,7 +179,7 @@ static SCOSSL_STATUS p_scossl_rsa_cipher_decrypt(_In_ SCOSSL_RSA_CIPHER_CTX *ctx
                              out, &cbResult, outsize);
     *outlen = ret ? (SIZE_T)cbResult : 0;
 
-    return ret;
+    return (ctx->padding == RSA_PKCS1_WITH_TLS_PADDING) || ret;
```
samuel-lee-msft, line 182: >(ctx->padding == RSA_PKCS1_WITH_TLS_PADDING) || [](http://example.com/codeflow?start=11&length=47)

Similar comment about confusing SCOSSL_STATUS with a boolean value. Clearer to rephrase as:

```
if(ctx->padding == RSA_PKCS1_WITH_TLS_PADDING)
    return SCOSSL_SUCCESS;
else
    return ret;
```

mlindgren, line 182: >(ctx->padding == RSA_PKCS1_WITH_TLS_PADDING) [](http://example.com/codeflow?start=11&length=44)

Why is `ret` no longer relevant when `(ctx->padding == RSA_PKCS1_WITH_TLS_PADDING)`?

SymCryptProvider/src/mac/p_scossl_hmac.c
```diff
@@ -82,9 +82,49 @@ static SCOSSL_STATUS p_scossl_hmac_get_ctx_params(_In_ SCOSSL_MAC_CTX *ctx, _Ino
...
+        {
+            mdNid = NID_sha256;
+        }
+        if (EVP_MD_is_a(md, SN_sha3_384))
+        {
+            mdNid = NID_sha384;
+        }
+        if (EVP_MD_is_a(md, SN_sha3_512))
+        {
+            mdNid = NID_sha512;
```
mlindgren, line 118: Are these supposed to be the same as the SHA2 NIDs?

SymCryptProvider/src/signature/p_scossl_rsa_signature.c
```diff
@@ -677,15 +699,53 @@ static ASN1_STRING *p_scossl_rsa_pss_params_to_asn1_sequence(_In_ SCOSSL_RSA_SIG
...
+    {
+        if (ctx->keyCtx == NULL ||
+            !ctx->keyCtx->initialized)
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_NO_KEY_SET);
+            goto cleanup;
+        }
+
+        cbHash = scossl_get_expected_hash_length(ctx->mdInfo->id);
+        cbSaltMax = ((SymCryptRsakeyModulusBits(ctx->keyCtx->key) + 6) / 8) - cbHash - 2; // ceil((ModulusBits - 1) / 8) - cbDigest - 2
```
mlindgren, line 715: >6 [](http://example.com/codeflow?start=68&length=1)

Why +6? Shouldn't it be +7?

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -1304,13 +1411,264 @@ const OSSL_DISPATCH p_scossl_x25519_keymgmt_functions[] = {
...
+                ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+                goto cleanup;
+            }
+
+            if (BN_bin2bn(pbPublicKey + (cbPublicKey/2), cbPublicKey/2, ecPubY) == NULL)
+            {
+                goto cleanup;
+            }
+
+            if (BN_is_odd(ecPubY))
```
samuel-lee-msft, line 1509: >if (BN_is_odd(ecPubY)) [](http://example.com/codeflow?start=12&length=22)

No need to create `ecPubY`, this line is equivalent to:
`if (pbPublicKey[cbPublicKey-1] & 1)`

SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
samuel-lee-msft: Hrmm, I guess the reason to convert to BIGNUM is that OpenSSL expect providers to give a BIGNUM for these parameters. For some reason I was expecting that OSSL_PARAM would just be a byte blob internally and set_BN was a convenience wrapper, but [https://www.openssl.org/docs/man3.1/man3/OSSL_PARAM_set_BN.html](https://www.openssl.org/docs/man3.1/man3/OSSL_PARAM_set_BN.html) states:
"Native types will be converted as required only if the value is exactly representable by the target type or parameter. Apart from that, the functions must be used appropriately for the expected type of the parameter."


I would expect that it is generally more efficient to get the byte buffers little endian (SYMCRYPT_NUMBER_FORMAT_LSB_FIRST), and place them in BIGNUMs with `BN_lebin2bn` (certainly avoids reversing bytes coming out of SymCrypt, and probably avoids reversing bytes going into OpenSSL's BIGNUMs)

SymCryptProvider/src/signature/p_scossl_rsa_signature.c
samuel-lee-msft: I think we have this construct elsewhere in PSS code

ceil((ModulusBits - 1) / 8) == floor(((ModulusBits - 1) + 7) / 8) == floor((ModulusBits + 6) / 8) == (ModulusBits + 6) // 8

samuel-lee-msft:
>                 !scossl_ec_point_to_pubkey(ecPoint, ecGroup, bnCtx, pbPublicKey, cbPublicKey))

It's a little odd that we always use OpenSSL functions to convert compressed points into uncompressed points to ingest into SymCrypt, but roll our own logic to convert from uncompressed points to compressed points.

Should we verify that the encoded point provided matches keyCtx->conversionFormat on ingesting a public key blob?

---
Refers to: SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c:668 in 49d8b21. [](commit_id = 49d8b21c12c37ee0c8d94fbbc0f90c04259c2975, deletion_comment = False)

SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c
```diff
@@ -1304,13 +1411,264 @@ const OSSL_DISPATCH p_scossl_x25519_keymgmt_functions[] = {
...
+        ERR_raise(ERR_LIB_PROV, ERR_R_INTERNAL_ERROR);
+        goto cleanup;
+    }
+
+    if (!keyCtx->isX25519)
+    {
+        pbPublicKeyStart[0] = keyCtx->conversionFormat;
+
+        // Only prime finite fields are supported, so the first
+        // byte should be 2 if y is even, and 3 if y is odd.
```
samuel-lee-msft, line 1495: >// byte should be 2 if y is even, and 3 if y is odd. [](http://example.com/codeflow?start=8&length=52)

Could you add a comment here saying there are 3 conversion formats, for uncompressed we do not care about the oddness of Y, for HYBRID / COMPRESSED we do. For COMPRESSED we also do not encode Y.

When I first read this code I thought we only supported UNCOMPRESSED and COMPRESSED but not HYBRID.

samuel-lee-msft (commented):
LGTM once comments are addressed.

SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
mamckee: Also, under the hood these are stored as the OSSL_PARAM_UNSIGNED_INTEGER parameter type, not OSSL_PARAM_OCTET_STRING. They're actually documented as such here: https://www.openssl.org/docs/man3.3/man7/EVP_PKEY-RSA.html. The parameter fetching functions do check the type so using OSSL_PARAM_set_octet_string would cause these to be rejected. I'm not sure why there's not just an explicit BN parameter type instead.

I'll make the change to LE though.

mamckee: I can also make the change in scossl_rsa_export_key

CMakeLists.txt
mamckee: Published but didn't update this file🤦I should look into a git hook for automating this on tag.

ScosslCommon/src/scossl_aes_aead.c
mamckee: It's only used in the engine and in this case yes (it knows the value because we called EVP_CIPHER_meth_set_iv_length). If the IV length is different, then that should be updated after EVP_CTRL_INIT.

SymCryptProvider/src/p_scossl_base.c
mamckee: No, these are passed by libcrypto, not the caller. I think the FIPS provider needs these to avoid escaping the FIPS boundary. Since that boundary is SymCrypt for us, we don't really need these.

Early on, these were the only functions the provider really needed from libcrypto, so this helped us avoid a dependency on libcrypto. Now though, there's a number of upcalls we need to make (e.g. for BIGNUM or ASN.1) that aren't found in this table.

SymCryptProvider/src/asymcipher/p_scossl_rsa_cipher.c
mamckee: If RSA_PKCS1_WITH_TLS_PADDING is used, then failures should return success with a random secret. RSA_PKCS1_PADDING should fail, but doesn't require the message is set to something random.

In both cases our provider does set the message to something random, but the expected return is different. I think the intent was to make it easier for callers to use PKCS1.5 safely (i.e not relying on the error code)

https://www.openssl.org/docs/man3.1/man3/EVP_PKEY_CTX_set_rsa_padding.html

SymCryptProvider/src/mac/p_scossl_hmac.c
mamckee: Nope

SymCryptProvider/src/signature/p_scossl_rsa_signature.c
mamckee: Yes, I should probably just put this in a helper function. It's duplicated a few places for PSS. This is one of the pieces carried over from the engine.

mamckee:
> > ```
> >             !scossl_ec_point_to_pubkey(ecPoint, ecGroup, bnCtx, pbPublicKey, cbPublicKey))
> > ```
> 
> It's a little odd that we always use OpenSSL functions to convert compressed points into uncompressed points to ingest into SymCrypt, but roll our own logic to convert from uncompressed points to compressed points.
> 
> Should we verify that the encoded point provided here matches keyCtx->conversionFormat on ingesting a public key blob?
> 
> Refers to: SymCryptProvider/src/keymgmt/p_scossl_ecc_keymgmt.c:668 in [49d8b21](https://github.com/microsoft/SymCrypt-OpenSSL/commit/49d8b21c12c37ee0c8d94fbbc0f90c04259c2975). [](commit_id = [49d8b21](https://github.com/microsoft/SymCrypt-OpenSSL/commit/49d8b21c12c37ee0c8d94fbbc0f90c04259c2975), deletion_comment = False)

The conversion format is an export/get only parameter so callers can decide which format to use instead of us always exporting uncompressed. For import we can't rely on this being set and should determine the format from the first byte. We're using the OpenSSL logic since that validates the encoding and group compatibility. On export we only really need to determine the first byte and whether to omit Y.

The default implementation also doesn't check that the compression format matches what's set in import so I don't think it would add anything for us to.

CMakeLists.txt
mlindgren: In SymCrypt, we don't use a Git hook, but we have a single file where the version is specified, and then we use a script to generate the correct output for all other places where it needs to be embedded, including CMakeLists.txt. See `scripts\version.py` in the SymCrypt ADO repo (hasn't been pushed to public GitHub yet, but should be soon.)

mlindgren (approved):
lgtm :shipit:

PR #82, opened 2024-07-25
Add curly braces around case block 

justanotherminh:
I'm adding curly braces around case block to ensure compatibility with the version of clang used for OE CI. Currently the lack of curly braces is causing clang to run into syntax errors due the declaration `SIZE_T outlPadded = *outl;` under `case TLS1_VERSION:`


mamckee:
Addressed in #83

PR #84, opened 2024-08-07
Fixes for tpm2

mamckee:
Testing with tpm2 and SymCrypt-OpenSSL on Azure Linux 3 uncovered a compatibility issue and a bug. This PR addresses both.

- Add EC key X and Y coordinates to gettable EC key parameters
- Add missing AES parameters. These are always 0 but may be requested by some callers
- Make AES-CFB compatible with OpenSSL's stream cipher calling pattern.
  - AES-CFB is expected to encrypt the entirety of inl on each cipher update call. The cipher final call should do nothing.

### DIFF
--- a/SymCryptProvider/src/ciphers/p_scossl_aes.c
+++ b/SymCryptProvider/src/ciphers/p_scossl_aes.c
@@ -22,6 +22,9 @@ static const OSSL_PARAM p_scossl_aes_generic_param_types[] = {
     OSSL_PARAM_size_t(OSSL_CIPHER_PARAM_BLOCK_SIZE, NULL),
     OSSL_PARAM_int(OSSL_CIPHER_PARAM_AEAD, NULL),
     OSSL_PARAM_int(OSSL_CIPHER_PARAM_CUSTOM_IV, NULL),
+    OSSL_PARAM_int(OSSL_CIPHER_PARAM_CTS, NULL),
+    OSSL_PARAM_int(OSSL_CIPHER_PARAM_TLS1_MULTIBLOCK, NULL),
+    OSSL_PARAM_int(OSSL_CIPHER_PARAM_HAS_RAND_KEY, NULL),
     OSSL_PARAM_END};
 
 static const OSSL_PARAM p_scossl_aes_generic_gettable_ctx_param_types[] = {
@@ -44,6 +47,9 @@ typedef struct
 
     BYTE iv[SYMCRYPT_AES_BLOCK_SIZE];
     BYTE pbChainingValue[SYMCRYPT_AES_BLOCK_SIZE];
+    // Needed for AES-CFB when input size in update
+    // is not a multiple of block size
+    BYTE pbChainingValueLast[SYMCRYPT_AES_BLOCK_SIZE];
     BOOL encrypt;
     BOOL pad;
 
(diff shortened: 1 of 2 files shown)


SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -742,7 +813,7 @@ static SCOSSL_STATUS scossl_aes_ecb_cipher(_Inout_ SCOSSL_AES_CTX *ctx,
     return SCOSSL_SUCCESS;
 }
 
-static SCOSSL_STATUS p_scossl_aes_cfb_cipher_internal(_Inout_ SCOSSL_AES_CTX *ctx, SIZE_T cbShift,
+static SCOSSL_STATUS p_scossl_aes_cfb_cipher_internal(_Inout_ SCOSSL_AES_CTX *ctx, SIZE_T cbShift, PBYTE pbChainingValue,
```
samuel-lee-msft, line 816: Ideally add 
`_Inout_updates_(SYMCRYPT_AES_BLOCKSIZE)`?

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -742,7 +813,7 @@ static SCOSSL_STATUS scossl_aes_ecb_cipher(_Inout_ SCOSSL_AES_CTX *ctx,
...
                                                       _Out_writes_bytes_(*outl) unsigned char *out, _Out_ size_t *outl, size_t outsize,
```
samuel-lee-msft, line 817: >_Out_ [](http://example.com/codeflow?start=100&length=5)

`_Out_opt_`?

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -759,31 +830,135 @@ static SCOSSL_STATUS p_scossl_aes_cfb_cipher_internal(_Inout_ SCOSSL_AES_CTX *ct
...
+    }
+
+    cbInFullBlocks = inl & ~(SYMCRYPT_AES_BLOCK_SIZE-1);
+    cbInRemaining = inl - cbInFullBlocks;
+
+    if (cbInFullBlocks > 0)
+    {
+        if (ctx->cbBuf != 0)
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_CIPHER_OPERATION_FAILED);
```
samuel-lee-msft, line 910: >PROV_R_CIPHER_OPERATION_FAILED [](http://example.com/codeflow?start=36&length=30)

Maybe `ERR_R_INTERNAL_ERROR` is more appropriate given this condition should never occur unless a logic issue is introduced somewhere or memory corruption occurs.

mamckee: Yeah that's a fair point. It wouldn't be a cipher error strictly speaking.

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -759,31 +830,135 @@ static SCOSSL_STATUS p_scossl_aes_cfb_cipher_internal(_Inout_ SCOSSL_AES_CTX *ct
...
+        // Last update call was a partial block. Fill buffer and perform cipher
+        // with previous chaining value before continuing.
+        cbBufRemaining = SYMCRYPT_MIN(SYMCRYPT_AES_BLOCK_SIZE - ctx->cbBuf, inl);
+
+        memcpy(ctx->buf + ctx->cbBuf, in, cbBufRemaining);
+
+        if (p_scossl_aes_cfb_cipher_internal(
+                ctx,
+                SYMCRYPT_AES_BLOCK_SIZE,
+                ctx->pbChainingValue,
```
samuel-lee-msft, line 878: >ctx->pbChainingValue, [](http://example.com/codeflow?start=16&length=21)

Hmm, don't you need to copy to `pbChainingValueLast` before this too if we are called with multiple shorter-than-block input buffers?

i.e. if we are here with ctx->cbBuf == 1 and inl==1, we will need to restore ctx->pbChainingValue after this call.

mamckee: Yeah, I think I talked myself out of this edge case but we'd hit this if a caller passed some inl < cbShift twice in a row.

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -759,31 +831,142 @@ static SCOSSL_STATUS p_scossl_aes_cfb_cipher_internal(_Inout_ SCOSSL_AES_CTX *ct
...
+        *outl = inl;
+    }
+
+    if (ctx->cbBuf > 0)
+    {
+        // Last update call was a partial block. Fill buffer and perform cipher
+        // with previous chaining value before continuing.
+        cbBufRemaining = SYMCRYPT_MIN(SYMCRYPT_AES_BLOCK_SIZE - ctx->cbBuf, inl);
+
+        // // Save the chaining value for the next call in case inl < cbBufRemaining
```
samuel-lee-msft, line 874: >inl < cbBufRemaining [](http://example.com/codeflow?start=64&length=20)

NP: `ctx->cbBuf + inl < SYMCRYPT_AES_BLOCK_SIZE`

samuel-lee-msft (approved):
NP about comment, but LGTM

samuel-lee-msft (approved):

PR #86, opened 2024-08-19
Behavior fixes for API compatability

mamckee:
This PR addresses behavior fixes for compatibility issues found when calling OpenSSL with the SymCryp provider though different wrapper libraries. These are behavior differences to make the SymCrypt provider act more like the default provider.
- Allow larger/smaller output sizes for ECDH secret. If the output size is smaller than the output of `SymCryptEcDhSecretAgreement`, the secret is truncated.
- Allow larger output sizes for DH secret.
- When `inl` passed to AES cipher update is zero, `*outl` should always be set to zero.
- Always return a sufficiently large buffer size for ECDSA digest sign

### DIFF
--- a/SymCryptProvider/src/ciphers/p_scossl_aes.c
+++ b/SymCryptProvider/src/ciphers/p_scossl_aes.c
@@ -248,6 +248,7 @@ static SCOSSL_STATUS p_scossl_aes_generic_block_update(_Inout_ SCOSSL_AES_CTX *c
 
     if (inl == 0)
     {
+        *outl = 0;
         return SCOSSL_SUCCESS;
     }
 
@@ -428,6 +429,7 @@ static SCOSSL_STATUS p_scossl_aes_generic_stream_update(_Inout_ SCOSSL_AES_CTX *
 {
     if (inl == 0)
     {
+        *outl = 0;
         return SCOSSL_SUCCESS;
     }
 
@@ -769,10 +771,7 @@ static SCOSSL_STATUS scossl_aes_cbc_cipher(_Inout_ SCOSSL_AES_CTX *ctx,
         return SCOSSL_FAILURE;
     }
 
-    if (outl != NULL)
-    {
-        *outl = inl;
-    }
+    *outl = inl;
 
     if (ctx->encrypt)
     {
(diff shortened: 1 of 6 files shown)


SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -796,10 +795,7 @@ static SCOSSL_STATUS scossl_aes_ecb_cipher(_Inout_ SCOSSL_AES_CTX *ctx,
         return SCOSSL_FAILURE;
     }
 
-    if (outl != NULL)
-    {
-        *outl = inl;
-    }
+    *outl = inl;
```
ccalik, line 798: Why drop the null-pointer check? SAL annotation hints that `outl` can be null.

mamckee: I should update the annotation. OpenSSL guarantees this will be non-null so it's unnecessary. The default provider doesn't make this check either.

mamckee: Got overzealous, this one should still have the check

vcsjones: Agree, without the null checks we get crashes in .NET. I think this should just be changes in `p_scossl_aes_generic_stream_update` and `p_scossl_aes_generic_block_update`, and put the null checks back.

mamckee: SAL annotations were incorrect for `outl`. They should have been `_Out_opt_` instead of `_Out_`

mlindgren (approved):
Just confirming: the expected behavior for OpenSSL is that the ECDH secret can be truncated if the output length is too short, but the DH secret cannot, and returns an error in that case? That seems very strange. But assuming that's the desired behavior, the changes look good to me.

mamckee:
> Just confirming: the expected behavior for OpenSSL is that the ECDH secret can be truncated if the output length is too short, but the DH secret cannot, and returns an error in that case? That seems very strange. But assuming that's the desired behavior, the changes look good to me.

Yeah, I think it's a little confusing they don't leave this up to the caller to truncate and keep interfaces consistent.
https://github.com/openssl/openssl/blob/2e966354956e55825331f61bf98976ffe60e4332/providers/implementations/exchange/ecdh_exch.c#L457-L460

PR #88, opened 2024-08-27
Remove GCM nonce restriction comment

vcsjones:
In the README, it is mentioned that GCM IVs are restricted to 96 bits. It looks like that restriction was removed as part of https://github.com/microsoft/SymCrypt-OpenSSL/pull/80/.


mamckee (approved):
This readme is outdated in general. I'll be making another pass later this week. Thanks for the PR.

PR #90, opened 2024-09-06
Add SHAKE and CSHAKE

mamckee:
This PR adds SHAKE and CSHAKE XOFs to the provider digest interface, and improves the maintainability and testing of digests.

- Refactored existing digests to use `SymCryptHash` functions instead of relying on large macro expansions
- Add support for querying `OSSL_DIGEST_PARAM_XOF` and `OSSL_DIGEST_PARAM_ALGID_ABSENT` parameters, which are already supported by the OpenSSL default provider
- Add digest test recipes to EvpRecipes
- Add SHAKE128/256
- Add CSHAKE128/256
  - CSHAKE is not currently supported by the OpenSSL default provider. If support is added, then the parameter names for the function name string and customization string may need to be updated to match.
  - Add custom parameters needed for using CSHAKE with the SymCrypt provider to `symcrypt_provider.h`


SymCryptProvider/src/digests/p_scossl_cshake.c
```diff
@@ -0,0 +1,416 @@
...
+    const SCOSSL_CSHAKE_HASH *pHash;
+    SCOSSL_CSHAKE_STATE *pState;
+    BOOL updating;
+
+    PBYTE pbFunctionNameString;
+    SIZE_T cbFunctionNameString;
+    PBYTE pbCustomizationString;
+    SIZE_T cbCustomizationString;
+
+    SIZE_T xofLen;
```
ccalik, line 73: Why do we need this, is that the default output size?

SymCryptProvider/src/digests/p_scossl_shake.c
```diff
@@ -0,0 +1,103 @@
...
+    ctx->xofLen = SymCryptHashResultSize(ctx->pHash);
+
+    return p_scossl_shake_set_ctx_params(ctx, params);
+}
+
+static SCOSSL_STATUS p_scossl_shake_extract(_Inout_ SCOSSL_DIGEST_CTX *ctx,
+                                            PSYMCRYPT_HASH_EXTRACT extractFunc, BOOLEAN wipeState,
+                                            _Out_writes_bytes_(*outl) unsigned char *out, _Out_ size_t *outl, size_t outlen)
+{
+    if (outlen < ctx->xofLen)
```
ccalik, line 49: Why are we generating outputs of exactly ctx->xofLen bytes? A XOF can generate any number of requested bytes.

SymCryptProvider/src/digests/p_scossl_cshake.c
mamckee: The requested output bytes for is passed as a parameter (`OSSL_DIGEST_PARAM_XOFLEN`) and stored in `ctx->xofLen`. We pass that later as the output length of the SHAKE operation.

SymCryptProvider/src/digests/p_scossl_shake.c
mamckee: See above for the use of `ctx->xofLen`. `outlen` just indicates the size of the buffer `out`. We need to make sure the output buffer is sufficiently large.

SymCryptProvider/src/digests/p_scossl_cshake.c
ccalik: Maybe it is the OpenSSL way of doing it, technically we should be able to generate whatever the number of output bytes is requested.

SymCryptProvider/src/digests/p_scossl_shake.c
ccalik: This comment is more for Line 55 below actually. We output `ctx->xofLen` bytes even when `outlen` is larger.

SymCryptProvider/src/digests/p_scossl_cshake.c
```diff
@@ -0,0 +1,416 @@
...
+    (PSYMCRYPT_HASH_APPEND_FUNC)SymCryptCShake256Append,
+    (PSYMCRYPT_HASH_EXTRACT)SymCryptCShake256Extract,
+    (PSYMCRYPT_HASH_STATE_COPY_FUNC)SymCryptCShake256StateCopy,
+    SYMCRYPT_CSHAKE256_INPUT_BLOCK_SIZE,
+    SYMCRYPT_CSHAKE256_RESULT_SIZE};
+
+typedef struct
+{
+    const SCOSSL_CSHAKE_HASH *pHash;
+    SCOSSL_CSHAKE_STATE *pState;
```
ccalik, line 65: Does the state have to be managed dynamically?

SymCryptProvider/src/digests/p_scossl_shake.c
mamckee: I see. `outlen` can be bigger than the requested output length, but we still should only fill the buffer to `ctx->xoflen`. This is what the default provider does as well.

SymCryptProvider/src/digests/p_scossl_cshake.c
mamckee: We do generate the requested number of bytes. The caller can either set it by calling `EVP_DigestInit_ex2` or `EVP_MD_CTX_set_params`

SymCryptProvider/src/digests/p_scossl_cshake.c
mamckee: No, I suppose not. We'd need to allocate the whole struct with proper alignment but that should be fine.

SymCryptProvider/inc/scossl_provider.h
```diff
@@ -0,0 +1,23 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+//
+// Digest parameters
+//
+
+#define SCOSSL_DIGEST_PARAM_FUNCTION_NAME "function-name"
```
ccalik, line 18: NP: `SCOSSL_DIGEST_PARAM_FUNCTION_NAME_STRING" for consistency. Is the string defined by OpenSSL?

mamckee: No, OpenSSL only supports SHAKE. It makes the parameter a bit wordy but I suppose that's fine since it's what's defined in the spec.

EvpTestRecipes/3.0/evpmd_sha.txt
```diff
@@ -0,0 +1,208 @@
...
+Output = b9b92544fb25cfe4ec6fe437d8da2bbe00f7bdaface3de97b8775a44d753c3adca3f7c6f183cc8647e229070439aa9539ae1f8f13470c9d3527fffdeef6c94f9f0520ff0c1ba8b16e16014e1af43ac6d94cb7929188cce9d7b02f81a2746f52ba16988e5f6d93298d778dfe05ea0ef256ae3728643ce3e29c794a0370e9ca6a8bf3e7a41e86770676ac106f7ae79e67027ce7b7b38efe27d253a52b5cb54d6eb4367a87736ed48cb45ef27f42683da140ed3295dfc575d3ea38cfc2a3697cc92864305407369b4abac054e497378dd9fd0c4b352ea3185ce1178b3dc1599df69db29259d4735320c8e7d33e8226620c9a1d22761f1d35bdff79a
+
+Digest = SHAKE256
+Input = 8d8001e2c096f1b88e7c9224a086efd4797fbf74a8033a2d422a2b6b8f6747e4
+Output = 2e975f6a8a14f0704d51b13667d8195c219f71e6345696c49fa4b9d08e9225d3d39393425152c97e71dd24601c11abcfa0f12f53c680bd3ae757b8134a9c10d429615869217fdd5885c4db174985703a6d6de94a667eac3023443a8337ae1bc601b76d7d38ec3c34463105f0d3949d78e562a039e4469548b609395de5a4fd43c46ca9fd6ee29ada5efc07d84d553249450dab4a49c483ded250c9338f85cd937ae66bb436f3b4026e859fda1ca571432f3bfc09e7c03ca4d183b741111ca0483d0edabc03feb23b17ee48e844ba2408d9dcfd0139d2e8c7310125aee801c61ab7900d1efc47c078281766f361c5e6111346235e1dc38325666c
+
+
+Digest = CSHAKE128
```
ccalik, line 191: When the string parameters are empty, it is actually the SHAKE output rather than CSHAKE. Not sure this file can handle them, though.

mamckee: Yep, the evp_test binary doesn't support custom parameters for digests. I wrote a quick test application and verified with the customization and function name strings. I'll probably need to update SslPlay to accommodate the test gap, but I'll open a separate PR for that. We'll need that change for testing KEMs as well.

SymCryptProvider/src/digests/p_scossl_cshake.c
```diff
@@ -0,0 +1,391 @@
...
+
+    if (ctx != NULL)
+    {
+        if (ctx->pbFunctionNameString != NULL)
+        {
+            copyCtx->pbFunctionNameString = OPENSSL_memdup(ctx->pbFunctionNameString, ctx->cbFunctionNameString);
+            if (copyCtx->pbFunctionNameString == NULL)
+            {
+                ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+                goto cleanup;
```
ccalik, line 135: I don't see that `copyCtx` is initialized to zero. If `OPENSSL_memdup` fails for `copyCtx->pbFunctionNameString` then `p_scossl_cshake_freectx` will call `OPENSSL_free(ctx->pbCustomizationString);`

SymCryptProvider/src/digests/p_scossl_cshake.c
```diff
@@ -0,0 +1,391 @@
...
+{
+    if (outlen < ctx->xofLen)
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_OUTPUT_BUFFER_TOO_SMALL);
+        return SCOSSL_FAILURE;
+    }
+
+    ctx->pHash->extractFunc(&ctx->state, out, ctx->xofLen, wipeState);
+    *outl = ctx->xofLen;
+
```
ccalik, line 224: Don't you need to set `ctx->updating` to `FALSE` if `wipeState = TRUE`?

SymCryptProvider/src/digests/p_scossl_cshake.c
```diff
@@ -0,0 +1,391 @@
...
+    return SCOSSL_SUCCESS;
+}
+
+static SCOSSL_STATUS p_scossl_cshake_final(_In_ SCOSSL_CSHAKE_CTX *ctx,
+                                           _Out_writes_bytes_(*outl) unsigned char *out, _Out_ size_t *outl, size_t outlen)
+{
+    return p_scossl_cshake_extract(ctx, TRUE, out, outl, outlen);
```
ccalik, line 231: What is the expected behavior when a CSHAKE computation is finalized; do you want to start the new computation with the previous strings, or do they have to be set again?

SymCryptProvider/src/digests/p_scossl_cshake.c
mamckee: No, this ensures the caller must call init again before setting any parameters or trying to call update again. The variable name is a little misleading though. `initialized` would probably be more appropriate.

SymCryptProvider/src/digests/p_scossl_cshake.c
mamckee: If the caller wants to re-use the context, they need to call init again. Looking back at the OpenSSL code though, I don't think we should actually wipe the strings on init.

SymCryptProvider/src/digests/p_scossl_cshake.c
ccalik: I don't quite understand why we set `ctx->updating = TRUE` when `wipeState = TRUE`. When `wipeState = TRUE`, `extractFunc` wipes the state and if we set `ctx->updating = TRUE` then the next call to `cshake_update` will append on an empty state.

SymCryptProvider/src/digests/p_scossl_cshake.c
```diff
@@ -0,0 +1,384 @@
...
+static SCOSSL_STATUS p_scossl_cshake_extract(_In_ SCOSSL_CSHAKE_CTX *ctx, BOOLEAN wipeState,
+                                             _Out_writes_bytes_(*outl) unsigned char *out, _Out_ size_t *outl, size_t outlen)
+{
+    if (outlen < ctx->xofLen)
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_OUTPUT_BUFFER_TOO_SMALL);
+        return SCOSSL_FAILURE;
+    }
+
+    ctx->pHash->extractFunc(&ctx->state, out, ctx->xofLen, wipeState);
```
ccalik, line 213: If `p_scossl_cshake_extract` is called without an `p_scossl_cshake_update` call then initialization is not performed and the output from `ctx->pHash->extractFunc` will not be correct?

SymCryptProvider/src/digests/p_scossl_cshake.c
ccalik: What is the desired state after `p_scossl_cshake_final` is called? Is it defaulting to SHAKE when `p_scossl_cshake_update` is called after?

mamckee: No, the old customization and function name strings are still present if the context is re-used. We could enforce that the caller has to call init agian, but all that does is set the parameters and defer the actual init call.

ccalik: If we set `ctx->updating = FALSE` unconditionally here, then the next update function would be able to reinitialize the state with the strings.

ccalik (approved):

PR #91, opened 2024-09-09
Add SRTP KDF

mamckee:
This PR adds SRTP and SRTCP KDF to the SymCrypt provider. OpenSSL itself does not implement SRTP, but can be used as part of an implementation: https://docs.openssl.org/3.3/man3/SSL_CTX_set_tlsext_use_srtp.

- Add SRTP and SRTCP KDF
    - Add custom parameters needed for using these KDFs with the SymCrypt provider to `symcrypt_provider.h`
    - Add test cases

### DIFF
--- a/SymCryptProvider/src/kdf/p_scossl_srtpkdf.c
+++ b/SymCryptProvider/src/kdf/p_scossl_srtpkdf.c
@@ -23,6 +23,7 @@ extern "C" {
 
 typedef struct
 {
+    BOOL isSrtcp;
     // pKey is immediately expanded into expandedKey. It is only kept
     // in the context for duplication and initialization checks.
     PBYTE pKey;
@@ -56,6 +57,18 @@ static SCOSSL_STATUS p_scossl_srtpkdf_set_ctx_params(_Inout_ SCOSSL_PROV_SRTPKDF
 static SCOSSL_PROV_SRTPKDF_CTX *p_scossl_srtpkdf_newctx(ossl_unused void *provctx)
 {
     return OPENSSL_zalloc(sizeof(SCOSSL_PROV_SRTPKDF_CTX));
+}
+
+static SCOSSL_PROV_SRTPKDF_CTX *p_scossl_srtcpkdf_newctx(ossl_unused void *provctx)
+{
+    SCOSSL_PROV_SRTPKDF_CTX *ctx = OPENSSL_zalloc(sizeof(SCOSSL_PROV_SRTPKDF_CTX));
+
+    if (ctx != NULL)
+    {
+        ctx->isSrtcp = TRUE;
+    }
+
+    return ctx;
 }
 
 static void p_scossl_srtpkdf_freectx(_Inout_ SCOSSL_PROV_SRTPKDF_CTX *ctx)
@@ -103,6 +116,7 @@ static SCOSSL_PROV_SRTPKDF_CTX *p_scossl_srtpkdf_dupctx(_In_ SCOSSL_PROV_SRTPKDF
             memcpy(copyCtx->pbSalt, ctx->pbSalt, SCOSSL_SRTP_KDF_SALT_SIZE);
         }
 
+        copyCtx->isSrtcp = ctx->isSrtcp;
         copyCtx->isSaltSet = ctx->isSaltSet;
         copyCtx->uKeyDerivationRate = ctx->uKeyDerivationRate;
         copyCtx->uIndex = ctx->uIndex;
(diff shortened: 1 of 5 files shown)


SymCryptProvider/src/kdf/p_scossl_srtpkdf.c
```diff
@@ -0,0 +1,390 @@
...
+} SCOSSL_PROV_SRTPKDF_CTX;
+
+static const OSSL_PARAM p_scossl_srtpkdf_gettable_ctx_param_types[] = {
+    OSSL_PARAM_size_t(OSSL_KDF_PARAM_SIZE, NULL),
+    OSSL_PARAM_END};
+
+static const OSSL_PARAM p_scossl_srtpkdf_settable_ctx_param_types[] = {
+    OSSL_PARAM_octet_string(OSSL_KDF_PARAM_KEY, NULL, 0),
+    OSSL_PARAM_octet_string(OSSL_KDF_PARAM_SALT, NULL, 0),
+    OSSL_PARAM_utf8_string(OSSL_KDF_PARAM_LABEL, NULL, 0),
```
ccalik, line 44: Label is a BYTE, wouldn't using OSSL_PARAM_UNSIGNED_INTEGER be better?

ccalik: I've noticed that labels are provided as strings.

SymCryptProvider/src/kdf/p_scossl_srtpkdf.c
```diff
@@ -0,0 +1,390 @@
...
+    OPENSSL_secure_clear_free(ctx->pbKey, ctx->cbKey);
+
+    ctx->pbKey = NULL;
+    ctx->cbKey = 0;
+    ctx->isSaltSet = FALSE;
+    ctx->uKeyDerivationRate = 0;
+    ctx->uIndex = 0;
+    ctx->uIndexWidth = 0;
+    ctx->label = 0;
+
```
ccalik, line 146: Wipe pbSalt?

SymCryptProvider/src/kdf/p_scossl_srtpkdf.c
```diff
@@ -0,0 +1,390 @@
...
+    if ((p = OSSL_PARAM_locate_const(params, SCOSSL_KDF_PARAM_SRTP_INDEX_WIDTH)) != NULL)
+    {
+        if (!OSSL_PARAM_get_uint32(p, &ctx->uIndexWidth))
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_GET_PARAMETER);
+            return SCOSSL_FAILURE;
+        }
+
+        if (ctx->uIndexWidth > 64)
```
ccalik, line 338: SymCrypt only allows index widths 0, 32, or 48.

SymCryptProvider/src/kdf/p_scossl_srtpkdf.c
mamckee: RFC 3711 says the salt can be public: https://datatracker.ietf.org/doc/html/rfc3711#section-4.3.1

Is there a FIPS requirement to keep the salt private?

ccalik: I don't think so. I was thinking it would be more consistent with the newctx function to reset everything. On way todo is to wipe ctx after freeing ctx->pbKey.

SymCryptProvider/src/kdf/p_scossl_srtpkdf.c
```diff
@@ -0,0 +1,390 @@
...
+        ERR_raise(ERR_LIB_PROV, PROV_R_MISSING_SALT);
+        return SCOSSL_FAILURE;
+    }
+
+    scError = SymCryptSrtpKdfDerive(
+        &ctx->expandedKey,
+        ctx->pbSalt, SCOSSL_SRTP_KDF_SALT_SIZE,
+        ctx->uKeyDerivationRate,
+        ctx->uIndex, ctx->uIndexWidth,
+        ctx->label,
```
ccalik, line 178: SymCrypSrtpKdfDerive will accept any label value, it may be uninitialized here.

ccalik (approved):

SymCryptProvider/src/kdf/p_scossl_srtpkdf.c
mamckee: I see. If `ctx->isSaltSet` is cleared though the caller will need to set a new salt, so having a zeroed salt wouldn't make a difference here.

ccalik: I guess the question is what the intended behavior of the reset function is and what part of the state needs to be preserved. Zero is a valid value for `uKeyDerivationRate`, `uIndex` and `uIndexWidth` but empty salt is not.

ccalik: But you're right, the code is functionally equivalent to wiping the state now.

SymCryptProvider/src/kdf/p_scossl_srtpkdf.c
mamckee: Right, which would mean we default to SYMCRYPT_SRTP_ENCRYPTION_KEY. I guess that would be fine for SRTP but an error case for SRTCP so this should have an 'unset' value that we check.

SymCryptProvider/src/kdf/p_scossl_srtpkdf.c
```diff
@@ -0,0 +1,409 @@
...
+        !OSSL_PARAM_set_size_t(p, SIZE_MAX))
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_SET_PARAMETER);
+        return SCOSSL_FAILURE;
+    }
+
+    return SCOSSL_SUCCESS;
+}
+
+static SCOSSL_STATUS p_scossl_srtpkdf_set_ctx_params(_Inout_ SCOSSL_PROV_SRTPKDF_CTX *ctx, _In_ const OSSL_PARAM params[])
```
ccalik, line 228: When one of the key, salt, or label parameters is invalid, the state is not updated. However, for other parameters the state could updated with an invalid value even though the function returns an error. Is this behavior OK?

mamckee: Yes, this is consistent with other provider interfaces. If a `XXX_set_ctx_params` function fails, the caller can't assume that the context is valid.

ccalik (approved):

PR #93, opened 2024-09-18
Add SSKDF (SP800-56C)

mamckee:
This PR adds SSKDF as specified in SP800-56C, and associated tests. This bumps the dependency on SymCrypt to 103.5.0, which introduces the KDF.

### DIFF
--- a/SymCryptProvider/src/kdf/p_scossl_sskdf.c
+++ b/SymCryptProvider/src/kdf/p_scossl_sskdf.c
@@ -214,7 +214,7 @@ SCOSSL_STATUS p_scossl_sskdf_derive(_In_ SCOSSL_PROV_SSKDF_CTX *ctx,
     else if (ctx->pHash != NULL)
     {
         scError = SymCryptSskdfHash(
-            ctx->pHash,
+            ctx->pHash, 0,
             ctx->pbSecret, ctx->cbSecret,
             ctx->pbInfo, ctx->cbInfo,
             key, keylen);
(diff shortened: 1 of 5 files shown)


SymCryptProvider/src/kdf/p_scossl_sskdf.c
```diff
@@ -0,0 +1,456 @@
...
+        if ((md = EVP_MD_fetch(ctx->libCtx, mdName, propq)) == NULL ||
+            (ctx->mdnid = EVP_MD_type(md)) == NID_undef ||
+            (ctx->pHash = scossl_get_symcrypt_hash_algorithm(ctx->mdnid)) == NULL)
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_DIGEST);
+            goto cleanup;
+        }
+    }
+
+    if ((p = OSSL_PARAM_locate_const(params, OSSL_KDF_PARAM_MAC)) != NULL)
```
ccalik, line 405: Is setting a MAC and a hash at the same time an error condition? Looks like the hash is ignored in derivation in that case.

mamckee: No, if the MAC is HMAC then the digest is used. If the MAC is KMAC then the hash is ignored, but this is the same behavior as the OpenSSL implementation.

ccalik: I see, it is `ctx->mdnid` that is used to get the HMAC algorithm rather than `ctx->pHash` in that case, that's why I missed it.

ccalik (approved):

ccalik (approved):

PR #94, opened 2024-09-27
cmake: fixup CMakeLists to avoid installing libraries in incorrect path

ahcbb6:
usr/CACHE;PATH;Path to OpenSSL providers;lib/symcryptprovider.so
vs
usr/lib/ossl-modules/symcryptprovider.so

Signed-off-by: Alejandro Hernandez Samaniego <alhe@linux.microsoft.com>

### DIFF
--- a/SymCryptProvider/CMakeLists.txt
+++ b/SymCryptProvider/CMakeLists.txt
@@ -65,7 +65,9 @@ target_link_libraries(scossl_provider PRIVATE scossl_common)
 target_link_libraries(scossl_provider PUBLIC ${SYMCRYPT_LIBRARY})
 target_link_libraries(scossl_provider PUBLIC ${OPENSSL_CRYPTO_LIBRARY})
 
-set(OPENSSL_PROVIDERS CACHE PATH "Path to OpenSSL providers" "${CMAKE_INSTALL_LIBDIR}/ossl-modules")
+if (NOT DEFINED OPENSSL_PROVIDERS)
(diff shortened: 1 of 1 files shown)


SymCryptProvider/CMakeLists.txt
```diff
@@ -65,12 +65,14 @@ target_link_libraries(scossl_provider PRIVATE scossl_common)
 target_link_libraries(scossl_provider PUBLIC ${SYMCRYPT_LIBRARY})
 target_link_libraries(scossl_provider PUBLIC ${OPENSSL_CRYPTO_LIBRARY})
 
-set(OPENSSL_PROVIDERS CACHE PATH "Path to OpenSSL providers" "${CMAKE_INSTALL_LIBDIR}/ossl-modules")
+if (NOT DEFINED OPENSSL_PROVIDERS)
+    set(OPENSSL_ENGINES "${CMAKE_INSTALL_LIBDIR}/ossl-modules" CACHE PATH "Path to OpenSSL providers")
```
mlindgren, line 69: Looks like we're now defining `OPENSSL_ENGINES` instead of `OPENSSL_PROVIDERS`. Do we need to define both?

ahcbb6: Indeed it should, fixing it right away.

ahcbb6: Done

mlindgren (approved):
Looks good to me, but I'll wait for @mamckee to sign off.

PR #97, opened 2024-11-14
Add missing digests

mamckee:
This PR adds the following digests to the SymCrypt provider and tests for interoperability. These algorithms will be exposed in an upcoming SymCrypt release.

- SHA224
- SHA512/224
- SHA512/256
- SHA3-224


ScosslCommon/src/scossl_helpers.c
```diff
@@ -442,12 +454,20 @@ PCSYMCRYPT_HASH scossl_get_symcrypt_hash_algorithm(int mdnid)
...
+    case NID_sha224:
+        return SymCryptSha224Algorithm;
     case NID_sha256:
         return SymCryptSha256Algorithm;
     case NID_sha384:
         return SymCryptSha384Algorithm;
     case NID_sha512:
         return SymCryptSha512Algorithm;
+    case NID_sha512_224:
+        return SymCryptSha512Algorithm;
```
samuel-lee-msft, line 466: >SymCryptSha512Algorithm [](http://example.com/codeflow?start=15&length=23)

SymCryptSha512_256Algorithm

ScosslCommon/src/scossl_helpers.c
```diff
@@ -442,12 +454,20 @@ PCSYMCRYPT_HASH scossl_get_symcrypt_hash_algorithm(int mdnid)
...
+    case NID_sha512_256:
+        return SymCryptSha512Algorithm;
```
samuel-lee-msft, line 468: >SymCryptSha512Algorithm [](http://example.com/codeflow?start=15&length=23)

SymCryptSha512_256Algorithm

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
```diff
@@ -439,6 +442,9 @@ static SCOSSL_STATUS p_scossl_ecdsa_get_ctx_params(_In_ SCOSSL_ECDSA_CTX *ctx, _
         case NID_sha512:
```
samuel-lee-msft, line 442: Any reason to skip truncated SHA512 here?

samuel-lee-msft (approved):
LGTM modulo comments

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
mamckee: There's no OID for ECDSA with either. The OIDs for ECDSA + SHA2 family are taken from here: https://www.rfc-editor.org/rfc/rfc5758.html.

OpenSSL will also fail in this case.

PR #98, opened 2024-11-20
Additional provider RSA compatibility

mamckee:
This PR updates the provider to support additional scenarios for RSA
- Allow RSA private key import with private exponent and no prime factors
- Allow RSA PKCS1 with no digest
- Fix `libctx` fetch in provider init to match recommendation from OpenSSL documentation


SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
```diff
@@ -1160,6 +1162,27 @@ static SCOSSL_STATUS p_scossl_rsa_keymgmt_import(_Inout_ SCOSSL_PROV_RSA_KEY_CTX
...
+                cbPrivateExponent = p->data_size;
+
+                pbPrivateExponent = OPENSSL_zalloc(cbPrivateExponent);
+                if(pbPrivateExponent == NULL)
+                {
+                    ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+                    goto cleanup;
+                }
+                if (!OSSL_PARAM_get_BN(p, &bn) ||
+                    !BN_bn2bin(bn, pbPrivateExponent))
```
ccalik, line 1180: Should we not check `cbPrivateExponent >= BN_num_bytes(bn)` or is it always satisfied?

mamckee: This will be satisfied by `pbPrivateExponent = p->data_size`.

Technically the BN is encoded as an octet string with an integer parameter type under the hood, but it's encoded using `BN_bn2nativepad`. Fetching as a BN and re-encoding with `BN_bn2bin` ensures we have the correct endianness.

samuel-lee-msft (dismissed):
LGTM

mamckee:
Overrode branch protections to merge with conflict resolutions.

PR #101, opened 2024-12-03
PBKDF2

mamckee:
This PR adds PBKDF2 and associated tests to the SymCrypt provider.


SymCryptProvider/src/kdf/p_scossl_pbkdf2.c
```diff
@@ -0,0 +1,362 @@
...
+
+    return ctx;
+}
+
+void p_scossl_pbkdf2_freectx(_Inout_ SCOSSL_PROV_PBKDF2_CTX *ctx)
+{
+    if (ctx == NULL)
+        return;
+
+    SymCryptWipeKnownSize(&ctx->expandedKey, sizeof(SYMCRYPT_SRTPKDF_EXPANDED_KEY));
```
samuel-lee-msft, line 70: >SYMCRYPT_SRTPKDF_EXPANDED_KEY [](http://example.com/codeflow?start=52&length=29)

SYMCRYPT_PBKDF2_EXPANDED_KEY?

SymCryptProvider/src/kdf/p_scossl_pbkdf2.c
```diff
@@ -0,0 +1,362 @@
...
+#include "p_scossl_base.h"
+
+#include <openssl/proverr.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Constants defined SP800-132 that should be checked
+// unless the OSSL_KDF_PARAM_PKCS5 paremeter gets set
```
samuel-lee-msft, line 14: >paremeter [](http://example.com/codeflow?start=35&length=9)

NP: parameter

SymCryptProvider/src/kdf/p_scossl_pbkdf2.c
```diff
@@ -0,0 +1,362 @@
...
+
+SCOSSL_PROV_PBKDF2_CTX *p_scossl_pbkdf2_newctx(SCOSSL_PROVCTX *provctx)
+{
+    SCOSSL_PROV_PBKDF2_CTX *ctx = OPENSSL_zalloc(sizeof(SCOSSL_PROV_PBKDF2_CTX));
+
+    if (ctx != NULL)
+    {
+        ctx->libctx = provctx->libctx;
+        ctx->pMac = SymCryptHmacSha1Algorithm;
+        ctx->iterationCount = PKCS5_DEFAULT_ITER;
```
samuel-lee-msft, line 58: >PKCS5_DEFAULT_ITER [](http://example.com/codeflow?start=30&length=18)

I think this constant should probably be defined locally in this file too

SymCryptProvider/src/kdf/p_scossl_pbkdf2.c
```diff
@@ -0,0 +1,362 @@
...
+        ctx->checkMinSizes = TRUE;
```
samuel-lee-msft, line 59: >TRUE [](http://example.com/codeflow?start=29&length=4)

I think if we want backwards compatibility then the default should be permissive (i.e. FALSE)?

SymCryptProvider/src/kdf/p_scossl_pbkdf2.c
```diff
@@ -0,0 +1,362 @@
...
+                    ERR_raise(ERR_LIB_PROV, ERR_R_INTERNAL_ERROR);
+                    goto cleanup;
+                }
+
+                copyCtx->initialized = TRUE;
+            }
+            else
+            {
+                copyCtx->initialized = FALSE;
+            }
```
samuel-lee-msft, line 112: This seems a little weird to me; when we set password / mac in `p_scossl_pbkdf2_set_ctx_params`, we explicitly do not eagerly initialize expandedKey, but wait until the next call to `p_scossl_pbkdf2_derive`.

This logic seems to eagerly expand the key in the destination context even if it may not yet be expanded in the source context; which just seems overcomplex vs. always setting copyCtx->initialized = FALSE?

SymCryptProvider/src/kdf/p_scossl_pbkdf2.c
mamckee: Yeah, I originally matched this to the FIPS provider's behavior but that would align with the rest of the design.

SymCryptProvider/src/kdf/p_scossl_pbkdf2.c
mamckee: There should have been a check for `ctx->initialized == TRUE` before expanding in `copyCtx`. I think your suggestion makes more sense though.

SymCryptProvider/src/kdf/p_scossl_pbkdf2.c
```diff
@@ -0,0 +1,350 @@
...
+        {
+            if ((ctx->pbPassword = OPENSSL_secure_malloc(p->data_size)) == NULL)
+            {
+                ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+                goto cleanup;
+            }
+
+            if (!OSSL_PARAM_get_octet_string(p, (void **)&ctx->pbPassword, p->data_size, &ctx->cbPassword))
+            {
+                OPENSSL_secure_clear_free(ctx->pbPassword, ctx->cbPassword);
```
samuel-lee-msft, line 255: >ctx->cbPassword [](http://example.com/codeflow?start=59&length=15)

Is this guaranteed to be set correctly on failure?
Suggest setting ctx->cbPassword to p->data_size before OPENSSL_secure_malloc and using ctx->cbPassword instead of p->data_size from then on?

samuel-lee-msft (dismissed):
LGTM, modulo single comment

samuel-lee-msft (approved):

PR #105, opened 2025-02-03
Fix SHAKE and CSHAKE squeeze output length

mamckee:
SHAKE and CSHAKE allow the `OSSL_DIGEST_PARAM_XOFLEN` parameter to be set. This should only be used as the output length for the "final" function. The `outlen` parameter passed to the "squeeze" function should be used as the output length for that case.

This PR updates the SymCrypt provider's SHAKE and CSHAKE interface to use the correct output length for squeeze.


ccalik (approved):

samuel-lee-msft (approved):
:shipit:

PR #110, opened 2025-03-03
KeysInUse and memory fixes

### LINKED ISSUE #108: Memory leak in EVP_PKEY_new_raw_private_key
### TL;DR

* Key type: **`EVP_PKEY_X25519`**
* Reproduced on **Azure Linux 3.0** with SymCrypt backend, **openssl 3.3.2**
* ASAN reports memory leaks when creating a private key from raw data with **`EVP_PKEY_new_raw_private_key`**.
* `strace`-ing proves SymCrypt backend is used.
* Building openssl 3.3.2 from source and linking against it fixed the leak, therefore pointing to SymCrypt backend involvement.

### More details

I'm not sure if `X25519` is fully supported here. I found https://github.com/microsoft/SymCrypt/issues/43 and [this](https://github.com/dotnet/runtime/issues/63174), which is about adding support for multiple EdDSA variations, and it seems it hasn't been done yet.

However, I'm able to successfully create `X25519` keys in a freshly created Azure Linux 3.0 container.

### Questions to address
* Is X25519 supported by SymCrypt? And if not, is it OK that `EVP_PKEY_CTX_new_id(EVP_PKEY_X25519, NULL)` succeeds?
* Am I using correct/wrong OpenSSL API to create/copy the keys?
* If the latter, how do I change it to avoid the leak?

### Code to repro

```
#include <cassert>
#include <openssl/evp.h>
#include <vector>

int main()
{
  const int curve_nid = EVP_PKEY_X25519;

  // Generate a brand new key.
  EVP_PKEY* key = EVP_PKEY_new();
  auto pctx = EVP_PKEY_CTX_new_id(curve_nid, NULL);

  assert(pctx != nullptr);
  assert(EVP_PKEY_keygen_init(pctx) == 1);
  assert(EVP_PKEY_keygen(pctx, &key) == 1);

  EVP_PKEY_CTX_free(pctx);

  // Copy private key bytes.
  std::vector<uint8_t> raw_priv(EVP_PKEY_size(key));
  size_t raw_priv_len = raw_priv.size();
  EVP_PKEY_get_raw_private_key(key, raw_priv.data(), &raw_priv_len);

  // Create new private key from raw bytes
  // ! LEAKS MEMORY !
  EVP_PKEY* another_key = EVP_PKEY_new_raw_private_key(
    curve_nid, nullptr, raw_priv.data(), raw_priv.size());

  // Free both.
  EVP_PKEY_free(key);
  EVP_PKEY_free(another_key);

  // Try to force-cleanup all potential leftovers, does not mitigate the leak.
  EVP_cleanup();
  CRYPTO_cleanup_all_ex_data();
}
```

Compile (replace paths if needed):  `clang file_name.cpp /usr/lib/libcrypto.so.3 /usr/lib/libstdc++.so -g -fsanitize=address`

Run: `./a.out`

Produces:
```
root [ /workspace ]# ./a.out
                                                                                                                                                        =================================================================
==691==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 56 byte(s) in 1 object(s) allocated from:
    #0 0x55e549f548df in malloc (/workspace/a.out+0x1158df)
    microsoft/SymCrypt#1 0x7ff165f63c30 in CRYPTO_malloc (/usr/lib/libcrypto.so.3+0x218c30)

SUMMARY: AddressSanitizer: 56 byte(s) leaked in 1 allocation(s).
```

Strace: 
```
root [ /workspace ]# strace -o out.txt ./a.out
...
root [ /workspace ]# grep "symcrypt" out.txt
access("/etc/pki/tls/symcrypt_prov.cnf", R_OK) = 0
openat(AT_FDCWD, "/etc/pki/tls/symcrypt_prov.cnf", O_RDONLY) = 3
openat(AT_FDCWD, "/usr/lib/ossl-modules/symcryptprovider.so", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/usr/lib/libsymcrypt.so.103", O_RDONLY|O_CLOEXEC) = 3
```

mamckee:
This PR fixes a couple bugs related to keysinuse and memory leaks found with additional valgrind testing.
- Create `/var/log/keysinuse` with correct permissions and ownership if not found on keysinuse start.
- Fix double alloc in X25519 case. Fixes #108 
- Fix edge case potential deadlock with RSA keys
- Add static flag to initialize DH and ECC curves only once. If the engine and provider are both loaded these hidden curves were re-initialized and the originally created curves are leaked.
- Fix some additional leaks found with valgrind testing

### DIFF
--- a/SymCryptProvider/src/p_scossl_keysinuse.c
+++ b/SymCryptProvider/src/p_scossl_keysinuse.c
@@ -149,7 +149,16 @@ static void p_scossl_keysinuse_init_once()
     sk_keysinuse_info = sk_SCOSSL_PROV_KEYSINUSE_INFO_new_null();
 
     // Make sure /var/log/keysinuse exists
-    if (mkdir(LOG_DIR, 1733) == -1 && errno != EEXIST)
+    if (mkdir(LOG_DIR, 1733) == 0)
+    {
+        if (chown(LOG_DIR, 0, 0) == -1)
+        {
+            p_scossl_keysinuse_log_error("Failed to set ownership of logging directory at %s,SYS_%d", LOG_DIR, errno);
+            rmdir(LOG_DIR);
+            goto cleanup;
+        }
+    }    
+    else if (errno != EEXIST)
     {
         p_scossl_keysinuse_log_error("Failed to create logging directory at %s,SYS_%d", LOG_DIR, errno);
         goto cleanup;
--- a/SymCryptProvider/src/p_scossl_keysinuse.c
+++ b/SymCryptProvider/src/p_scossl_keysinuse.c
@@ -30,7 +30,8 @@ static BOOL keysinuse_enabled = FALSE;
 #define KEYSINUSE_NOTICE 1
 // Log files separated by UID.
 // /var/log/keysinuse/keysinuse_<level>_<euid>.log
-#define LOG_PATH_TMPL "/var/log/keysinuse/keysinuse_%.3s_%08x.log"
+#define LOG_DIR       "/var/log/keysinuse"
+#define LOG_PATH_TMPL LOG_DIR "/keysinuse_%.3s_%08x.log"
 #define LOG_MSG_MAX 256
 static const char *default_prefix = "";
 static char *prefix = NULL;
@@ -146,6 +147,13 @@ static void p_scossl_keysinuse_init_once()
 
     sk_keysinuse_info_lock = CRYPTO_THREAD_lock_new();
     sk_keysinuse_info = sk_SCOSSL_PROV_KEYSINUSE_INFO_new_null();
+
+    // Make sure /var/log/keysinuse exists
+    if (mkdir(LOG_DIR, 1733) == -1 && errno != EEXIST)
+    {
+        p_scossl_keysinuse_log_error("Failed to create logging directory at %s,SYS_%d", LOG_DIR, errno);
+        goto cleanup;
+    }
 
     // Start the logging thread. Monotonic clock needs to be set to
     // prevent wall clock changes from affecting the logging delay sleep time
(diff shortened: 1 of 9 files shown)


samuel-lee-msft (approved):
:shipit:

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -147,6 +150,28 @@ static void p_scossl_keysinuse_init_once()
...
+    if (mkdirResult == 0)
+    {
+        if (chown(LOG_DIR, 0, 0) == -1)
+        {
+            p_scossl_keysinuse_log_error("Failed to set ownership of logging directory at %s,SYS_%d", LOG_DIR, errno);
+            rmdir(LOG_DIR);
+            goto cleanup;
+        }
+    }    
+    else if (errno != EACCES && errno != EEXIST)
```
samuel-lee-msft, line 169: >errno != EACCES [](http://example.com/codeflow?start=13&length=15)

I'm confused.

This seems to be saying that we continue with KeysInUse initialization if mkdir fails because we have no permissions, even though we won't have a directory to write to?

mamckee: Looks like I got the order of errors mixed up in testing. I thought I `EACCES` was returned if the caller didn't have permissions, even if the directory was present, but it looks like `EEXIST` is returned first. The order isn't guaranteed by the documentation though: https://man7.org/linux/man-pages/man2/mkdir.2.html.

I think in the case of `EACCES` we should still proceed though. KeysInUse just won't log until the directory is created. It may be created later by another process with correct permissions or an agent consuming these logs. This is just a best attempt to create the directory in case it's missing. The rest of the errors from `mkdir` would indicate KeysInUse will never succeed though (such as `EDQUOT`), in which case we'd just want to bail early.

samuel-lee-msft (approved):

PR #112, opened 2025-03-19
Proper state management for AES GCM and CCM

mamckee:
The SymCrypt unit tests with OpenSSL and SCOSSL revealed an edge case when calling `EVP_CipherInit_ex2` and passing the IV length as a parameter alongside the IV. The SCOSSL implementation unconditionally freed the set IV if an IV length was set, even if the IV length did not change. If the IV length matches what's set, then the IV should not be reset.

This PR also adds state checks for the IV and tag in CCM to bring the behavior and failure points more in line with the default provider.

### DIFF
--- a/SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
+++ b/SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
@@ -1064,8 +1064,8 @@ static BOOL p_scossl_rsa_keymgmt_match(_In_ SCOSSL_PROV_RSA_KEY_CTX *keyCtx1, _I
 cleanup:
     OPENSSL_free(pbModulus1);
     OPENSSL_free(pbModulus2);
-    OPENSSL_secure_free(pbPrivateExponent1);
-    OPENSSL_secure_free(pbPrivateExponent2);
+    OPENSSL_secure_clear_free(pbPrivateExponent1, cbModulus);
+    OPENSSL_secure_clear_free(pbPrivateExponent2, cbModulus);
 
     return ret;
 }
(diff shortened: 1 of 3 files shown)


SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
```diff
@@ -1271,6 +1271,7 @@ static SCOSSL_STATUS p_scossl_rsa_keymgmt_import(_Inout_ SCOSSL_PROV_RSA_KEY_CTX
     ret = SCOSSL_SUCCESS;
 
 cleanup:
+    OPENSSL_free(pbPrivateExponent);
```
samuel-lee-msft, line 1274: >OPENSSL_free [](http://example.com/codeflow?start=4&length=12)

Should this have been `OPENSSL_secure_malloc`ed and be `OPENSSL_secure_clear_free`ed?

SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
```diff
@@ -1271,6 +1271,7 @@ static SCOSSL_STATUS p_scossl_rsa_keymgmt_import(_Inout_ SCOSSL_PROV_RSA_KEY_CTX
...
     OPENSSL_free(pbModulus);
     OPENSSL_free(ppbPrimes[0]);
     OPENSSL_free(ppbPrimes[1]);
```
samuel-lee-msft, line 1277: Ditto for the private primes.

samuel-lee-msft (approved):
LGTM modulo comments

SymCryptProvider/src/keymgmt/p_scossl_rsa_keymgmt.c
mamckee: Yeah, same for the primes.

samuel-lee-msft (approved):

PR #117, opened 2025-05-14
Support digest state export for MD5, SHA1, and SHA2 digests

mamckee:
OpenSSL does not currently support resumable hashing, where an in-progress hash operation's state is saved, and can be loaded later to finish the hash. Some crypto libraries do support this such as golang/crypto. For libraries build as a compatibility layer on top of OpenSSL (e.g., [golang-fips/openssl](https://github.com/golang-fips/openssl), there is no way through the OpenSSL APIs to support resumable hashing.

- Adds custom parameters `state` (`OSSL_PARAM_OCTET_STRING`) and `recompute_checksum` (`OSSL_PARAM_INTEGER`) to the SymCrypt provider for MD5, SHA1, and SHA2 digests. 
    - The exported hash state can be exported and imported by supplying the `state` parameter to `EVP_MD_CTX_get_params` and `EVP_MD_CTX_set_params` respectively. 
    - If the state is constructed outside of SymCrypt, then `recompute_checksum` can optionally be set to tell the SymCrypt provider to recalculate the checksum SymCrypt expects at the end of the export blob.
- Adds imported/exported digest tests to SslPlay.
- Add an option to SslPlay to disable loading the SymCrypt engine. 
    - If the engine is loaded, OpenSSL will always forward digest and cipher operations to the engine instead of the provider
   
- Bump version to 1.9.0
- Fixed typo in SymCrypt provider readme
- Disable optimizations for debug builds
- Fix SslPlay AES-XTS test. This test set an 8 byte IV but didn't explicitly try to change the IV size, causing a buffer overrun if the provider's default IV length was > 8

### DIFF
--- a/SslPlay/SslPlay.cpp
+++ b/SslPlay/SslPlay.cpp
@@ -1178,6 +1178,7 @@ end:
 bool TestDigestImportExport(const char *digestname, const char *expected, size_t export_state_size)
 {
     bool result = false;
+    EVP_MD *md = NULL;
     EVP_MD_CTX *mdctx;
     char mess1[] = "Test Message1234567";
     char mess2[] = "Hello World";
@@ -1197,7 +1198,7 @@ bool TestDigestImportExport(const char *digestname, const char *expected, size_t
 
     printf("\nTestDigestImportExport: %s\n\n", digestname);
 
-    EVP_MD *md = EVP_MD_fetch(nullptr, digestname, "provider=symcryptprovider");
+    md = EVP_MD_fetch(nullptr, digestname, "provider=symcryptprovider");
     if (md == NULL)
     {
         printf("No Digest found for %s\n", digestname);
@@ -1285,6 +1286,8 @@ bool TestDigestImportExport(const char *digestname, const char *expected, size_t
         goto end;
     }
 
+    result = true;
+
 end:
     EVP_MD_free(md);
 
@@ -1332,13 +1335,13 @@ void TestDigests(bool useEngine)
         {"SHA512-256", "036dbd97db1e37aabe6ded8ef9ead203e9adb02ad5596ac5af072dd7374993a0",
             SYMCRYPT_SHA512_256_STATE_EXPORT_SIZE},
         {"SHA3-224", "489a032b8923a05eca5b40f2ed9838f218c65bd082acc48fa2067213",
-            0},
+            SYMCRYPT_SHA3_224_STATE_EXPORT_SIZE},
         {"SHA3-256", "375e793a6d4e4947658e78cb697789434b8279feb2ec9595d03e44473ac478f6",
-            0},
+            SYMCRYPT_SHA3_256_STATE_EXPORT_SIZE},
         {"SHA3-384", "1d47002a9e96d5b6bdd70d476fd2038e50ac3eb0d4202b4eb988f02185fbb9c85cb7ed62804ddaff894e84d62e5832f2",
-            0},
+            SYMCRYPT_SHA3_384_STATE_EXPORT_SIZE},
         {"SHA3-512", "bf63544ae59243a5419a3ff5f598352eb1409d41dc746c9e9d5f258cddaff4e7f7b9d9ae13e90eb07f27e4e157b3fcf796f6554732a2e78a621f7313aba827f3",
-            0},
+            SYMCRYPT_SHA3_512_STATE_EXPORT_SIZE},
 #endif
     };
 
@@ -1556,15 +1559,15 @@ void TestAesXts()
 {
     unsigned char plaintext[8192];
     int plaintext_len = 64;
-    unsigned char iv[8];
+    unsigned char iv[16];
     unsigned char key[64];
 
     while(!RAND_bytes(key, 64));
-    while(!RAND_bytes(iv, 8));
+    while(!RAND_bytes(iv, 16));
     while(!RAND_bytes(plaintext, plaintext_len));
 
-    TestAesCipher("EVP_aes_128_xts", EVP_aes_128_xts(), key, 32, iv, 8, plaintext, plaintext_len);
-    TestAesCipher("EVP_aes_256_xts", EVP_aes_256_xts(), key, 64, iv, 8, plaintext, plaintext_len);
+    TestAesCipher("EVP_aes_128_xts", EVP_aes_128_xts(), key, 32, iv, 16, plaintext, plaintext_len);
+    TestAesCipher("EVP_aes_256_xts", EVP_aes_256_xts(), key, 64, iv, 16, plaintext, plaintext_len);
 
     printf("%s", SeparatorLine);
     return;
@@ -2363,7 +2366,7 @@ void TestSshKdf(void)
         goto end;
     }
 
-    //f
+    //
     // Test both EVP_KDF_ctrl and EVP_KDF_ctrl_str functions.
     // bCtrlStrMode = 0 uses EVP_KDF_ctrl functions and
     // bCtrlStrMode = 1 uses EVP_KDF_ctrl_str functions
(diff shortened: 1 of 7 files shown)


SslPlay/SslPlay.cpp
```diff
@@ -2184,7 +2363,7 @@ void TestSshKdf(void)
         goto end;
     }
 
-    //
+    //f
```
samuel-lee-msft, line 2366: >f [](http://example.com/codeflow?start=6&length=1)

NP: stray change

SymCryptProvider/src/digests/p_scossl_digest_generic.c
```diff
@@ -10,6 +11,94 @@
...
+typedef VOID (SYMCRYPT_CALL * PSYMCRYPT_DIGEST_STATE_EXPORT) (PVOID pState, PBYTE pbBlob);
+typedef SYMCRYPT_ERROR (SYMCRYPT_CALL * PSYMCRYPT_DIGEST_STATE_IMPORT) (PVOID pState, PCBYTE pbBlob);
+
+const OSSL_PARAM p_scossl_digest_export_settable_param_types[] = {
+    OSSL_PARAM_octet_string(SCOSSL_DIGEST_PARAM_STATE, NULL, 0),
+    OSSL_PARAM_int(SCOSSL_DIGEST_PARAM_RECOMPUTE_CHECKSUM, NULL),
+    OSSL_PARAM_END};
+
+const OSSL_PARAM p_scossl_digest_export_gettable_ctx_param_types[] = {
+    OSSL_PARAM_int(SCOSSL_DIGEST_PARAM_STATE, NULL),
```
samuel-lee-msft, line 26: >OSSL_PARAM_int [](http://example.com/codeflow?start=4&length=14)

OSSL_PARAM_octet_string?

samuel-lee-msft:
> IMPLEMENT_SCOSSL_DIGEST_GENERIC(Sha3_512, sha3_512, SCOSSL_DIGEST_FLAG_ALGID_ABSENT)

Just skipped these to reduce risk of regression, or because there is something fundamentally harder about making SHA3 exportable?

If the implementation is the same shape I would go for it in this PR even if golang won't use it yet.

---
Refers to: SymCryptProvider/src/digests/p_scossl_digest_generic.c:184 in 520bd0b. [](commit_id = 520bd0b51756de46378fece0e1070bece51293c7, deletion_comment = False)

SymCryptProvider/src/digests/p_scossl_digest_generic.c
```diff
@@ -33,23 +122,60 @@ static SCOSSL_STATUS p_scossl_digest_generic_final(_In_ SCOSSL_DIGEST_CTX *ctx,
...
+                                                             _In_ const OSSL_PARAM params[])        \
+    {                                                                                               \
+        SymCryptHashInit(ctx->pHash, ctx->pState);                                                  \
+        return p_scossl_digest_set_##state_name(ctx, params);                                       \
+    }
+
+IMPLEMENT_SCOSSL_DIGEST_EXPORT_FUNCTIONS(md5_state, SymCryptMd5StateExport, SymCryptMd5StateImport, SYMCRYPT_MD5_STATE_EXPORT_SIZE)
+IMPLEMENT_SCOSSL_DIGEST_EXPORT_FUNCTIONS(sha1_state, SymCryptSha1StateExport, SymCryptSha1StateImport, SYMCRYPT_SHA1_STATE_EXPORT_SIZE)
+IMPLEMENT_SCOSSL_DIGEST_EXPORT_FUNCTIONS(sha256_state, SymCryptSha256StateExport, SymCryptSha256StateImport, SYMCRYPT_SHA256_STATE_EXPORT_SIZE)
+IMPLEMENT_SCOSSL_DIGEST_EXPORT_FUNCTIONS(sha512_state, SymCryptSha512StateExport, SymCryptSha512StateImport, SYMCRYPT_SHA512_STATE_EXPORT_SIZE)
```
samuel-lee-msft, line 150: >SymCryptSha512StateExport, SymCryptSha512StateImport [](http://example.com/codeflow?start=55&length=52)

We have distinct SymCryptSha(512|384|512_224|512_256)State(Export|Import) functions - won't `SymCryptSha512StateImport` fail at import time with a constructed Sha384 blob?

Similar for SHA256 vs. SHA224

SymCryptProvider/src/digests/p_scossl_digest_generic.c
```diff
@@ -10,6 +11,94 @@
...
+const OSSL_PARAM *p_scossl_digest_export_gettable_ctx_params(ossl_unused void *ctx, ossl_unused void *provctx)
+{
+    return p_scossl_digest_export_gettable_ctx_param_types;
+}
+
+static SCOSSL_STATUS p_scossl_digest_get_state_internal(_In_ SCOSSL_DIGEST_CTX *ctx, _Inout_ OSSL_PARAM params[],
+                                                        _In_ PSYMCRYPT_DIGEST_STATE_EXPORT pExportFunc,
+                                                        SIZE_T cbExportBlob)
+{
+    BYTE pbExportBlob[cbExportBlob];
```
samuel-lee-msft, line 43: >cbExportBlob [](http://example.com/codeflow?start=22&length=12)

I don't think we use VLAs anywhere else in SymCrypt/SCOSSL.

Can we just set this to `SCOSSL_MAX_STATE_EXPORT_BLOB_SIZE` (currently `SYMCRYPT_SHA512_STATE_EXPORT_SIZE`)?

mamckee:
> IMPLEMENT_SCOSSL_DIGEST_GENERIC(Sha3_512, sha3_512, SCOSSL_DIGEST_FLAG_ALGID_ABSENT)

The PR is still a draft. I wanted to validate existing hashes with the golang-fips toolchain first. I'll be pushing SHA3 and cleanup before publishing this PR.

---
In reply to: [2881798636](https://github.com/microsoft/SymCrypt-OpenSSL/pull/117#issuecomment-2881798636) [](http://example.com/codeflow?ancestors=2881798636)

---
Refers to: SymCryptProvider/src/digests/p_scossl_digest_generic.c:184 in 520bd0b. [](commit_id = 520bd0b51756de46378fece0e1070bece51293c7, deletion_comment = False)

samuel-lee-msft (dismissed):
LGTM - assuming everything is working with the golang side, I'm happy to merge.

SslPlay/SslPlay.cpp
```diff
@@ -2302,17 +2490,79 @@ int main(int argc, char** argv)
...
+            printf("  --log-level <log level>           Specify the SCOSSL logging level\n");
+            printf("  --err-level <err level>           Specify the SCOSSL error logging level\n");
+#if OPENSSL_VERSION_MAJOR == 3
+            printf("  --provider-path <provider path>   Specify a directory to locate the symcrypt provider\n");
+            printf("  --no-engine                       Disable the SymCrypt engine for testing\n");
+#endif
+            return 0;
+        }
+    }
     if (argc >= 2) {
```
tobiasb-ms, line 2552: These are now wrong and if any parameters are passed with the new model, they will overwrite `scossl_log_level` and `scossl_ossl_ERR_level`, usually with `0` (since that's what `atoi` returns when it can't parse something).

This brings up an interesting point -- this is "just" test code, but by adding this (objectively better) way of parsing parameters, it creates an incompatibility that could break an existing test pipeline.

tobiasb-ms (changes requested):

SslPlay/SslPlay.cpp
samuel-lee-msft: Great catch - I would say we make absolutely no guarantees about SslPlay compatibility across SCOSSL versions, but we should remove this old arg parsing logic which is broken.

mamckee: Thanks for catching this. I must have undone removing this code when I was looking at the diff.

tobiasb-ms: Yeah, I guess that `SslPlay` would only be used if you build `SCOSSL` yourself, in which case you're actively upgrading so that makes sense.

samuel-lee-msft (approved):

PR #118, opened 2025-05-20
TLS1.3KDF support

MS-megliu:
SymCrypt-OpenSSL Provider should implement the TLS13-KDF.  Added support.
https://docs.openssl.org/3.0/man7/EVP_KDF-TLS13_KDF/

Testing:
Used Openssl's evpkdf_tls13_kdf.txt data file, all 562 test cases passed.

    # Reading EvpTestRecipes/3.0/evpkdf_tls13_kdf.txt
    # INFO:  @ test/testutil/stanza.c:122
    # Starting "TLS 1.3 KDF tests (from ACVP test vectors)" tests at line 14
    # INFO:  @ test/testutil/stanza.c:122
    # Starting "TLS13-KDF bad mode test" tests at line 4931
    # INFO:  @ test/testutil/stanza.c:32
    # Completed 562 tests with 0 errors and 0 skipped
    ok 1 - iteration 1
ok 1 - run_file_tests


qmuntal:
FYI: The Go OpenSSL backend just recently added support for using the TLS13-KDFwhen available to do TLS 1.3 handshakes: https://github.com/golang-fips/openssl/pull/272.

mamckee:
> 

We should match the OpenSSL behavior and fail in `p_scossl_tls13kdf_set_ctx_params` if the mode is `EVP_KDF_HKDF_MODE_EXTRACT_AND_EXPAND` instead of in derive. In either case the operation fails, but we should minimize behavior differences with the OpenSSL default wherever possible.

---
Refers to: EvpTestRecipes/3.0/evpkdf_tls13_kdf.txt:4937 in 8e600be. [](commit_id = 8e600bea53c7a1a5a04ffd5a8ddb6d32310088b1, deletion_comment = False)

ScosslCommon/inc/scossl_hkdf.h
```diff
@@ -6,6 +6,7 @@
 
 #include <openssl/kdf.h>
 
+
```
mamckee, line 9: Nit: extra line

ScosslCommon/inc/scossl_hkdf.h
```diff
@@ -51,6 +58,9 @@ SCOSSL_STATUS scossl_hkdf_append_info(_Inout_ SCOSSL_HKDF_CTX *ctx,
 SCOSSL_STATUS scossl_hkdf_derive(_In_ SCOSSL_HKDF_CTX *ctx,
                                  _Out_writes_bytes_(keylen) PBYTE key, SIZE_T keylen);
 
+SCOSSL_STATUS scossl_tls13kdf_derive(_In_ SCOSSL_HKDF_CTX *ctx,
```
mamckee, line 61: Remove this now that the code has been moved.

ScosslCommon/inc/scossl_hkdf.h
```diff
@@ -37,6 +38,12 @@ typedef struct
     SIZE_T cbKey;
     BYTE info[HKDF_MAXBUF];
     SIZE_T cbInfo;
+    PBYTE pbLabel;
```
mamckee, line 41: Add a comment here to indicate these fields are only used for TLS1.3 KDF.

ScosslCommon/src/scossl_hkdf.c
```diff
@@ -4,6 +4,7 @@
 
 #include "scossl_hkdf.h"
 
+
```
mamckee, line 7: Nit: extra line

ScosslCommon/src/scossl_hkdf.c
```diff
@@ -41,6 +42,39 @@ SCOSSL_HKDF_CTX *scossl_hkdf_dupctx(SCOSSL_HKDF_CTX *ctx)
         }
         copyCtx->cbKey = ctx->cbKey;
 
+        if (ctx->pbPrefix == NULL)
```
mamckee, line 45: >if (ctx->pbPrefix == NULL) [](http://example.com/codeflow?start=8&length=26)

This whole function could use some cleanup, especially now that we have these extra fields. We should just zalloc at the top, only copy if `ctx->someMember != NULL`, and just assign all of the sizes at the bottom.

Something like what we do in `p_scossl_sskdf_dupctx` would probably be best.

SymCryptProvider/src/p_scossl_names.h
```diff
@@ -67,6 +67,7 @@ extern "C" {
 #define SCOSSL_ALG_NAME_SSHKDF      SN_sshkdf
 #define SCOSSL_ALG_NAME_SSKDF       SN_sskdf
 #define SCOSSL_ALG_NAME_TLS1_PRF    SN_tls1_prf
+#define SCOSSL_ALG_NAME_TLS13_KDF   OSSL_KDF_NAME_TLS1_3_KDF
```
mamckee, line 70: I think it makes sense to replace all of these KDF names with the `OSSL_KDF_NAME_*` macro. SRTPKDF and SRTCPKDF need to stay hardcoded though since OpenSSL doesn't support them by default.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+    return SCOSSL_SUCCESS;
+}
+
+const OSSL_DISPATCH p_scossl_tls13kdf_kdf_functions[] = {
+    {OSSL_FUNC_KDF_NEWCTX, (void (*)(void))p_scossl_hkdf_newctx},
+    {OSSL_FUNC_KDF_FREECTX, (void (*)(void))p_scossl_hkdf_freectx},
+    {OSSL_FUNC_KDF_DUPCTX, (void (*)(void))p_scossl_hkdf_dupctx},
+    {OSSL_FUNC_KDF_RESET, (void (*)(void))p_scossl_hkdf_reset},
+    {OSSL_FUNC_KDF_DERIVE, (void (*)(void))p_scossl_tls13kdf_derive},
+    {OSSL_FUNC_KDF_GETTABLE_CTX_PARAMS, (void (*)(void))p_scossl_tls13kdf_gettable_ctx_params},
```
mamckee, line 762: >p_scossl_tls13kdf_gettable_ctx_params [](http://example.com/codeflow?start=56&length=37)

It looks like we can just use `p_scossl_hkdf_gettable_ctx_params` and `p_scossl_hkdf_get_ctx_params`

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
     {0, NULL}};
 
+/*
+ * TLS1.3KDF uses slight variations of the above,
+ * they need to be present here.
+ * Refer to RFC 8446 section 7 for specific details.
+ */
+
+
+#define HKDF_COMMON_SETTABLES                                       \
```
mamckee, line 363: Move these macros to the top. You can use this for `p_scossl_hkdf_settable_ctx_param_types` too

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+    OSSL_PARAM_int(OSSL_KDF_PARAM_MODE, NULL),                      \
+    OSSL_PARAM_utf8_string(OSSL_KDF_PARAM_DIGEST, NULL, 0),         \
+    OSSL_PARAM_octet_string(OSSL_KDF_PARAM_SALT, NULL, 0),          \
+    OSSL_PARAM_octet_string(OSSL_KDF_PARAM_INFO, NULL, 0)
+
+static const OSSL_PARAM p_scossl_tls13kdf_gettable_ctx_param_types[] = {
+    HKDF_COMMON_GETTABLES,
+    OSSL_PARAM_END};
+
+static const OSSL_PARAM p_scossl_tls13kdf_settable_ctx_param_types[] = {
```
mamckee, line 387: This should go under `p_scossl_hkdf_settable_ctx_param_types`. Generally these files should be structured:  
1. Macros
2. Constants (like gettable/settable arrays)
3. Functions
4. `OSSL_DISPATCH` tables

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+ * struct {
+ *     uint16 length;             // Desired length of the output keying material (2 bytes, big-endian)
+ *     opaque label<7..255>;      // A variable-length label prefixed with "tls13 " followed by a custom label
+ *     opaque context<0..255>;    // A variable-length context (e.g., handshake transcript hash)
+ * } HkdfLabel;
+ *
+ */
+
+ static
+SCOSSL_STATUS p_scossl_tls13_hkdf_expand(SCOSSL_HKDF_CTX *ctx, PBYTE key, SIZE_T keylen)
```
mamckee, line 409: This, `p_scossl_tls13kdf_generate_secret` and `p_scossl_tls13kdf_derive` should go under `p_scossl_hkdf_derive`

SymCryptProvider/src/kdf/p_scossl_hkdf.c
mamckee: In cases where we have multiple dispatch tables because of shared functions, then the "variants" of the same function should be next to each other. X25519 is a good example. We have the ECDH function, then the X25519 version.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+
+
+
+
+const OSSL_PARAM *p_scossl_tls13kdf_gettable_ctx_params(ossl_unused void *ctx, ossl_unused void *provctx)
+{
+    return p_scossl_tls13kdf_gettable_ctx_param_types;
+}
+
+const OSSL_PARAM *p_scossl_tls13kdf_settable_ctx_params(ossl_unused void *ctx, ossl_unused void *provctx)
```
mamckee, line 683: >p_scossl_tls13kdf_settable_ctx_params [](http://example.com/codeflow?start=18&length=37)

This should go under `p_scossl_hkdf_settable_ctx_params` and `p_scossl_tls13kdf_set_ctx_params` should go under `p_scossl_hkdf_set_ctx_params`

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+ *
+ * struct {
+ *     uint16 length;             // Desired length of the output keying material (2 bytes, big-endian)
+ *     opaque label<7..255>;      // A variable-length label prefixed with "tls13 " followed by a custom label
+ *     opaque context<0..255>;    // A variable-length context (e.g., handshake transcript hash)
+ * } HkdfLabel;
+ *
+ */
+
+ static
```
mamckee, line 408: Nit: remove newline

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+        &scExpandedKey,
+        hkdflabel, hkdflabellen,
+        key, keylen);
+    if (scError != SYMCRYPT_NO_ERROR)
+        return SCOSSL_FAILURE;
+
+    return SCOSSL_SUCCESS;
+}
+
+static
```
mamckee, line 484: Nit: remove newline

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+    if (expanded_secret != NULL) {
+        OPENSSL_free(expanded_secret);
+        expanded_secret = NULL;
+    }
+
+    return scError;
+}
+
+
+SCOSSL_STATUS p_scossl_tls13kdf_derive(_In_ SCOSSL_PROV_HKDF_CTX *ctx,
```
mamckee, line 626: The TLS1.3 KDF functions should all be static. Some of the HKDF functions aren't static because they're used in p_scossl_kdf_keyexch.c. That's for backwards compatibility scenarios that don't apply to TLS 1.3 so we can make all of these static.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+
+    SIZE_T totalLen = 2 + 1 + labelLen + 1 + ctx->cbData;
+    if (totalLen > HKDF_MAXBUF) {
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_HKDF_DERIVE, ERR_R_INTERNAL_ERROR,
+                         "Total size exceeds maximum buffer size allowed");
+        return SCOSSL_FAILURE;
+    }
+
+    BYTE hkdflabel[HKDF_MAXBUF];
+    size_t hkdflabellen = 0;
```
mamckee, line 442: >size_t [](http://example.com/codeflow?start=4&length=6)

Nit: SIZE_T

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+
+    // Output length (2 bytes) in big endian format
+    hkdflabel[hkdflabellen++] = (BYTE)((keylen >> 8) & 0xFF); //high byte
```
mamckee, line 445: >hkdflabel[hkdflabellen++] = (BYTE)((keylen >> 8) & 0xFF); //high byte [](http://example.com/codeflow?start=4&length=69)

Use `SYMCRYPT_STORE_MSBFIRST16(hkdflabel[hkdflabellen], keylen)`

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+        return SCOSSL_FAILURE;
+    }
+
+    symcryptHmacAlg = scossl_get_symcrypt_hmac_algorithm(EVP_MD_type(ctx->md));
+    if (symcryptHmacAlg == NULL) {
+        return SCOSSL_FAILURE;
+    }
+
+    SIZE_T labelLen = ctx->cbPrefix + ctx->cbLabel;
+    if (labelLen > 255)
```
mamckee, line 431: >255 [](http://example.com/codeflow?start=19&length=3)

Nit: Can you make this 0xFF and add a comment that the least-significant byte is copied into hkdlabel? If this exceeds 0xFF then we overflow on this field.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+        return SCOSSL_FAILURE;
+
+    SIZE_T totalLen = 2 + 1 + labelLen + 1 + ctx->cbData;
```
mamckee, line 434: Add a comment explaining the size (where the 2s and 1s come from)

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+
+    // Output length (2 bytes) in big endian format
+    hkdflabel[hkdflabellen++] = (BYTE)((keylen >> 8) & 0xFF); //high byte
+    hkdflabel[hkdflabellen++] = (BYTE)(keylen & 0xFF);  //low byte
+
+    // Label length
+    hkdflabel[hkdflabellen++] = (BYTE)labelLen;
+
+    // Label = prefix + label
+    memcpy(hkdflabel + hkdflabellen, ctx->pbPrefix, ctx->cbPrefix);
```
mamckee, line 452: >hkdflabel + hkdflabellen [](http://example.com/codeflow?start=11&length=24)

Nit: `hkdflabel[hkdflabellen]`

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+    hkdflabellen += ctx->cbPrefix;
+    memcpy(hkdflabel + hkdflabellen, ctx->pbLabel, ctx->cbLabel);
+    hkdflabellen += ctx->cbLabel;
+
+    // Context length
+    hkdflabel[hkdflabellen++] = (BYTE)ctx->cbData;
```
mamckee, line 458: >cbData [](http://example.com/codeflow?start=43&length=6)

Should we check that this is less than 256 like we do for labelLen or is that enforced elsewhere?

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+        memcpy(hkdflabel + hkdflabellen, ctx->pbData, ctx->cbData);
+        hkdflabellen += ctx->cbData;
+    }
+
+    // Expand PRK
+    scError = SymCryptHkdfPrkExpandKey(
+        &scExpandedKey,
+        symcryptHmacAlg,
+        ctx->pbKey, ctx->cbKey);
+    if (scError != SYMCRYPT_NO_ERROR)
```
mamckee, line 471: Add `SCOSSL_LOG_SYMCRYPT_ERROR` in cases where a SymCrypt call fails and we have an scError.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_HKDF_DERIVE, ERR_R_INTERNAL_ERROR, "Missing Key");
+        return SCOSSL_FAILURE;
+    }
+
+    symcryptHmacAlg = scossl_get_symcrypt_hmac_algorithm(EVP_MD_type(ctx->md));
+    if (symcryptHmacAlg == NULL) {
+        return SCOSSL_FAILURE;
+    }
+
+    SIZE_T labelLen = ctx->cbPrefix + ctx->cbLabel;
```
mamckee, line 430: >SIZE_T labelLen [](http://example.com/codeflow?start=4&length=15)

Nit: It looks like your declaring all your variables at the start of `p_scossl_tls13kdf_generate_secret` so probably best to do that here too.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
mamckee, line 409: >SCOSSL_HKDF_CTX *ctx, PBYTE key, SIZE_T keylen [](http://example.com/codeflow?start=41&length=46)

Missing SAL annotations here and on `p_scossl_tls13kdf_generate_secret`

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+    PCSYMCRYPT_MAC symcryptHmacAlg = NULL;
+    BYTE *default_zeros = NULL;
+    BYTE *empty_hash = NULL;
+    BYTE *expanded_secret = NULL;
+    EVP_MD_CTX *mctx = NULL;
+    SIZE_T mdlen;
+    BOOL salt_need_reset = FALSE;
+    BOOL data_need_reset = FALSE;
+    BOOL key_need_reset = FALSE;
+    PBYTE saved_key = NULL;
```
mamckee, line 497: >saved_key [](http://example.com/codeflow?start=10&length=9)

Nit: pbSavedKey and cbSavedKey.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+
+static
+SCOSSL_STATUS p_scossl_tls13kdf_generate_secret(SCOSSL_HKDF_CTX *ctx, PBYTE key, SIZE_T keylen)
+{
+    SYMCRYPT_ERROR scError = SYMCRYPT_NO_ERROR;
+    PCSYMCRYPT_MAC symcryptHmacAlg = NULL;
+    BYTE *default_zeros = NULL;
+    BYTE *empty_hash = NULL;
+    BYTE *expanded_secret = NULL;
+    EVP_MD_CTX *mctx = NULL;
```
mamckee, line 492: >mctx [](http://example.com/codeflow?start=16&length=4)

Nit: We've used mdctx elsewhere in the provider and engine.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+    if (mdlen <= 0)
+        return SCOSSL_FAILURE;
+
+    default_zeros = OPENSSL_zalloc(EVP_MAX_MD_SIZE);
+    if (default_zeros == NULL)
+        return SCOSSL_FAILURE;
+
+    symcryptHmacAlg = scossl_get_symcrypt_hmac_algorithm(EVP_MD_type(ctx->md));
+    if (symcryptHmacAlg == NULL) {
+        scError = SCOSSL_FAILURE;
```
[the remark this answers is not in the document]
mamckee: Probably best to replace all `return SCOSSL_FAILURE;` with `goto cleanup;` in this function to reduce the risk of bugs in the future (e.g., if the `default_zeros` alloc gets moved earlier in the function then we risk a memory leak.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+        goto cleanup;
+    }
+
+    if (ctx->pbKey == NULL) {
+        ctx->pbKey = default_zeros;
```
mamckee, line 522: >ctx->pbKey = default_zeros; [](http://example.com/codeflow?start=8&length=27)

I think setting and resetting these context members is prone to bugs. Can we just use a local `PBYTE pbKey, pbSalt , pbData` and `SIZE_T cbKey, cbSalt, cbData` that get set to the context value or default?

mamckee: Then we also only need to allocate `default_zeros` and `empty_hash` if the context members are empty.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+ */
+
+ static
+SCOSSL_STATUS p_scossl_tls13_hkdf_expand(SCOSSL_HKDF_CTX *ctx, PBYTE key, SIZE_T keylen)
+{
+    SYMCRYPT_ERROR scError = SYMCRYPT_NO_ERROR;
+    PCSYMCRYPT_MAC symcryptHmacAlg = NULL;
+    SYMCRYPT_HKDF_EXPANDED_KEY scExpandedKey;
+
+    if (ctx->md == NULL) {
```
mamckee, line 415: >{ [](http://example.com/codeflow?start=25&length=1)

Nit: Curly braces on new lines

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+        empty_hash = OPENSSL_zalloc(EVP_MAX_MD_SIZE);
+        if (empty_hash == NULL) {
+            scError = SCOSSL_FAILURE;
+            goto cleanup;
+        }
+        mctx = EVP_MD_CTX_new();
+        if (mctx == NULL ||
+            EVP_DigestInit_ex(mctx, ctx->md, NULL) <= 0 ||
+            EVP_DigestFinal_ex(mctx, empty_hash, NULL) <= 0) {
+            EVP_MD_CTX_free(mctx);
```
mamckee, line 541: This should just go after cleanup

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+        ctx->cbKey = mdlen;
+        key_need_reset = TRUE;
+    }
+
+    if (ctx->pbSalt == NULL) {
+        ctx->pbSalt = default_zeros;
+        ctx->cbSalt = mdlen;
+        salt_need_reset = TRUE;
+    } else {
+        empty_hash = OPENSSL_zalloc(EVP_MAX_MD_SIZE);
```
mamckee, line 532: >empty_hash = OPENSSL_zalloc(EVP_MAX_MD_SIZE); [](http://example.com/codeflow?start=8&length=45)

Known size. This can be declared as an array on the stack.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
mamckee: Oh, I see that these are passed into `p_scossl_tls13_hkdf_expand` by the context... I don't really like the idea of having to pass all of these as parameters to `p_scossl_tls13_hkdf_expand`, but I'm worried about bugs since we modify and reset the state of the context.

Maybe have a local `SCOSSL_HKDF_CTX tmpCtx = *ctx;` that you modify and pass to `p_scossl_tls13_hkdf_expand`. Looking at the OpenSSL function, nothing in this function should affect the state of `ctx`

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+            scError = SCOSSL_FAILURE;
+            goto cleanup;
+        }
+        EVP_MD_CTX_free(mctx);
+
+        ctx->pbData = empty_hash;
+        ctx->cbData = mdlen;
+        data_need_reset = TRUE;
+
+        expanded_secret = OPENSSL_zalloc(EVP_MAX_MD_SIZE);
```
mamckee, line 551: >expanded_secret = OPENSSL_zalloc(EVP_MAX_MD_SIZE); [](http://example.com/codeflow?start=8&length=50)

This should also be declared as an array on the stack.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+        expanded_secret = NULL;
+    }
+
+    return scError;
+}
+
+
+SCOSSL_STATUS p_scossl_tls13kdf_derive(_In_ SCOSSL_PROV_HKDF_CTX *ctx,
+                                _Out_writes_bytes_(keylen) unsigned char *key, size_t keylen,
+                                _In_ const OSSL_PARAM params[])
```
mamckee, line 628: Nit: indentation

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_KEY_LENGTH);
+        return 0;
+    }
+
+
+    switch (ctx->hkdfCtx->mode)
+    {
+    case EVP_KDF_HKDF_MODE_EXTRACT_ONLY:
+        scError = p_scossl_tls13kdf_generate_secret(ctx->hkdfCtx, key, keylen);
```
mamckee, line 653: You can just `return p_scossl_tls13kdf_generate_secret(...` in this block and the next.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+        SIZE_T cbPrefix = 0;
+
+        if (p->data_size > 0 &&
+            !OSSL_PARAM_get_octet_string(p, (void **)&pbPrefix, 0, &cbPrefix))
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_GET_PARAMETER);
+            return SCOSSL_FAILURE;
+        }
+
+        OPENSSL_clear_free(ctx->hkdfCtx->pbPrefix, ctx->hkdfCtx->cbPrefix);
```
mamckee, line 715: >OPENSSL_clear_free(ctx->hkdfCtx->pbPrefix, ctx->hkdfCtx->cbPrefix); [](http://example.com/codeflow?start=8&length=67)

Move these frees to the top of the block and set to NULL. then you can call `OSSL_PARAM_get_octet_string(p, (void **)&ctx->hkdfCtx->pbPrefix, 0, &ctx->hkdfCtx->cbPrefix)` and avoid these temporary variables. That also matches the OpenSSL behavior (where these are cleared even in the case `OSSL_PARAM_get_octet_string` fails)

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+        SIZE_T cbData = 0;
+
+        if (p->data_size > 0 &&
+            !OSSL_PARAM_get_octet_string(p, (void **)&pbData, 0, &cbData))
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_GET_PARAMETER);
+            return SCOSSL_FAILURE;
+        }
+
+        OPENSSL_clear_free(ctx->hkdfCtx->pbData, ctx->hkdfCtx->cbData);
```
mamckee, line 749: OpenSSL clears and resets this to NULL outside of this block: [https://github.com/openssl/openssl/blob/a198caa5e34a2003f83b5cd4419f864da55eb0a5/providers/implementations/kdfs/hkdf.c#L927-L928](https://github.com/openssl/openssl/blob/a198caa5e34a2003f83b5cd4419f864da55eb0a5/providers/implementations/kdfs/hkdf.c#L927-L928).

That feels like the wrong thing to do to me and may actually be a bug on OpenSSL so I think you're doing the right thing here...

SymCryptProvider/src/kdf/p_scossl_hkdf.h
```diff
@@ -32,6 +32,16 @@ const OSSL_PARAM *p_scossl_hkdf_settable_ctx_params(ossl_unused void *ctx, ossl_
...
+//TLS1.3 KDF functions
+SCOSSL_STATUS p_scossl_tls13kdf_derive(_In_ SCOSSL_PROV_HKDF_CTX *ctx,
+    _Out_writes_bytes_(keylen) unsigned char *key, size_t keylen,
+    _In_ const OSSL_PARAM params[]);
+
+const OSSL_PARAM *p_scossl_tls13kdf_gettable_ctx_params(ossl_unused void *ctx, ossl_unused void *provctx);
+const OSSL_PARAM *p_scossl_tls13kdf_settable_ctx_params(ossl_unused void *ctx, ossl_unused void *provctx);
+SCOSSL_STATUS p_scossl_tls13kdf_get_ctx_params(_In_ SCOSSL_PROV_HKDF_CTX *ctx, _Inout_ OSSL_PARAM params[]);
+SCOSSL_STATUS p_scossl_tls13kdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, const _In_ OSSL_PARAM params[]);
+
```
mamckee, line 44: We don't need to expose these functions. See my other commend, but we only expose these for HKDF since that needs to also be exposed through the key exchange interface for legacy compat.

ScosslCommon/src/scossl_hkdf.c
```diff
@@ -67,6 +104,9 @@ SCOSSL_STATUS scossl_hkdf_reset(SCOSSL_HKDF_CTX *ctx)
 {
     OPENSSL_clear_free(ctx->pbSalt, ctx->cbSalt);
     OPENSSL_clear_free(ctx->pbKey, ctx->cbKey);
+    OPENSSL_clear_free(ctx->pbPrefix, ctx->cbPrefix);
+    OPENSSL_clear_free(ctx->pbLabel, ctx->cbLabel);
```
mamckee, line 108: I don't think prefix and label are secret so you don't need to clear free. Also we should just refactor `scossl_hkdf_freectx` to call `scossl_hkdf_reset`

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+    }
+
+    symcryptHmacAlg = scossl_get_symcrypt_hmac_algorithm(EVP_MD_type(ctx->md));
+    if (symcryptHmacAlg == NULL) {
+        return SCOSSL_FAILURE;
+    }
+
+    SIZE_T labelLen = ctx->cbPrefix + ctx->cbLabel;
+    if (labelLen > 255)
+        return SCOSSL_FAILURE;
```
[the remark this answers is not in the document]
MS-megliu: I used SCOSSL_LOG_ERROR. if you prefer ERR_raise, plese let me know and I will make changes.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
MS-megliu: In OpenSSL, default_zeros is defined as static const. I needed to pass it in as pbKey, but it didn’t seem to work with a stack-allocated version—it caused memory issues. So I’m switching to a heap-allocated buffer instead.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
MS-megliu: I increased hkdflabellen every step, probably hkdflabel + hkdflabellen pointer format would be better?

ScosslCommon/src/scossl_hkdf.c
MS-megliu: do we need to clear and free the memory? label and prefix are still part of the structure.  refactor scossl_hkdf_freectx to call scossl_hkdf_reset done.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -353,6 +353,418 @@ const OSSL_DISPATCH p_scossl_hkdf_kdf_functions[] = {
...
+    }
+    
+    if (!p_scossl_tls13kdf_set_ctx_params(ctx, params))
+    {
+        return SCOSSL_FAILURE;
+    }
+    if (keylen == 0)
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_KEY_LENGTH);
+        return 0;
```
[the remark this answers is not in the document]
MS-megliu: forgot this one. will fix it after you review this version.

mamckee:
> Result = KDF_DERIVE_ERROR

This should fail with KDF_CTRL_ERROR

(see [https://github.com/microsoft/SymCrypt-OpenSSL/pull/118/files#r2114832877](https://github.com/microsoft/SymCrypt-OpenSSL/pull/118/files#r2114832877))

---
Refers to: EvpTestRecipes/3.0/evpkdf_tls13_kdf.txt:4936 in 8581172. [](commit_id = 858117267e7fb789edd57365d8b0a3158654f3d4, deletion_comment = False)

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +330,367 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+    SYMCRYPT_HKDF_EXPANDED_KEY scExpandedKey;
+    SIZE_T labelLen = 0;
+    SIZE_T totalLen = 0;
+        
+    BYTE hkdflabel[HKDF_MAXBUF];
+    SIZE_T hkdflabellen = 0;
+    
+    if (ctx->md == NULL) 
+    {
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_HKDF_DERIVE, ERR_R_INTERNAL_ERROR, "Missing Digest");
```
mamckee, line 434: This should just be `ERR_raise(ERR_LIB_PROV, PROV_R_MISSING_MESSAGE_DIGEST)` instead of an internal error. The caller neglected to set a digest before calling this function..

SymCryptProvider/src/kdf/p_scossl_hkdf.c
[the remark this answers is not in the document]
mamckee: I think that's fine. Maybe use `PROV_R_LENGTH_TOO_LARGE` instead of `ERR_R_INTERNAL_ERROR`.

Also use`SCOSSL_PROV_LOG_ERROR` instead of `SCOSSL_LOG_ERROR`. `SCOSSL_LOG_ERROR` takes a function name argument that isn't supported on OpenSSL 3 and raises the error under the wrong category.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +330,367 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+    {
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_HKDF_DERIVE, ERR_R_INTERNAL_ERROR, "Label+Prefix length exceeds limit");
+        return SCOSSL_FAILURE;
+    }
+    
+    // 2 bytes for output length, 1 byte for label length, and 1 byte for context length
+    totalLen = 2 + 1 + labelLen + 1 + ctx->cbData;
+    if (totalLen > HKDF_MAXBUF) 
+    {
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_HKDF_DERIVE, ERR_R_INTERNAL_ERROR,
```
mamckee, line 464: >ERR_R_INTERNAL_ERROR [](http://example.com/codeflow?start=51&length=20)

Should probably be PROV_R_LENGTH_TOO_LARGE here too

SymCryptProvider/src/kdf/p_scossl_hkdf.c
mamckee: Either is fine but we should just be consistent in this function. `hkdflabel` is declared as an array so that just seemed more idiomatic to me.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +330,367 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+    }
+
+    // Expand PRK
+    scError = SymCryptHkdfPrkExpandKey(
+        &scExpandedKey,
+        symcryptHmacAlg,
+        ctx->pbKey, ctx->cbKey);
+    if (scError != SYMCRYPT_NO_ERROR)
+    {
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_HKDF_DERIVE, ERR_R_INTERNAL_ERROR, "SymCrypt expand failed");
```
mamckee, line 504: >SCOSSL_LOG_ERROR(SCOSSL_ERR_F_HKDF_DERIVE, ERR_R_INTERNAL_ERROR, "SymCrypt expand failed"); [](http://example.com/codeflow?start=8&length=91)

Use `SCOSSL_PROV_LOG_SYMCRYPT_ERROR` for SymCrypt failures

```suggestion
SCOSSL_PROV_LOG_SYMCRYPT_ERROR("SymCryptHkdfPrkExpandKey failed, scError);
```

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +330,367 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+        return SCOSSL_FAILURE;
+    }
+    scError = SymCryptHkdfDerive(
+        &scExpandedKey,
+        hkdflabel, hkdflabellen,
+        key, keylen);
+    if (scError != SYMCRYPT_NO_ERROR)
+    {
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_HKDF_DERIVE, ERR_R_INTERNAL_ERROR, "SymCrypt derive failed");
```
mamckee, line 513: >SCOSSL_LOG_ERROR(SCOSSL_ERR_F_HKDF_DERIVE, ERR_R_INTERNAL_ERROR, "SymCrypt derive failed"); [](http://example.com/codeflow?start=8&length=91)

Use `SCOSSL_PROV_LOG_SYMCRYPT_ERROR`

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +330,367 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+    
+    symcryptHmacAlg = scossl_get_symcrypt_hmac_algorithm(EVP_MD_type(ctx->md));
+    if (symcryptHmacAlg == NULL) 
+    {
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_HKDF_DERIVE, ERR_R_INTERNAL_ERROR, "Invalid HMAC algorithm");
+        scError = SCOSSL_FAILURE;
+        goto cleanup;
+    }
+ 
+    SIZE_T mdlen = EVP_MD_get_size(ctx->md);
```
mamckee, line 546: >SIZE_T mdlen [](http://example.com/codeflow?start=4&length=12)

Nit: move declaration to the top

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +330,367 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+    if (mdlen <= 0)
+    {
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_HKDF_DERIVE, ERR_R_INTERNAL_ERROR, "Invalid digest size");
+        return SCOSSL_FAILURE;
+    }
+
+    // duplicate a ctx to use as pass-in parameter for Symcrypt
+    dupCtx = OPENSSL_zalloc(sizeof(*ctx));
+    if (!dupCtx)
```
mamckee, line 555: >dupCtx [](http://example.com/codeflow?start=9&length=6)

Nit: 1dupCtx == NULL1

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +330,367 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+ 
+    SIZE_T mdlen = EVP_MD_get_size(ctx->md);
+    if (mdlen <= 0)
+    {
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_HKDF_DERIVE, ERR_R_INTERNAL_ERROR, "Invalid digest size");
+        return SCOSSL_FAILURE;
+    }
+
+    // duplicate a ctx to use as pass-in parameter for Symcrypt
+    dupCtx = OPENSSL_zalloc(sizeof(*ctx));
```
mamckee, line 554: >dupCtx = OPENSSL_zalloc(sizeof(*ctx)); [](http://example.com/codeflow?start=4&length=38)

You can use `OPENSSL_memdup` instead off alloc/copy

ScosslCommon/src/scossl_hkdf.c
```diff
@@ -3,7 +3,6 @@
 //
 
 #include "scossl_hkdf.h"
```
mamckee, line 5: Nit: Removed a line here that doesn't need to be removed.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +330,367 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_KEY_LENGTH);
+        return 0;
+    }
+
+    switch (ctx->hkdfCtx->mode)
+    {
+    case EVP_KDF_HKDF_MODE_EXTRACT_ONLY:
+        return p_scossl_tls13kdf_generate_secret(ctx->hkdfCtx, key, keylen);
+        break;
```
mamckee, line 681: >break; [](http://example.com/codeflow?start=8&length=6)

These breaks are unreachable.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +330,367 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+
+cleanup:
+    //restore original key
+    if (key_need_reset == TRUE) {
+        ctx->pbKey = NULL;
+        ctx->cbKey = 0;
+    }
+    OPENSSL_free(dupCtx);
+    OPENSSL_free(default_zeros);
+    if (mdctx != NULL)
```
mamckee, line 647: >if (mdctx != NULL) [](http://example.com/codeflow?start=4&length=18)

You don't need this check

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +330,367 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+
+    if (dupCtx->pbSalt == NULL) 
+    {
+        dupCtx->pbSalt = default_zeros;
+        dupCtx->cbSalt = mdlen;
+    } else {
+        mdctx = EVP_MD_CTX_new();
+        if (mdctx == NULL ||
+            EVP_DigestInit_ex(mdctx, dupCtx->md, NULL) <= 0 ||
+            EVP_DigestFinal_ex(mdctx, empty_hash, NULL) <= 0)
```
mamckee, line 586: Nit: These functions are boolean returns so it would be more idiomatic to just do `if (!EVP_digest...`

MS-megliu: EVP_DigestInit_ex() and EVP_DigestFinal_ex() has int return value.

https://docs.openssl.org/3.0/man3/EVP_DigestInit/#synopsis

ScosslCommon/src/scossl_hkdf.c
MS-megliu: done

SymCryptProvider/src/kdf/p_scossl_hkdf.c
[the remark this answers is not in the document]
MS-megliu: changed most of the log to ERR_raise.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +330,367 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+    if (key_need_reset == TRUE) {
+        ctx->pbKey = NULL;
+        ctx->cbKey = 0;
+    }
+    OPENSSL_free(dupCtx);
+    OPENSSL_free(default_zeros);
+    if (mdctx != NULL)
+        EVP_MD_CTX_free(mdctx);
+    
+    return scError;
```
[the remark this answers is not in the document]
MS-megliu: Thanks for catching it!

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +330,367 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+        break;
+    case EVP_KDF_HKDF_MODE_EXPAND_ONLY:
+        return p_scossl_tls13_hkdf_expand(ctx->hkdfCtx, key, keylen);
+        break;
+    default:
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_HKDF_DERIVE, ERR_R_INTERNAL_ERROR,
+            "Invalid Mode: %d", ctx->hkdfCtx->mode);
+        return SCOSSL_FAILURE;
+    }
+    return SCOSSL_SUCCESS;
```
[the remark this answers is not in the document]
MS-megliu: I would keep the breaks and end-of-everything return. otherwise it is kind of confusing to see return of failure?

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +330,367 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+    {
+        SCOSSL_LOG_ERROR(SCOSSL_ERR_F_HKDF_DERIVE, ERR_R_INTERNAL_ERROR, "Failed to allocate default_zeros");
+        return SCOSSL_FAILURE;
+    }
+   
+    if (dupCtx->pbKey == NULL) 
+    {
+        dupCtx->pbKey = default_zeros;
+        dupCtx->cbKey = mdlen;
+        ctx->pbKey = default_zeros; //SymCryptHkdfExtractPrk uses original ctx, have to set pbKey as well
```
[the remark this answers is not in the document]
MS-megliu: dupCtx is a copy of ctx, so ctx->pbKey and ctx->cbKey would be the original copy. only when pbKey is NULL, ctx->pbKey will be modified. so reset it to NULL should be good.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
MS-megliu: ctx has BYTE info[HKDF_MAXBUF]; it would be big for stack. shall we use heap instead?

ScosslCommon/src/scossl_hkdf.c
mamckee: You still need to `OPENSSL_free` these. It's just that the wiping part of `OPENSSL_clear_free` wasn't needed.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
mamckee: Hmm yeah that's true. That's probably best then. You can use `OPENSSL_memdup` then instead of zalloc/mempy to save a line.

SymCryptProvider/src/kdf/p_scossl_hkdf.c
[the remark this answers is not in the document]
mamckee: Right but what I'm suggesting is initialize `PBYTE pbSavedKey = dupCtx->pbKey` and `SIZE_T cbSavedKey = dupCtx->cbKey`  right before calling `dupCtx->pbKey = dupCtx->pbSalt` and `dupCtx->cbKey = dupCtx->cbSalt` and restore with those instead of modifying the original context. Also then we only bother saving the original key when `dupCtx->pbSalt != NULL`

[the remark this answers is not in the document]
MS-megliu: Thanks for catching this bug!

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +331,357 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+        goto cleanup;
+    }
+    status = SCOSSL_SUCCESS;
+
+cleanup:
+    //restore original key
+    if (key_need_reset == TRUE) {
+        ctx->pbKey = NULL;
+        ctx->cbKey = 0;
+    }
```
mamckee, line 643: No longer needed

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +331,357 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+        dupCtx->pbKey = dupCtx->pbSalt;
+        dupCtx->cbKey = dupCtx->cbSalt;
+
+        if (SCOSSL_SUCCESS != p_scossl_tls13_hkdf_expand(dupCtx, expanded_secret, keylen)) 
+        {
+            goto cleanup;
+        }
+        //restore pbKey/cbKey
+        dupCtx->pbKey = ctx->pbKey;
+        dupCtx->cbKey = ctx->cbKey;
```
mamckee, line 608: Remove

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +331,357 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+    {
+        ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+        goto cleanup;
+    }
+   
+    if (dupCtx->pbKey == NULL) 
+    {
+        dupCtx->pbKey = default_zeros;
+        dupCtx->cbKey = mdlen;
+        key_need_reset = TRUE;
```
mamckee, line 575: No longer needed

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +331,357 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+{
+    SYMCRYPT_ERROR scError = SYMCRYPT_NO_ERROR;
+    SCOSSL_STATUS status = SCOSSL_FAILURE;
+    PCSYMCRYPT_MAC symcryptHmacAlg = NULL;
+    BYTE *default_zeros = NULL;
+    BYTE empty_hash[EVP_MAX_MD_SIZE]; 
+    BYTE expanded_secret[EVP_MAX_MD_SIZE];
+    EVP_MD_CTX *mdctx = NULL;
+    SCOSSL_HKDF_CTX *dupCtx;
+    BOOL key_need_reset =  FALSE;
```
mamckee, line 533: >BOOL key_need_reset =  FALSE; [](http://example.com/codeflow?start=4&length=29)

No longer needed

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +331,357 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+SCOSSL_STATUS p_scossl_tls13kdf_generate_secret(_In_ SCOSSL_HKDF_CTX *ctx, 
+                                                _Out_writes_bytes_(keylen) unsigned char *key, size_t keylen)
+{
+    SYMCRYPT_ERROR scError = SYMCRYPT_NO_ERROR;
+    SCOSSL_STATUS status = SCOSSL_FAILURE;
+    PCSYMCRYPT_MAC symcryptHmacAlg = NULL;
+    BYTE *default_zeros = NULL;
+    BYTE empty_hash[EVP_MAX_MD_SIZE]; 
+    BYTE expanded_secret[EVP_MAX_MD_SIZE];
+    EVP_MD_CTX *mdctx = NULL;
```
mamckee, line 531: >EVP_MD_CTX *mdctx = NULL; [](http://example.com/codeflow?start=4&length=25)

No longer needed

SymCryptProvider/src/kdf/p_scossl_hkdf.c
```diff
@@ -341,6 +331,357 @@ SCOSSL_STATUS p_scossl_hkdf_set_ctx_params(_Inout_ SCOSSL_PROV_HKDF_CTX *ctx, co
...
+        symcryptHmacAlg,
+        dupCtx->pbKey, dupCtx->cbKey,
+        dupCtx->pbSalt, dupCtx->cbSalt,
+        key, keylen);
+    if (scError != SYMCRYPT_NO_ERROR) 
+    {
+        SCOSSL_PROV_LOG_SYMCRYPT_ERROR("SymCryptHkdfExtractPrk failed", scError);
+        goto cleanup;
+    }
+    status = SCOSSL_SUCCESS;
```
mamckee, line 636: I think we could rework this section to avoid duplication and this goto.

You could remove the `SymCryptHkdfExtractPrk` call inside this block, and set `dupCtx->pbSalt = expanded_secret` and `dupCtx->cbKey = keylen` at the end of this block.

mamckee (approved):
LGTM

PR #120, opened 2025-05-21
RI fixes in 1.8.1 into main

mamckee:


CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.13.0)
 
 project(SymCrypt-OpenSSL
-    VERSION 1.8.0
+    VERSION 1.9.0
```
samuel-lee-msft, line 4: I wonder whether we can/should have an indicator that this is a 1.9.x dev version now we have specifically made a scossl-1.9 branch?

mamckee: Yeah, porbably a good idea. Although the main branch would be 1.10-dev now.

samuel-lee-msft (approved):

PR #125, opened 2025-05-28
Export KeysInUse as an API

mamckee:
This PR exports KeysInUse as a library available for use by applications.
- Moved KeysInUse code to it's own folder to separate it from the provider.
  - The KeysInUse functionality is still built into the SymCrypt provider by default.
  - Added a standalone KeysInUse library target for use on platforms without the SymCrypt provider.
- Made KeysInUse available as an API that is exported from the SymCrypt provider and standalone library.
- Reworked the internals of KeysInUse to maintain a hash table of loaded keys to improve thread safety for callers and minimize duplicate logging of the same key material.
  - The key for the hash table is the key identifier. This means that if an application loads the same key into two separate objects, KeysInUse will track and log their usage with the same internal KeysInUse context.
  - Removed the stack-based logging operations. The logging thread will now wake every hour and log each loaded key from the hash table. The logging thread will still be woken early to log first key uses.
  - Improved cleanup with `OPENSSL_atexit`.
  - Calling `keysinuse_init` is now optional, but can still be called as an optimization.
- Added an option to log to syslog instead of `/var/log/keysinuse`.
- Added a KeysInUse test, which tests the KeysUse API and KeysInUse logging for engines and providers that leverage the KeysInUse API.


mamckee:
Rebased to resolve merge conflicts

KeysInUse/keysinuse.c
```diff
@@ -0,0 +1,1147 @@
...
+
+        if (lh_keysinuse_ctx_imp == NULL)
+        {
+            keysinuse_log_error("Keysinuse context hash table is missing in keysinuse_load_key");
+            goto cleanup;
+        }
+    }
+    else
+    {
+        keysinuse_log_error("Failed to keysinuse context hash table for reading in keysinuse_load_key,OPENSSL_%d", ERR_get_error());
```
samuel-lee-msft, line 530: NP: Failed to lock keysinuse...

KeysInUse/keysinuse.c
```diff
@@ -0,0 +1,1147 @@
...
+            (ctx->lock = CRYPTO_THREAD_lock_new()) == NULL)
+        {
+            keysinuse_log_error("malloc failure in keysinuse_load_key,OPENSSL_%d", ERR_R_MALLOC_FAILURE);
+            goto cleanup;
+        }
+
+        memcpy(ctx->keyIdentifier, ctxTmpl.keyIdentifier, sizeof(ctxTmpl.keyIdentifier));
+        ctx->refCount = 1;
+
+        if (CRYPTO_THREAD_write_lock(lh_keysinuse_ctx_imp_lock))
```
samuel-lee-msft, line 547: >CRYPTO_THREAD_write_lock [](http://example.com/codeflow?start=12&length=24)

The current code can get to a race condition here where multiple threads are inserting the same context which they failed to read in parallel, and you end up with some callers using an orphaned SCOSSL_KEYSINUSE_CTX*.

It looks like with OpenSSL's `lh_*_insert`, you're expected to look at the return to see if you overwrote an entry?

I don't think you can just copy the data from the old entry and free it here, as the caller which inserted the old entry for a given key then maintains a pointer to that entry - we cannot update their reference to it from here.

Instead we could insert the old entry again, up-ref it, and free the entry we were going to use. This means in the normal case we just have a single insert per fresh key. Alternatively, we could `lh_SCOSSL_KEYSINUSE_CTX_IMP_retrieve` under the write lock before inserting, only inserting if there is no entry. This is slightly more costly in the normal case and less costly in the contended case.

Really we want `lh_SCOSSL_KEYSINUSE_CTX_IMP_try_insert` but it's not there in the OpenSSL API.

KeysInUse/keysinuse.c
```diff
@@ -0,0 +1,1147 @@
...
+    keysinuse_running = TRUE;
+    status = SCOSSL_SUCCESS;
+
+#ifdef KEYSINUSE_LOG_SYSLOG
+    keysinuse_log_notice("KeysInUse initialized");
+#endif
+
+cleanup:
+    if (status == SCOSSL_FAILURE)
+    {
```
MS-megliu, line 272: in failed case, do we need to CRYPTO_THREAD_lock_free(lh_keysinuse_ctx_imp_lock)?

KeysInUse/keysinuse.c
```diff
@@ -0,0 +1,1147 @@
...
+        return KEYSINUSE_KEYID_SIZE;
+    }
+
+    if (cbKeyIdentifier < KEYSINUSE_KEYID_SIZE)
+    {
+        keysinuse_log_error("Insufficient buffer size for key identifier, expected %d, got %lu", KEYSINUSE_KEYID_SIZE, cbKeyIdentifier);
+        return 0;
+    }
+
+    if (CRYPTO_THREAD_write_lock(ctxImp->lock))
```
samuel-lee-msft, line 634: >CRYPTO_THREAD_write_lock [](http://example.com/codeflow?start=8&length=24)

This could be a read lock, right?

KeysInUse/keysinuse.c
MS-megliu, line 530: failed to lock?

KeysInUse/keysinuse.c
```diff
@@ -0,0 +1,1147 @@
...
+        ctx = NULL;
+        goto cleanup;
+    }
+
+    status = SCOSSL_SUCCESS;
+
+cleanup:
+    if (status != SCOSSL_SUCCESS)
+    {
+        keysinuse_free_key_ctx(ctx);
```
samuel-lee-msft, line 584: >keysinuse_free_key_ctx(ctx); [](http://example.com/codeflow?start=8&length=28)

We never want to free a ctx we have retrieved from the hash table here, reasoning about this is fiddly when `ctx` can mean a locally allocated ctx we are trying to insert or a pointer to an entry we have retrieved. I think you have got it correct now where on failure we make ctx NULL before this unless we allocated ctx, but this seems fragile.

Suggest allocating into a different pointer (`localCtx`?) which aliases with `ctx` on successful insertion, and we can only free `localCtx` here.

KeysInUse/keysinuse.c
```diff
@@ -0,0 +1,1147 @@
...
+
+            if (lh_keysinuse_ctx_imp == NULL)
+            {
+                keysinuse_log_error("Keysinuse context hash table is missing in logging thread");
+                goto cleanup;
+            }
+        }
+        else
+        {
+            keysinuse_log_error("Logging thread failed to lock keysinuse context hash table for reading,OPENSSL_%d", ERR_get_error());
```
samuel-lee-msft, line 1131: >reading [](http://example.com/codeflow?start=96&length=7)

Should be updated to writing

KeysInUse/keysinuse.c
```diff
@@ -0,0 +1,1147 @@
...
+            goto cleanup;
+        }
+    }
+    while (isLoggingThreadRunning);
+
+cleanup:
+    is_logging = FALSE;
+    keysinuse_running = FALSE;
+    logging_thread_exit_status = SCOSSL_SUCCESS;
```
samuel-lee-msft, line 1140: >logging_thread_exit_status = SCOSSL_SUCCESS; [](http://example.com/codeflow?start=4&length=44)

This seems a little weird - does `logging_thread_exit_status` still have value in latest?

It looks like it should be indicating unexpected failure in the logging thread - i.e. any error+goto cleanup flow should keep the value as SCOSSL_FAILURE. Right now it seems to be effectively indicating whether the logging thread returned or not.

KeysInUse/keysinuse.c
mamckee: Yeah good catch. I'm inclined towards the first option here, since that contended case is unlikely. However, I worry about if the re-insert operation fails. I would need some way to invalidate the context so the old entry is freed in `keysinuse_unload_key`, since we failed to insert it into the hash table. This could potentially happen if the hash table was at capacity and expansion failed, since insert will always check the capacity and try to expand, even if the data is replaced ([see here](https://github.com/openssl/openssl/blob/b7861e55c2d1de50da7b746107c679a5038ba04d/crypto/lhash/lhash.c#L124C71-L124C75)). 

I could have some boolean `inMap` that we set to false in this case, but then there's a new race condition if the thread that created the original context calls `keysinuse_unload_key` before we can invalidate the context. I think for simplicity and safety we should just call `lh_SCOSSL_KEYSINUSE_CTX_IMP_retrieve` again. That way we can guarantee both threads have a valid reference and it will be less likely to introduce regressions.

KeysInUse/keysinuse.c
mamckee: Yeah

KeysInUse/keysinuse.c
mamckee: I agree, a small change here where ctx is not properly set to NULL on a retrieve and fail case would be a bug.

KeysInUse/keysinuse.c
```diff
@@ -0,0 +1,1147 @@
...
+            goto cleanup;
+        }
+
+        if (waitStatus != 0 && waitStatus != ETIMEDOUT)
+        {
+            keysinuse_log_error("Logging thread woken up with unexpected status, SYS_%d", waitStatus);
+            goto cleanup;
+        }
+
+        if (CRYPTO_THREAD_read_lock(lh_keysinuse_ctx_imp_lock))
```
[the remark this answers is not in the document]
mamckee: Yeah that's true. Holding the lock during this operation might have a small impact on callers, but it should be minimal.

KeysInUse/keysinuse.c
```diff
@@ -0,0 +1,1165 @@
...
+            {
+                keysinuse_log_error("Keysinuse context hash table is missing in keysinuse_load_key");
+            }
+
+            CRYPTO_THREAD_unlock(lh_keysinuse_ctx_imp_lock);
+
+            if (ctx == NULL)
+            {
+                goto cleanup;
+            }
```
samuel-lee-msft, line 588: NP: This and the following goto cleanups are redundant

Actually now I'm thinking about it, the whole function could get a lot less indented with some early error handling + goto cleanup. But I don't think it's a blocker by any means. The code is readable to me as is.

mamckee: Ok yeah I think I'll just clean this up when we decouple this from the SymCrypt provider. I don't want to block on current timelines and the compiler should optimize this out anyways.

PR #129, opened 2025-07-07
fix: compatibility issue between SCOSSL and TLS 1.2 ciphersuites which use HMAC in AzL3

MS-megliu:
Address bug 58142883
Error Message:
openssl s_client -connect tcs.microsoftstore.com.cn:443 -tls1_2 -cipher ECDHE-RSA-AES256-SHA384
40D79C8E857B0000:error:0A0003FC:SSL routines:ssl3_read_bytes:ssl/tls alert bad record mac:ssl/record/rec_layer_s3.c:907:SSL alert number 20

Root cause:
(1) TLS1.2 use 0 based padding, scossl is currently using 1 based, caused bad mac in encryption path
(2) copy_mac didn't copy the correct mac for aes get parameter to consume, caused bad mac in decryption path.

Test:
(1) sslplay passed

### DIFF
--- a/SymCryptProvider/src/ciphers/p_scossl_aes.c
+++ b/SymCryptProvider/src/ciphers/p_scossl_aes.c
@@ -155,19 +155,16 @@ static SCOSSL_STATUS p_scossl_aes_generic_decrypt_init(_Inout_ SCOSSL_AES_CTX *c
 // This function is adapted from ssl3_cbc_copy_mac in ssl/record/tls_pad.c
 // and runs in constant time. In case of bad padding, a random MAC is assigned instead
 static SCOSSL_STATUS p_scossl_aes_copy_mac(_Inout_ SCOSSL_AES_CTX *ctx,
-                                           _In_reads_bytes_(buflen) unsigned char *record, _Inout_ SIZE_T *recordLen,
+                                           _In_reads_bytes_(buflen) unsigned char *record,
+                                           _Inout_ SIZE_T *recordLen,
                                            SIZE_T recordLenPadded,
                                            SCOSSL_STATUS paddingStatus)
 {
-    // MAC rotation is performed in place
     BYTE rotatedMacBuf[64 + EVP_MAX_MD_SIZE];
     PBYTE rotatedMac;
-    BYTE aux1, aux2, aux3, mask;
-    BYTE paddingStatusByte = (BYTE) (paddingStatus & 0xff);
-    // Random mac set in case padding removal failed
     BYTE randMac[EVP_MAX_MD_SIZE];
 
-    UINT32 macEnd = *recordLen;
+    UINT32 macEnd = (UINT32)(*recordLen);  // mac end index
     UINT32 macStart = macEnd - ctx->tlsMacSize;
     UINT32 inMac = 0;
 
@@ -178,8 +175,6 @@ static SCOSSL_STATUS p_scossl_aes_copy_mac(_Inout_ SCOSSL_AES_CTX *ctx,
     OPENSSL_free(ctx->tlsMac);
     ctx->tlsMac = NULL;
 
-    // Public info, safe to branch
-    // No mac to copy
     if (ctx->tlsMacSize == 0)
     {
         return paddingStatus;
@@ -187,7 +182,6 @@ static SCOSSL_STATUS p_scossl_aes_copy_mac(_Inout_ SCOSSL_AES_CTX *ctx,
 
     *recordLen -= ctx->tlsMacSize;
 
-    // Generate random bytes in case of bad padding
     if (RAND_bytes_ex(ctx->libctx, randMac, ctx->tlsMacSize, 0) <= 0)
     {
         return SCOSSL_FAILURE;
@@ -201,13 +195,12 @@ static SCOSSL_STATUS p_scossl_aes_copy_mac(_Inout_ SCOSSL_AES_CTX *ctx,
 
     rotatedMac = rotatedMacBuf + ((0 - (SIZE_T)rotatedMacBuf) & 0x3f);
 
-    // Public info, safe to branch
     if (recordLenPadded > ctx->tlsMacSize + 255 + 1)
     {
         scanStart = recordLenPadded - (ctx->tlsMacSize + 255 + 1);
     }
 
-    // Find and extract MAC
+    // Extract MAC bytes in constant time
     memset(rotatedMac, 0, ctx->tlsMacSize);
     for (i = scanStart, j = 0; i < recordLenPadded; i++)
     {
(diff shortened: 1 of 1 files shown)


SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -268,11 +294,22 @@ static SCOSSL_STATUS p_scossl_aes_generic_block_update(_Inout_ SCOSSL_AES_CTX *c
...
         {
-            // in == out
-            SymCryptPaddingPkcs7Add(
-                SYMCRYPT_AES_BLOCK_SIZE,
-                in, inl,
-                out, outsize, &inl);
+            // TLS 1.1+ uses explicit IV and TLS-style (zero-based) padding
+            if (ctx->tlsVersion >= TLS1_1_VERSION)
+            {
+                if (!scossl_tls_add_padding(out, outsize, in, inl, &inl))
```
samuel-lee-msft, line 300: >if (!scossl_tls_add_padding(out, outsize, in, inl, &inl))  [](http://example.com/codeflow?start=16&length=58)

I believe this should apply even to TLS 1.0 - no TLS version uses bare PKCS7 padding.

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -221,27 +218,56 @@ static SCOSSL_STATUS p_scossl_aes_copy_mac(_Inout_ SCOSSL_AES_CTX *ctx,
...
+    *len -= totalPad;
     return SCOSSL_SUCCESS;
 }
 
+static int scossl_tls_add_padding(unsigned char *out, size_t outsize, const unsigned char *in, size_t inl, size_t *outlen)
+{
+    size_t blocksize = 16;
+    size_t padlen = blocksize - (inl % blocksize);
+    if (padlen == 0)
+        padlen = blocksize;
```
samuel-lee-msft, line 261: >    padlen = blocksize; [](http://example.com/codeflow?start=4&length=23)

NP: we prefer to always use enclose scopes in { } rather than having one-line if statements

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -221,27 +218,56 @@ static SCOSSL_STATUS p_scossl_aes_copy_mac(_Inout_ SCOSSL_AES_CTX *ctx,
...
+    for (size_t i = *len - totalPad; i < *len; i++) {
+        if (buf[i] != padlen)
+            return SCOSSL_FAILURE;
     }
 
-    // If we failed, we still succeed, but the MAC is set to some
-    // random value. It's up to the caller to check the MAC.
+    *len -= totalPad;
     return SCOSSL_SUCCESS;
 }
```
samuel-lee-msft, line 254: It is very important that this function is constant time w.r.t. the data in buf, we should do something very similar to SymCryptPaddingPkcs7Remove ([https://github.com/microsoft/SymCrypt/blob/769e0a33f76af987e0ab11e041202caf328ffe86/lib/paddingPkcs7.c#L61-L167](https://github.com/microsoft/SymCrypt/blob/769e0a33f76af987e0ab11e041202caf328ffe86/lib/paddingPkcs7.c#L61-L167)) here, accumulating errors into a variable with bitwise operations and having a data-invariant memory access pattern (always accessing the last full block).

We can afford to simplify to only support cbBlockSize = SYMCRYPT_AES_BLOCK_SIZE

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -221,27 +218,56 @@ static SCOSSL_STATUS p_scossl_aes_copy_mac(_Inout_ SCOSSL_AES_CTX *ctx,
         j &= SYMCRYPT_MASK32_LT(j, ctx->tlsMacSize);
     }
 
+    // Convert paddingStatus to full 0xFF mask or 0x00
+    BYTE mask = paddingStatus ? 0xFF : 0x00;
```
samuel-lee-msft, line 222: I think revert this to the previous
BYTE paddingStatusByte = (BYTE) (paddingStatus & 0xff);

ternary operations can be compiled as a branch - bitwise and will not

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -222,23 +220,74 @@ static SCOSSL_STATUS p_scossl_aes_copy_mac(_Inout_ SCOSSL_AES_CTX *ctx,
...
-        rotateOffset = (rotateOffset & SYMCRYPT_MASK32_LT(rotateOffset, ctx->tlsMacSize));
+    // Copy input to output
+    for (UINT32 i = 0; i < inl; i++)
+        out[i] = in[i];
+
+    // Write padding in constant time
+    for (UINT32 i = 0; i < SYMCRYPT_AES_BLOCK_SIZE; i++) {
+        UINT32 mask = (UINT32)(i < cbPadVal); // mask = 1 if i < padval, else 0
+        out[inl + i] = (BYTE)((out[inl + i] & ~mask) | ((cbPadVal - 1) & mask));
     }
```
samuel-lee-msft, line 288: On the encryption path it's actually fine to be non constant time - the caller of the encryption is the producer of the plaintext, so there is not an attack. You can use memcpy / memset as before.

The decryption path is sensitive to sidechannels because the ciphertext could be legitimate or could be tampered with over the wire.

samuel-lee-msft:
>                                              SymCryptMapUint32(scError, SCOSSL_FAILURE, scErrorMap, 1));

this should be something like:
```
paddingFailure = scossl_tls_remove_padding(out, outl);
return p_scossl_aes_copy_mac(ctx,
    out, outl,
    outlPadded,
    paddingFailure );
```

---
Refers to: SymCryptProvider/src/ciphers/p_scossl_aes.c:370 in 54e25da. [](commit_id = 54e25da8b67c6ef0af3da663addb36e906e1078e, deletion_comment = False)

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -222,23 +220,74 @@ static SCOSSL_STATUS p_scossl_aes_copy_mac(_Inout_ SCOSSL_AES_CTX *ctx,
...
+    for (UINT32 i = 0; i < SYMCRYPT_AES_BLOCK_SIZE; i++) {
+        UINT32 mask = (i >= SYMCRYPT_AES_BLOCK_SIZE - cbPadVal);
+        UINT32 diff = last_block[i] ^ cbPadVal;
+        mPaddingError |= (diff & -mask);
+    }
+
+    ret ^= mPaddingError & (ret ^ SYMCRYPT_INVALID_ARGUMENT);
+
+    *len -= cbPadVal;
+    return SCOSSL_SUCCESS;
```
samuel-lee-msft, line 267: >return SCOSSL_SUCCESS; [](http://example.com/codeflow?start=4&length=22)

return ret;

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -222,23 +220,74 @@ static SCOSSL_STATUS p_scossl_aes_copy_mac(_Inout_ SCOSSL_AES_CTX *ctx,
...
+        return SCOSSL_FAILURE;
+
+    cbPadVal = buf[*len - 1] + 1;
+    mPaddingError |= SymCryptMask32IsZeroU31(cbPadVal);
+
+    // If cbPadVal is greater than cbBlockSize,
+    // we have to limit cbPadVal to be at most equal to cbBlockSize.
+    cbPadVal = 1 + ((cbPadVal - 1) & (SYMCRYPT_AES_BLOCK_SIZE - 1));
+    // Copy last block to local buffer to ensure data-invariant access
+    start = *len - SYMCRYPT_AES_BLOCK_SIZE;
```
samuel-lee-msft, line 253: >*len - SYMCRYPT_AES_BLOCK_SIZE; [](http://example.com/codeflow?start=12&length=31)

off by on error - start should be
`(*len - 1) - SYMCRYPT_AES_BLOCK_SIZE`

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -222,23 +220,74 @@ static SCOSSL_STATUS p_scossl_aes_copy_mac(_Inout_ SCOSSL_AES_CTX *ctx,
...
+    for (UINT32 i = 0; i < SYMCRYPT_AES_BLOCK_SIZE; i++)
+        last_block[i] = buf[start + i];
```
samuel-lee-msft, line 255: we don't need to copy the data to a local last_block buffer - we can validate the padding directly in `buf` with the correct offsets

samuel-lee-msft:
>                                              SymCryptMapUint32(scError, SCOSSL_FAILURE, scErrorMap, 1));

actually, I'm thinking it probably makes sense to merge the padding removal into `p_scossl_aes_copy_mac` - something like `p_scossl_aes_tls_remove_padding_and_copy_mac`

---
In reply to: [3046740157](https://github.com/microsoft/SymCrypt-OpenSSL/pull/129#issuecomment-3046740157) [](http://example.com/codeflow?ancestors=3046740157)

---
Refers to: SymCryptProvider/src/ciphers/p_scossl_aes.c:370 in 54e25da. [](commit_id = 54e25da8b67c6ef0af3da663addb36e906e1078e, deletion_comment = False)

PR #130, opened 2025-07-08
Fix for TLS 1.x CBC ciphersuites

samuel-lee-msft:
+ Builds on https://github.com/microsoft/SymCrypt-OpenSSL/pull/129 but handles overly large padding on decryption

+ Addresses bug 58142883
Error Message:
openssl s_client -connect tcs.microsoftstore.com.cn:443 -tls1_2 -cipher ECDHE-RSA-AES256-SHA384
40D79C8E857B0000:error:0A0003FC:SSL routines:ssl3_read_bytes:ssl/tls alert bad record mac:ssl/record/rec_layer_s3.c:907:SSL alert number 20

Root cause:
(1) TLS1.2 uses PKCS7-like padding for CBC cipher modes, but it is not PKCS7. SCOSSL was using PKCS7 padding incorrectly in TLS CBC paths, both causing errors on encryption path (other party would expect padding to be one byte longer and normally fail in padding check, if not in subsequent MAC check), and the decryption path (one byte of other party's padding would be interpreted as the last byte of the MAC)
(2) copy_mac didn't copy the correct mac for aes get parameter to consume, caused bad mac in decryption path.

Test:
(1) sslplay passed

### DIFF
--- a/SymCryptProvider/src/ciphers/p_scossl_aes.c
+++ b/SymCryptProvider/src/ciphers/p_scossl_aes.c
@@ -270,6 +270,7 @@ static SCOSSL_STATUS p_scossl_aes_tls_add_padding(const unsigned char *in, size_
 
     if (inl + cbPad > outsize)
     {
+        ERR_raise(ERR_LIB_PROV, PROV_R_OUTPUT_BUFFER_TOO_SMALL);
         return SCOSSL_FAILURE; // Buffer too small
     }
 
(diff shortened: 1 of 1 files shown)


SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -150,43 +150,49 @@ static SCOSSL_STATUS p_scossl_aes_generic_decrypt_init(_Inout_ SCOSSL_AES_CTX *c
 
 #define SYMCRYPT_OPENSSL_MASK8_SELECT( _mask, _a, _b ) (SYMCRYPT_FORCE_READ8(&_mask) & _a) | (~(SYMCRYPT_FORCE_READ8(&_mask)) & _b)
 
-// Extracts the MAC from the end of out and saves the result to ctx->tlsMac
+// Verifies the TLS padding from the end of record, extracts the MAC from the end of
+// the unpadded record, and saves the result to ctx->tlsMac.
+//
 // The mac will later be fetched through p_scossl_aes_generic_get_ctx_params
```
mlindgren, line 156: np: `MAC`

mlindgren (approved):
LGTM

MS-megliu (approved):
LGTM

PR #134, opened 2025-07-11
Cherrypick TLS 1.x CBC ciphersuites fix on SCOSSL 1.9

samuel-lee-msft:
Fix for TLS 1.x CBC ciphersuites (#130)

Builds on fix: compatibility issue between SCOSSL and TLS 1.2 ciphersuites which use HMAC in AzL3 #129 but handles overly large padding on decryption

Addresses bug 58142883
Error Message:
openssl s_client -connect tcs.microsoftstore.com.cn:443 -tls1_2 -cipher ECDHE-RSA-AES256-SHA384 40D79C8E857B0000:error:0A0003FC:SSL routines:ssl3_read_bytes:ssl/tls alert bad record mac:ssl/record/rec_layer_s3.c:907:SSL alert number 20

Root cause:
(1) TLS1.2 uses PKCS7-like padding for CBC cipher modes, but it is not PKCS7. SCOSSL was using PKCS7 padding incorrectly in TLS CBC paths, both causing errors on encryption path (other party would expect padding to be one byte longer and normally fail in padding check, if not in subsequent MAC check), and the decryption path (one byte of other party's padding would be interpreted as the last byte of the MAC) (2) copy_mac didn't copy the correct mac for aes get parameter to consume, caused bad mac in decryption path. (3) fix typo in 1.10 for public key export

Test:
(1) sslplay passed


MS-megliu (approved):
LGTM

PR #145, opened 2025-09-04
Additional fixes for block ciphers in TLS connections

mamckee:
This PR fixes a regression where encrypt-then-mac was failing with block ciphers provided by the symcrypt provider. When ETM is used, the mac size passed to the provider is 0. Instead of removing the padding and skipping mac extraction, the SymCrypt provider treated a zero-mac length as invalid and failed before removing padding.

This PR also fixes a corner case issue when the SymCrypt provider is used with TLSv1, where an empty block was passed for encryption, and the caller expected an encrypted block of padding bytes.

### DIFF
--- a/SymCryptProvider/src/ciphers/p_scossl_aes.c
+++ b/SymCryptProvider/src/ciphers/p_scossl_aes.c
@@ -199,12 +199,6 @@ static SCOSSL_STATUS p_scossl_aes_tls_remove_padding_and_copy_mac(
         return SCOSSL_FAILURE;
     }
 
-    if ((ctx->tlsMac = OPENSSL_malloc(ctx->tlsMacSize)) == NULL)
-    {
-        ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
-        return SCOSSL_FAILURE;
-    }
-
     // We only care about the tail of the input buffer, which we can index with UINT32 indices
     // The if() is safe as both cbData and u32 are public values.
     u32 = ctx->tlsMacSize + 255 + 1;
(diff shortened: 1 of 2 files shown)


samuel-lee-msft:
> // the unpadded record, and saves the result to ctx->tlsMac.

Worth documenting that this part is optional based on `ctx->tlsMacSize = 0` and Encrypt then MAC vs. MAC then Encrypt ciphersuites

---
Refers to: SymCryptProvider/src/ciphers/p_scossl_aes.c:154 in 1640b26. [](commit_id = 1640b264dc9649b9d9ccbca66e2ae6d8fca005bd, deletion_comment = False)

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -292,11 +296,6 @@ static SCOSSL_STATUS p_scossl_aes_generic_block_update(_Inout_ SCOSSL_AES_CTX *c
     SIZE_T cbInFullBlocks = 0;
     *outl = 0;
 
-    if (inl == 0)
-    {
-        return SCOSSL_SUCCESS;
-    }
```
samuel-lee-msft, line 298: This seems removal right, but are we sure this does not regress the non-TLS case?

Specifically it looks like there would be a change in behavior when calling incremental padded decryption with an empty buffer (before no-op with success, after no-op which may fail depending on value of `ctx->cbBuf`)

samuel-lee-msft (approved):
LGTM modulo comments

SymCryptProvider/src/ciphers/p_scossl_aes.c
mamckee: It looks like this matches the current OpenSSL behavior, and all the EVP tests pass, but I'll double check the behavior of the OpenSSL implementation to make sure we match.

samuel-lee-msft (approved):

PR #146, opened 2025-09-17
Support forking processes for KeysInUse

mamckee:
KeysInUse relies on the logging thread to periodically log events. The logging thread is created when the SymCrypt provider is initialized, and runs for the lifetime of the provider. If the process forks, only the main thread and all global data are copied. Any additional threads, including the main thread, are not cloned in the child process. This means that only the parent process will emit KeysInUse events, and child processes will not.

This PR adds a new function to keysinuse `p_scossl_keysinuse_atfork_reinit`. This function is registered to run in the child process after `fork` with [`pthread_atfork`](https://man7.org/linux/man-pages/man3/pthread_atfork.3.html). The function will recreate the logging thread and reinitialize global state in the child process to enable continued keysinuse logging.

This change was verified with a test application that creates a private key, forks child processes, and uses the cloned key object in the child processes to ensure keys loaded in the parent process continue to emit keysinuse telemetry in child processes. I also verified this with nginx and confirmed keysinuse events were correctly written after the change.

- Added `p_scossl_keysinuse_atfork_reinit` callback to recreate the logging thread in the child process after `fork`
- Moved the `sk_keysinuse_info_pending` pointer to be global instead of local to the logging thread
    - The underlying stack object is allocated on the heap and therefore gets cloned after `fork`. It needs to be properly cleaned up in the child process to avoid duplicate events and prevent a memory leak.

### DIFF
--- a/SymCryptProvider/src/p_scossl_keysinuse.c
+++ b/SymCryptProvider/src/p_scossl_keysinuse.c
@@ -241,6 +241,8 @@ static void p_scossl_keysinuse_atfork_reinit()
     is_logging = FALSE;
     logging_thread_exit_status = SCOSSL_FAILURE;
 
+    logging_thread_mutex = (pthread_mutex_t) PTHREAD_MUTEX_INITIALIZER;
+
     // Recreate global locks in case they were held by the logging
     // thread in the parent process at the time of the fork.
     CRYPTO_THREAD_lock_free(sk_keysinuse_info_lock);
@@ -250,8 +252,6 @@ static void p_scossl_keysinuse_atfork_reinit()
         p_scossl_keysinuse_log_error("Failed to create keysinuse lock in child process");
         goto cleanup;
     }
-
-    logging_thread_mutex = (pthread_mutex_t) PTHREAD_MUTEX_INITIALIZER;
 
     // If any keysinuseInfo were in either stack, they will
     // be logged by the parent process. Remove them from the child process's


SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -210,6 +224,94 @@ void p_scossl_keysinuse_init()
...
+    {
+        pKeysinuseInfo = sk_SCOSSL_PROV_KEYSINUSE_INFO_pop(sk_keysinuse_info);
+        if (pKeysinuseInfo != NULL)
+        {
+            pKeysinuseInfo->logPending = FALSE;
+            pKeysinuseInfo->decryptCounter = 0;
+            pKeysinuseInfo->signCounter = 0;
+
+            p_scossl_keysinuse_info_free(pKeysinuseInfo);
+        }
```
samuel-lee-msft, line 264: Seems like this should be guarded by `CRYPTO_THREAD_write_lock(pKeysinuseInfo->lock)` to prevent getting into an inconsistent state if the child process uses the pKeysinuseInfo in another thread while this function executes? (i.e. `p_scossl_keysinuse_add_use`)

Same for `sk_keysinuse_info_pending` below

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -210,6 +224,94 @@ void p_scossl_keysinuse_init()
...
+
+    // Only recreate logging thread if it was running in the parent process
+    if (is_parent_logging)
+    {
+        sk_keysinuse_info_lock = CRYPTO_THREAD_lock_new();
+        if (sk_keysinuse_info_lock == NULL)
+        {
+            p_scossl_keysinuse_log_error("Failed to create keysinuse lock in child process");
+            goto cleanup;
+        }
```
samuel-lee-msft, line 288: I wonder whether you re-initialize `sk_keysinuse_info_lock` directly with CRYPTO_THREAD_lock_new rather than have a window where it's NULL.

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: We'd only have to worry about other threads in the child process if another handler is registered with `pthread_atfork` so it's a low likelihood but probably best to just be safe then. 

Did you mean lock `sk_keysinuse_info`? `sk_keysinuse_info_pending` is only used in the logging thread and has no associated lock. It has to be managed outside of the logging thread like `sk_keysinuse_info` because it's allocated on the heap, and we lose our pointer if we manage this in the logging thread.

samuel-lee-msft: This comment is for the per-info locks: `pKeysinuseInfo->lock`

I'm not entirely sure how `pthread_atfork` works in practice, but my impression from reading https://linux.die.net/man/3/pthread_atfork was that the threads in the child process may execute stuff before the child fork handler completes (the documentation talks about taking all locks in the prepare fork handler and then unlocking in the after fork handlers).

samuel-lee-msft (approved):
LGTM

PR #160, opened 2026-02-19
Clear HKDF info when set from params

mamckee:
HKDF info can be passed piecemeal to `p_scossl_hkdf_set_ctx_params`. The provider is responsible for concatenating all `OSSL_KDF_PARAM_INFO` parameters to construct the final info used for HKDF. The SymCrypt provider currently allows concatenation between subsequent `p_scossl_hkdf_set_ctx_params` calls, however, OpenSSL default implementation clears the stored info each time.

This PR brings the SymCrypt provider's behavior in line with the default OpenSSL implementation. This was discovered when testing the SymCrypt provider with OpenSSL 3.5.


MS-megliu (approved):
LGTM

PR #161, opened 2026-02-25
Add Pure ML-DSA to the SymCrypt provider

mamckee:
This PR adds pure ML-DSA to the SymCrypt provider through the provider signature interface. This PR was tested with the EVP crypto API, as well as TLS and CMS APIS.

This PR also adds an ML-DSA encoder/decoder for testing on platforms with OpenSSL versions that don't support ML-DSA by default (e.g. AZL3 with OpenSSL 3.3).

Composite ML-DSA and external mu will be added later.


SymCryptProvider/src/keymgmt/p_scossl_mldsa_keymgmt.c
```diff
@@ -0,0 +1,909 @@
...
+            }
+
+            if (p_scossl_mldsa_keymgmt_get_encoded_key(keyCtx1, format, &pbKey1, &cbKey1) != SCOSSL_SUCCESS ||
+                p_scossl_mldsa_keymgmt_get_encoded_key(keyCtx2, format, &pbKey2, &cbKey2) != SCOSSL_SUCCESS)
+            {
+                goto cleanup;
+            }
+
+            if (cbKey1 != cbKey2 ||
+                memcmp(pbKey1, pbKey2, cbKey1) != 0)
```
jasjivsingh-corp, line 491: is this safe from side-channel attacks? #Resolved

SymCryptProvider/src/keymgmt/p_scossl_mldsa_keymgmt.c
```diff
@@ -0,0 +1,909 @@
...
+    return copyCtx;
+}
+
+static SCOSSL_STATUS p_scossl_mldsa_keygen_set_params(_Inout_ SCOSSL_MLDSA_KEYGEN_CTX *genCtx, _In_ const OSSL_PARAM params[])
+{
+    PBYTE pbSeed = genCtx->abSeed;
+    const OSSL_PARAM *p;
+
+    if ((p = OSSL_PARAM_locate_const(params, OSSL_PKEY_PARAM_ML_DSA_SEED)) != NULL &&
+        !OSSL_PARAM_get_octet_string(p, (void **)&pbSeed, SCOSSL_MLDSA_PRIVATE_SEED_LENGTH, &genCtx->cbSeed))
```
jasjivsingh-corp: should there be a length check like it's done in mlkem? #Resolved

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+        return SCOSSL_SUCCESS;
+    }
+
+    if ((p = OSSL_PARAM_locate(params, OSSL_SIGNATURE_PARAM_ALGORITHM_ID)) != NULL)
+    {
+        BYTE abAid[SCOSSL_MLDSA_MAX_ALGORITHM_ID_SIZE];
+        PBYTE pbAid = abAid;
+        SIZE_T cbAid;
+
+        if (p_scossl_mldsa_get_alg_id(ctx->mldsaParams, &pbAid, &cbAid) != SCOSSL_SUCCESS)
```
jasjivsingh-corp, line 362: this will increment pbAid before it will used below? - https://linux.die.net/man/3/i2d_x509  #Resolved

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+{
+    SCOSSL_MLDSA_SIGNATURE_CTX *copyCtx = OPENSSL_malloc(sizeof(SCOSSL_MLDSA_SIGNATURE_CTX));
+    if (copyCtx == NULL)
+    {
+        ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+        return NULL;
+    }
+
+    copyCtx->mldsaParams = ctx->mldsaParams;
+    copyCtx->keyCtx = ctx->keyCtx;
```
jasjivsingh-corp, line 77: can we document if ref count is needed or not #Resolved

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+    {
+        ctx->mldsaParams = mldsaParams;
+    }
+
+    return ctx;
+}
+
+static void p_scossl_mldsa_freectx(SCOSSL_MLDSA_SIGNATURE_CTX *ctx)
+{
+    OPENSSL_free(ctx);
```
jasjivsingh-corp, line 64: ctx = null after this? #Resolved

SymCryptProvider/src/keymgmt/p_scossl_mldsa_keymgmt.c
mamckee: Yes, this function isn't expected to be constant time w.r.t. the key material. I should replace this memcmp with a CRYPTO_memcmp though.

SymCryptProvider/src/signature/p_scossl_mldsa.c
mamckee: That's just setting parameter to NULL, which doesn't exist after this function.

SymCryptProvider/src/signature/p_scossl_mldsa.c
mamckee: Historically we haven't needed to in SCOSSL, since freeing the key while it's still in use by another context is undefined behavior. Its assumed that the key is valid and will not be freed while this context is in use. OpenSSL does refcount these, but we haven't found a case yet where a caller depends on this context being valid after freeing the keyCtx.

Getting refcounting right can be tricky, and if we make a mistake, then we risk memory leaks. I would air on the side of not refcounting these keys (current behavior for other key types as well), but adding ref counts for all key types if we find a caller that expects this behavior.

SymCryptProvider/src/keymgmt/p_scossl_mldsa_keymgmt.c
mamckee: Yeah, good catch.

SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -338,7 +338,7 @@ static SCOSSL_STATUS p_scossl_aes_generic_block_update(_Inout_ SCOSSL_AES_CTX *c
             case DTLS1_BAD_VER:
                 out += SYMCRYPT_AES_BLOCK_SIZE;
                 *outl -= SYMCRYPT_AES_BLOCK_SIZE;
-                __attribute__ ((fallthrough));
+                /* fallthrough */
```
mamckee: Nit: move this to bugfixed PR

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+            return SCOSSL_FAILURE;
+        }
+    }
+
+    return SCOSSL_SUCCESS;
+}
+
+#define SCOSSL_MLDSA_FNS(bits)                                                                          \
+    static SCOSSL_MLDSA_SIGNATURE_CTX                                                                   \
+    *p_scossl_mldsa##bits##_newctx(ossl_unused SCOSSL_PROVCTX *provctx, ossl_unused const char *propq)  \
```
samuel-lee-msft: >##bits## [](http://example.com/codeflow?start=19&length=8)

NP: not really bits - 44, 65, and 87 refer to matrix dimensions.

A bit funky that we have 3 flavors of signing functions rather than just 3 flavors of ML-DSA key management functions, but I guess it is to match upstream OpenSSL behavior?

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+    {NID_undef, SCOSSL_OID_MLDSA65, SCOSSL_SN_MLDSA65, SCOSSL_LN_MLDSA65, SYMCRYPT_MLDSA_PARAMS_MLDSA65},
+    {NID_undef, SCOSSL_OID_MLDSA87, SCOSSL_SN_MLDSA87, SCOSSL_LN_MLDSA87, SYMCRYPT_MLDSA_PARAMS_MLDSA87}};
+
+typedef struct
+{
+    SYMCRYPT_MLDSA_PARAMS mldsaParams;
+    SCOSSL_MLDSA_KEY_CTX *keyCtx;
+    int operation;
+
+    BYTE pbContext[SYMCRYPT_MLDSA_CONTEXT_MAX_LENGTH];
```
samuel-lee-msft: >pbContext [](http://example.com/codeflow?start=9&length=9)

NP: This should be `abContext`, it is not a pointer

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+    const OSSL_PARAM *p;
+
+    if ((p = OSSL_PARAM_locate_const(params, OSSL_SIGNATURE_PARAM_CONTEXT_STRING)) != NULL)
+    {
+        PVOID pbContext = ctx->pbContext;
+        if (!OSSL_PARAM_get_octet_string(p, (void **)&pbContext, sizeof(ctx->pbContext), &ctx->cbContext))
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_GET_PARAMETER);
+            return SCOSSL_FAILURE;
+        }
```
samuel-lee-msft, line 280: I believe that if this call fails because the provided ctx string is too big, `ctx->cbContext` is still set to the too large value, and subsequent operations using `cbContext` (e.g. p_scossl_mldsa_dupctx) may read OOB of `ctx->abContext`.

samuel-lee-msft: Probably worth setting ctx->cbContext to 0 before this, and only setting it to the returned value on success?

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+    }
+
+    if ((p = OSSL_PARAM_locate_const(params, OSSL_SIGNATURE_PARAM_TEST_ENTROPY)) != NULL &&
+        p->data_size != 0)
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_NOT_SUPPORTED);
+        return SCOSSL_FAILURE;
+    }
+
+    if ((p = OSSL_PARAM_locate_const(params, OSSL_SIGNATURE_PARAM_MU)) != NULL)
```
samuel-lee-msft, line 323: >if ((p = OSSL_PARAM_locate_const(params, OSSL_SIGNATURE_PARAM_MU)) != NULL) [](http://example.com/codeflow?start=4&length=75)

We should ideally support external mu with SymCryptExternalMuMlDsaSign / SymCryptExternalMuMlDsaVerify

samuel-lee-msft: Can be a followup

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+    {
+        ctx->mldsaParams = mldsaParams;
+    }
+
+    return ctx;
+}
+
+static void p_scossl_mldsa_freectx(SCOSSL_MLDSA_SIGNATURE_CTX *ctx)
+{
+    OPENSSL_free(ctx);
```
samuel-lee-msft, line 64: >OPENSSL_free(ctx); [](http://example.com/codeflow?start=4&length=18)

In other OPENSSL_free paths the free is normally guarded by a NULL check.

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+    copyCtx->cbContext = ctx->cbContext;
+    memcpy(copyCtx->pbContext, ctx->pbContext, ctx->cbContext);
+
+    return copyCtx;
+}
+
+static SCOSSL_STATUS p_scossl_mldsa_signverify_init(_Inout_ SCOSSL_MLDSA_SIGNATURE_CTX *ctx, _In_ SCOSSL_MLDSA_KEY_CTX *keyCtx,
+                                                    _In_ const OSSL_PARAM params[], int operation)
+{
+    if (keyCtx == NULL || keyCtx->key == NULL)
```
samuel-lee-msft: In `p_scossl_mlkem_init` we also check `ctx` is not NULL

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+    }
+
+    return SCOSSL_SUCCESS;
+}
+
+static SCOSSL_STATUS p_scossl_mldsa_digest_signverify_init(_In_ SCOSSL_MLDSA_SIGNATURE_CTX *ctx, _In_ const char *mdname,
+                                                           _In_ SCOSSL_MLDSA_KEY_CTX *keyCtx, _In_ const OSSL_PARAM params[],
+                                                           int operation)
+{
+    if (mdname != NULL && mdname[0] != '\0')
```
samuel-lee-msft, line 233: >&& mdname[0] != '\0' [](http://example.com/codeflow?start=23&length=20)

NP: Technically the empty string is not allowed per the documentation: [https://docs.openssl.org/3.5/man7/EVP_SIGNATURE-ML-DSA/#notes](https://docs.openssl.org/3.5/man7/EVP_SIGNATURE-ML-DSA/#notes)

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_DIGEST);
+        return SCOSSL_FAILURE;
+    }
+
+    if (keyCtx == NULL && ctx->keyCtx != NULL)
```
samuel-lee-msft: >ctx->keyCtx [](http://example.com/codeflow?start=26&length=11)

May want to check for ctx != NULL before this

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+    {
+        return p_scossl_mldsa_set_ctx_params(ctx, params);
```
samuel-lee-msft, line 244: >return p_scossl_mldsa_set_ctx_params(ctx, params); [](http://example.com/codeflow?start=8&length=50)

This path leaves old operation - is that intentional?

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+    }
```
samuel-lee-msft, line 245: It's a bit weird to me that this is a special case for digestSign/digestVerify. I get there may a use case for reusing a ctx without changing the key, but I would expect that to be in the p_scossl_mldsa_signverify_init logic which also applies to plain normal sign/verify init.

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+
+    aobj = OBJ_txt2obj(oid, 1);
+    if (aobj == NULL)
+    {
+        SCOSSL_PROV_LOG_ERROR(ERR_R_INTERNAL_ERROR, "OBJ_txt2obj failed");
+        goto cleanup;
+    }
+
+    X509_ALGOR_set0(x509Alg, aobj, V_ASN1_UNDEF, NULL);
```
samuel-lee-msft: >X509_ALGOR_set0 [](http://example.com/codeflow?start=4&length=15)

Check this returns 1 for success?

SymCryptProvider/src/signature/p_scossl_mldsa.c
```diff
@@ -0,0 +1,523 @@
...
+    aobj = NULL; // X509_ALGOR_set0 takes ownership
+
+    if ((cbAid = i2d_X509_ALGOR(x509Alg, &pbAlgId)) < 0)
```
samuel-lee-msft, line 482: >&pbAlgId [](http://example.com/codeflow?start=41&length=8)

This seems funky to me; there is nothing about `i2d_X509_ALGOR` that fundamentally prevents it from writing more than `SCOSSL_MLDSA_MAX_ALGORITHM_ID_SIZE` bytes to `pbAlgId`

It almost seems better to just pass a pointer to NULL and let `i2d_X509_ALGOR` allocate, rather than using `abAid` in `p_scossl_mldsa_get_ctx_params` which leaves the potential for buffer overrun.

SymCryptProvider/src/signature/p_scossl_mldsa.c
mamckee: Yeah, default OpenSSL has a different keymgmt and signature dispatch for each ML-DSA flavor

SymCryptProvider/src/signature/p_scossl_mldsa.c
mamckee: You're right. I'll set this to 0 in the case of failure to just match the default OpenSSL behavior.

SymCryptProvider/src/signature/p_scossl_mldsa.c
mamckee: Yeah, I plan on adding external MU and ML-DSA composite in a follow up PR. Neither were finished when I implemented this and this PR is already large enough as is.

SymCryptProvider/src/signature/p_scossl_mldsa.c
mamckee: While this is safe I agree adding a NULL check makes sense for consistency, especially if the context adds members in the future that need to be cleared/freed

SymCryptProvider/src/signature/p_scossl_mldsa.c
mamckee: This is to match the default OpenSSL behavior. I agree that the correct interpretation of the docs is this must only be NULL, but it seems the intent is that the caller can't try and pass an mdname.

SymCryptProvider/src/signature/p_scossl_mldsa.c
mamckee: OpenSSL default doesn't, and I'm trying to keep the behavior 1:1, but I don't see the harm in adding this check.

SymCryptProvider/src/signature/p_scossl_mldsa.c
mamckee: This was to match the default provider, but it looks like the default provider always sets the operation type to signature...

SymCryptProvider/src/signature/p_scossl_mldsa.c
mamckee: I agree this is weird behavior. I'm trying to match the default OpenSSL behavior as much as makes sense. I don't think many callers will actually be using these functions for ML-DSA but I want to make sure the behavior is as close as possible to the upstream.

SymCryptProvider/src/keymgmt/p_scossl_mldsa_keymgmt.h
```diff
@@ -0,0 +1,26 @@
...
+
+SCOSSL_MLDSA_KEY_CTX *p_scossl_mldsa_keymgmt_new_ctx(_In_ SYMCRYPT_MLDSA_PARAMS mldsaParams);
+void p_scossl_mldsa_keymgmt_free_key_ctx(_In_ SCOSSL_MLDSA_KEY_CTX *keyCtx);
+
+SCOSSL_STATUS p_scossl_mldsa_keymgmt_import(_Inout_ SCOSSL_MLDSA_KEY_CTX *keyCtx, int selection, _In_ const OSSL_PARAM params[]);
+SCOSSL_STATUS p_scossl_mldsa_keymgmt_export(_In_ SCOSSL_MLDSA_KEY_CTX *keyCtx, int selection,
+                                            _In_ OSSL_CALLBACK *param_cb, _In_ void *cbarg);
+
+SCOSSL_STATUS p_scossl_mldsa_keymgmt_get_encoded_key(_In_ const SCOSSL_MLDSA_KEY_CTX *keyCtx, SYMCRYPT_MLDSAKEY_FORMAT format,
+                                                     _Out_writes_bytes_(*pcbKey) PBYTE *ppbKey, _Out_ SIZE_T *pcbKey);
```
samuel-lee-msft: >_Out_writes_bytes_(*pcbKey) PBYTE *ppbKey, _Out_ [](http://example.com/codeflow?start=53&length=48)

NP: _Inout_ for both of these; same as ML-KEM

SymCryptProvider/src/keymgmt/p_scossl_mldsa_keymgmt.c
```diff
@@ -0,0 +1,920 @@
...
+#include "p_scossl_mldsa_keymgmt.h"
+
+#include <openssl/param_build.h>
+#include <openssl/proverr.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define SCOSSL_MLDSA_PRIVATE_SEED_LENGTH 64
```
samuel-lee-msft: >SCOSSL_MLDSA_PRIVATE_SEED_LENGTH 64 [](http://example.com/codeflow?start=8&length=35)

The private seed length for ML-DSA is normally considered to be 32 bytes (it's exposed as SYMCRYPT_MLDSA_PRIVATE_SEED_SIZE in recent SymCrypt.h)

samuel-lee-msft: And yeah, that is the expectation of OpenSSL too: [https://docs.openssl.org/master/man7/EVP_PKEY-ML-DSA/#keygen-parameters](https://docs.openssl.org/master/man7/EVP_PKEY-ML-DSA/#keygen-parameters)

SymCryptProvider/src/keymgmt/p_scossl_mldsa_keymgmt.c
```diff
@@ -0,0 +1,920 @@
...
+int p_scossl_mldsa_get_bits(SYMCRYPT_MLDSA_PARAMS mldsaParams)
+{
+    switch (mldsaParams)
+    {
+        case SYMCRYPT_MLDSA_PARAMS_MLDSA44:
+            return 1312;
+        case SYMCRYPT_MLDSA_PARAMS_MLDSA65:
+            return 1952;
+        case SYMCRYPT_MLDSA_PARAMS_MLDSA87:
+            return 2592;
```
samuel-lee-msft: A couple of comments:
1) these are byte lengths, not bit lengths - you'll want to multiply by 8 I think
2) Consider using `SYMCRYPT_MLDSA_PUBLIC_KEY_SIZE_MLDSAxx` macros rather than magic values for the byte-lengths

SymCryptProvider/src/keymgmt/p_scossl_mldsa_keymgmt.c
```diff
@@ -0,0 +1,920 @@
...
+    }
+    *pcbKey = cbKey;
+
+    ret = SCOSSL_SUCCESS;
+
+cleanup:
+    if (ret != SCOSSL_SUCCESS && allocatedKey)
+    {
+        OPENSSL_secure_clear_free(pbKey, cbKey);
+    }
```
samuel-lee-msft, line 817: NP: This logic seems fine, but diverges from ML-KEM for not much reason, prefer both to have the same shape

SymCryptProvider/src/keymgmt/p_scossl_mldsa_keymgmt.c
```diff
@@ -0,0 +1,920 @@
...
+{
+    return p_scossl_mldsa_keygen_settable_param_types;
+}
+
+static void p_scossl_mldsa_keygen_cleanup(_Inout_ SCOSSL_MLDSA_KEYGEN_CTX *genCtx)
+{
+    OPENSSL_secure_clear_free(genCtx, sizeof(SCOSSL_MLDSA_KEYGEN_CTX));
+}
+
+static SCOSSL_MLDSA_KEYGEN_CTX *p_scossl_mldsa_keygen_init(_In_ const OSSL_PARAM params[], _In_ SYMCRYPT_MLDSA_PARAMS mldsaParams)
```
samuel-lee-msft: >_In_  [](http://example.com/codeflow?start=91&length=5)

NP: _In_ doesn't mean anything for an enum type which is always passed by value

SymCryptProvider/src/signature/p_scossl_mldsa.h
```diff
@@ -0,0 +1,36 @@
...
+    int nid;
+    const char *oid;
+    const char *snGroupName;
+    const char *lnGroupName;
+    SYMCRYPT_MLDSA_PARAMS mldsaParams;
+} SCOSSL_MLDSA_ALG_INFO;
+
+typedef struct
+{
+    OSSL_LIB_CTX *libctx;
```
samuel-lee-msft: I don't see that this is ever set or used?

SymCryptProvider/src/keymgmt/p_scossl_mldsa_keymgmt.c
```diff
@@ -0,0 +1,920 @@
...
+
+static const OSSL_PARAM p_scossl_mldsa_keymgmt_gettable_param_types[] = {
+    OSSL_PARAM_int(OSSL_PKEY_PARAM_BITS, NULL),
+    OSSL_PARAM_int(OSSL_PKEY_PARAM_SECURITY_BITS, NULL),
+    OSSL_PARAM_int(OSSL_PKEY_PARAM_MAX_SIZE, NULL),
+    OSSL_PARAM_utf8_string(OSSL_PKEY_PARAM_MANDATORY_DIGEST, NULL, 0),
+    OSSL_PARAM_octet_string(OSSL_PKEY_PARAM_ML_DSA_SEED, NULL, 0),
+    OSSL_PARAM_octet_string(OSSL_PKEY_PARAM_PRIV_KEY, NULL, 0),
+    OSSL_PARAM_octet_string(OSSL_PKEY_PARAM_PUB_KEY, NULL, 0),
+    OSSL_PARAM_octet_string(OSSL_PKEY_PARAM_ENCODED_PUBLIC_KEY, NULL, 0),
```
samuel-lee-msft, line 37: > OSSL_PARAM_octet_string(OSSL_PKEY_PARAM_ENCODED_PUBLIC_KEY, NULL, 0), [](http://example.com/codeflow?start=3&length=70)

Nit: upstream doesn't seem to support this for ML-DSA, should SCOSSL?

SymCryptProvider/src/keymgmt/p_scossl_mldsa_keymgmt.c
```diff
@@ -0,0 +1,920 @@
...
+        SIZE_T cbSignature;
+
+        scError = SymCryptMlDsaSizeofSignatureFromParams(keyCtx->mldsaParams, &cbSignature);
+        if (scError != SYMCRYPT_NO_ERROR)
+        {
+            SCOSSL_PROV_LOG_SYMCRYPT_ERROR("SymCryptMlDsaSizeofSignatureFromParams failed", scError);
+            return SCOSSL_FAILURE;
+        }
+
+        if (!OSSL_PARAM_set_int(p, cbSignature))
```
samuel-lee-msft: >int [](http://example.com/codeflow?start=28&length=3)

NP: size_t?

SymCryptProvider/src/keymgmt/p_scossl_mldsa_keymgmt.c
```diff
@@ -0,0 +1,920 @@
...
+                p_scossl_mldsa_keymgmt_get_encoded_key(keyCtx2, format, &pbKey2, &cbKey2) != SCOSSL_SUCCESS)
+            {
+                goto cleanup;
+            }
+
+            if (cbKey1 != cbKey2 ||
+                memcmp(pbKey1, pbKey2, cbKey1) != 0)
+            {
+                goto cleanup;
+            }
```
samuel-lee-msft, line 494: Tend to think it's safer to just compare public keys; we don't ever have a scenario where we have an initialized key and the public key is not available for comparison.

memcmp with the private keys could possibly be considered a sidechannel issue in some scenario.

SymCryptProvider/src/decoder/p_scossl_decode_mldsa.c
```diff
@@ -0,0 +1,184 @@
...
+        !PKCS8_pkey_get0(&algorithm, &pbKey, &cbKey, NULL, p8Info) ||
+        d2i_ASN1_OCTET_STRING(&p8Data, &pbKey, cbKey) == NULL)
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_BAD_ENCODING);
+        goto cleanup;
+    }
+
+    cbKey = ASN1_STRING_length(p8Data);
+
+    format = cbKey == 64 ? SYMCRYPT_MLDSAKEY_FORMAT_PRIVATE_SEED : SYMCRYPT_MLDSAKEY_FORMAT_PRIVATE_KEY;
```
samuel-lee-msft: >64 [](http://example.com/codeflow?start=22&length=2)

This should be 32, and should probably be SYMCRYPT_MLDSA_PRIVATE_SEED_SIZE rather than a magic number

SymCryptProvider/src/decoder/p_scossl_decode_mldsa.c
```diff
@@ -0,0 +1,184 @@
...
+        ERR_raise(ERR_LIB_PROV, ERR_R_PASSED_INVALID_ARGUMENT);
+        return SCOSSL_FAILURE;
+    }
+
+    return p_scossl_mldsa_keymgmt_export(keyCtx, ctx->desc->selection, exportCb, exportCbArg);
+}
+
+#define SCOSSL_MAKE_MLDSA_DECODER(decoderType, bits)                                                        \
+    static SCOSSL_MLDSA_KEY_CTX                                                                             \
+    *p_scossl_##decoderType##_to_mldsa##bits(_In_ SCOSSL_DECODE_CTX *ctx,  _In_ BIO *bio)                   \
```
samuel-lee-msft: >bits [](http://example.com/codeflow?start=40&length=4)

NP: paramSet

SymCryptProvider/src/encoder/p_scossl_encode_mldsa.c
```diff
@@ -0,0 +1,501 @@
...
+    ASN1_OBJECT *p8Obj;
+    SCOSSL_STATUS status = SCOSSL_FAILURE;
+
+    if (keyCtx->key == NULL)
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_MISSING_KEY);
+        goto cleanup;
+    }
+
+    if (p_scossl_mldsa_keymgmt_get_encoded_key(keyCtx, SYMCRYPT_MLDSAKEY_FORMAT_PUBLIC_KEY, &pbKey, &cbKey) != SCOSSL_SUCCESS)
```
samuel-lee-msft: >p_scossl_mldsa_keymgmt_get_encoded_key [](http://example.com/codeflow?start=8&length=38)

Should align this with ML-KEM per [https://github.com/microsoft/SymCrypt-OpenSSL/pull/171#discussion_r3516924071](https://github.com/microsoft/SymCrypt-OpenSSL/pull/171#discussion_r3516924071) - either both let get_encoded_key allocate, or both allocate the public key

SymCryptProvider/src/encoder/p_scossl_encode_mldsa.c
```diff
@@ -0,0 +1,501 @@
...
+        }
+
+        if (p_scossl_mldsa_keymgmt_get_encoded_key(keyCtx, keyCtx->format, &pbKey, &cbKey) != SCOSSL_SUCCESS)
+        {
+            goto cleanup;
+        }
+
+        if (keyCtx->format == SYMCRYPT_MLDSAKEY_FORMAT_PRIVATE_SEED)
+        {
+            if (BIO_printf(out, "MLDSA Private-Key (512 bit private seed encoding):\nprivate-seed") <= 0)
```
samuel-lee-msft: >512 [](http://example.com/codeflow?start=52&length=3)

256

SymCryptProvider/src/encoder/p_scossl_encode_mldsa.c
```diff
@@ -0,0 +1,501 @@
...
+    {                                                                                                               \
+        return p_scossl_encode_newctx(provctx, &p_scossl_mldsa_##encoderType##_pem_desc);                           \
+    }                                                                                                               \
+                                                                                                                    \
+    static BOOL p_scossl_der_to_mldsa_##encoderType##_does_selection(ossl_unused void *provctx, int selection)      \
+    {                                                                                                               \
+        return p_scossl_encode_does_selection(&p_scossl_mldsa_##encoderType##_der_desc, selection);                 \
+    }
+
+#define MLDSA_ENCODER_DISPATCH(encoderType, bits)                                                                   \
```
samuel-lee-msft: >bits [](http://example.com/codeflow?start=44&length=4)

one more bits->paramSet nitpick

samuel-lee-msft (dismissed):
LGTM modulo changes

SymCryptProvider/src/signature/p_scossl_mldsa.h
mamckee: Yeah, leftover from a previous iteration. We would have needed it for ML-DSA composite, but that's going to be implemented in SymCrypt now.

SymCryptProvider/src/keymgmt/p_scossl_mldsa_keymgmt.c
mamckee: So I think when I originally added this I was looking at the OQS provider as reference. I don't think it's strictly necessary, but it's relatively cheap to keep (just acting as an alias for `OSSL_PKEY_PARAM_PUB_KEY`). I would lean towards leaving it in for now, especially since this isn't a hot path.

SymCryptProvider/src/keymgmt/p_scossl_mldsa_keymgmt.c
mamckee: That's what we're doing for ML-KEM too.

samuel-lee-msft (approved):

PR #163, opened 2026-04-01
Add generic and AES SKEY support

mamckee:
OpenSSL 3.5 added [SKEYs](https://docs.openssl.org/3.5/man7/provider-skeymgmt) for opaque symmetric key management. This PR adds support for the new key types to the SymCrypt provider.

- Add generic and AES SKEY keymgmt interfaces
- Add support for AES SKEYs to AES cipher interfaces
- Add support for AES SKEYs to CMAC interface

### DIFF
--- a/SymCryptProvider/CMakeLists.txt
+++ b/SymCryptProvider/CMakeLists.txt
@@ -11,7 +11,7 @@ configure_file(
     ${CMAKE_CURRENT_SOURCE_DIR}/inc/p_scossl_base.h.in
     ${CMAKE_CURRENT_SOURCE_DIR}/inc/p_scossl_base.h)
 
-find_package(OpenSSL REQUIRED)
+find_package(OpenSSL 3.5 REQUIRED)
 include_directories(${OPENSSL_INCLUDE_DIR})
 
 find_library(SYMCRYPT_LIBRARY symcrypt PATHS ${CMAKE_SOURCE_DIR})
@@ -52,18 +52,14 @@ set(SCOSSL_SOURCES
     ./src/mac/p_scossl_kmac.c
     ./src/signature/p_scossl_ecdsa_signature.c
     ./src/signature/p_scossl_rsa_signature.c
+    ./src/skeymgmt/p_scossl_aes_skeymgmt.c
+    ./src/skeymgmt/p_scossl_generic_skeymgmt.c
     ./src/p_scossl_bio.c
     ./src/p_scossl_ecc.c
     ./src/p_scossl_rand.c
     ./src/p_scossl_rsa.c
     ./src/p_scossl_base.c
 )
-
-if (${OPENSSL_VERSION} VERSION_GREATER_EQUAL 3.5)
-    list(APPEND SCOSSL_SOURCES
-        ./src/skeymgmt/p_scossl_generic_skeymgmt.c
-        ./src/skeymgmt/p_scossl_aes_skeymgmt.c)
-endif()
 
 if (KEYSINUSE_ENABLED)
     list(APPEND SCOSSL_SOURCES ../KeysInUse/keysinuse.c)
--- a/SymCryptProvider/CMakeLists.txt
+++ b/SymCryptProvider/CMakeLists.txt
@@ -59,6 +59,12 @@ set(SCOSSL_SOURCES
     ./src/p_scossl_base.c
 )
 
+if (${OPENSSL_VERSION} VERSION_GREATER_EQUAL 3.5)
+    list(APPEND SCOSSL_SOURCES
+        ./src/skeymgmt/p_scossl_generic_skeymgmt.c
+        ./src/skeymgmt/p_scossl_aes_skeymgmt.c)
+endif()
+
 if (KEYSINUSE_ENABLED)
     list(APPEND SCOSSL_SOURCES ../KeysInUse/keysinuse.c)
 endif()
--- a/SymCryptProvider/src/ciphers/p_scossl_aes.c
+++ b/SymCryptProvider/src/ciphers/p_scossl_aes.c
@@ -10,6 +10,7 @@
 #include "scossl_helpers.h"
 #include "p_scossl_base.h"
 #include "p_scossl_aes.h"
+#include "p_scossl_skey.h"
 
 #ifdef __cplusplus
 extern "C" {
@@ -148,12 +149,26 @@ static SCOSSL_STATUS p_scossl_aes_generic_decrypt_init(_Inout_ SCOSSL_AES_CTX *c
     return p_scossl_aes_generic_init_internal(ctx, FALSE, key, keylen, iv, ivlen, params);
 }
 
+static SCOSSL_STATUS p_scossl_aes_generic_skey_encrypt_init(_Inout_ SCOSSL_AES_CTX *ctx, _In_ SCOSSL_SKEY *skey,
+                                                            _In_reads_bytes_opt_(ivlen) const unsigned char *iv, size_t ivlen,
+                                                            _In_ const OSSL_PARAM params[])
+{
+    return p_scossl_aes_generic_init_internal(ctx, TRUE, skey->pbKey, skey->cbKey, iv, ivlen, params);
+}
+
+static SCOSSL_STATUS p_scossl_aes_generic_skey_decrypt_init(_Inout_ SCOSSL_AES_CTX *ctx, _In_ SCOSSL_SKEY *skey,
+                                                            _In_reads_bytes_opt_(ivlen) const unsigned char *iv, size_t ivlen,
+                                                            _In_ const OSSL_PARAM params[])
+{
+    return p_scossl_aes_generic_init_internal(ctx, FALSE, skey->pbKey, skey->cbKey, iv, ivlen, params);
+}
+
 #define SYMCRYPT_OPENSSL_MASK8_SELECT( _mask, _a, _b ) (SYMCRYPT_FORCE_READ8(&_mask) & _a) | (~(SYMCRYPT_FORCE_READ8(&_mask)) & _b)
 
 // Verifies the TLS padding from the end of record, extracts the MAC from the end of
-// the unpadded record, and saves the result to ctx->tlsMac. 
-// 
-// If ctx->tlsMacSize is 0 (in the case of encrypt-then-mac), no MAC is extracted, 
+// the unpadded record, and saves the result to ctx->tlsMac.
+//
+// If ctx->tlsMacSize is 0 (in the case of encrypt-then-mac), no MAC is extracted,
 // but the padding is still verified and removed.
 //
 // The MAC will later be fetched through p_scossl_aes_generic_get_ctx_params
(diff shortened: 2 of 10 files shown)


SymCryptProvider/src/ciphers/p_scossl_aes.c
```diff
@@ -1028,6 +1043,8 @@ static SCOSSL_STATUS scossl_aes_cfb8_cipher(_Inout_ SCOSSL_AES_CTX *ctx,
         {OSSL_FUNC_CIPHER_GETTABLE_PARAMS, (void (*)(void))p_scossl_aes_generic_gettable_params},         \
         {OSSL_FUNC_CIPHER_GETTABLE_CTX_PARAMS, (void (*)(void))p_scossl_aes_generic_gettable_ctx_params}, \
         {OSSL_FUNC_CIPHER_SETTABLE_CTX_PARAMS, (void (*)(void))p_scossl_aes_generic_settable_ctx_params}, \
+        {OSSL_FUNC_CIPHER_ENCRYPT_SKEY_INIT, (void (*)(void))p_scossl_aes_generic_skey_encrypt_init},     \
+        {OSSL_FUNC_CIPHER_DECRYPT_SKEY_INIT, (void (*)(void))p_scossl_aes_generic_skey_decrypt_init},     \
```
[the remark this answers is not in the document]
mamckee: All supported platforms for SymCrypt provider 1.10+ will be on OpenSSL 3.5+

SymCryptProvider/src/ciphers/p_scossl_aes_aead.c
```diff
@@ -579,6 +608,8 @@ static SCOSSL_STATUS p_scossl_aes_ccm_set_ctx_params(_Inout_ SCOSSL_CIPHER_CCM_C
         {OSSL_FUNC_CIPHER_GETTABLE_PARAMS, (void (*)(void))p_scossl_aes_generic_gettable_params},            \
         {OSSL_FUNC_CIPHER_GETTABLE_CTX_PARAMS, (void (*)(void))p_scossl_aes_##lcmode##_gettable_ctx_params}, \
         {OSSL_FUNC_CIPHER_SETTABLE_CTX_PARAMS, (void (*)(void))p_scossl_aes_##lcmode##_settable_ctx_params}, \
+        {OSSL_FUNC_CIPHER_ENCRYPT_SKEY_INIT, (void (*)(void))p_scossl_aes_##lcmode##_skey_encrypt_init},     \
+        {OSSL_FUNC_CIPHER_DECRYPT_SKEY_INIT, (void (*)(void))p_scossl_aes_##lcmode##_skey_decrypt_init},     \
```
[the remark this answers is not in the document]
mamckee: All supported platforms for SymCrypt provider 1.10+ will be on OpenSSL 3.5+

SymCryptProvider/src/ciphers/p_scossl_aes_xts.c
```diff
@@ -274,6 +288,8 @@ static SCOSSL_STATUS p_scossl_aes_xts_set_ctx_params(_Inout_ SCOSSL_AES_XTS_CTX
         {OSSL_FUNC_CIPHER_GETTABLE_PARAMS, (void (*)(void))p_scossl_aes_generic_gettable_params},     \
         {OSSL_FUNC_CIPHER_GETTABLE_CTX_PARAMS, (void (*)(void))p_scossl_aes_xts_gettable_ctx_params}, \
         {OSSL_FUNC_CIPHER_SETTABLE_CTX_PARAMS, (void (*)(void))p_scossl_aes_xts_settable_ctx_params}, \
+        {OSSL_FUNC_CIPHER_ENCRYPT_SKEY_INIT, (void (*)(void))p_scossl_aes_xts_skey_encrypt_init},     \
+        {OSSL_FUNC_CIPHER_DECRYPT_SKEY_INIT, (void (*)(void))p_scossl_aes_xts_skey_decrypt_init},     \
         {0, NULL}};
```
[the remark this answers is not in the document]
mamckee: All supported platforms for SymCrypt provider 1.10+ will be on OpenSSL 3.5+

SymCryptProvider/src/skeymgmt/p_scossl_aes_skeymgmt.c
```diff
@@ -0,0 +1,92 @@
...
+    return skey;
+}
+
+static SCOSSL_STATUS p_scossl_aes_skeymgmt_export(_In_ SCOSSL_SKEY *skey, int selection,
+                                                  _In_ OSSL_CALLBACK *param_cb, _In_ void *cbarg)
+{
+    if (skey->type != SCOSSL_SKEY_TYPE_AES)
+    {
+        return SCOSSL_FAILURE;
+    }
```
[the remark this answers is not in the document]
mamckee: This follows the OpenSSL implementation and assumes that skey isn't NULL.

SymCryptProvider/src/skeymgmt/p_scossl_generic_skeymgmt.c
```diff
@@ -0,0 +1,196 @@
...
+            ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+            goto cleanup;
+        }
+
+        skey->cbKey = cbKey;
+        memcpy(skey->pbKey, pcbKey, cbKey);
+    }
+
+    status = SCOSSL_SUCCESS;
+
```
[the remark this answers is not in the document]
mamckee: This follows the OpenSSL implementation. Status indicates whether or not key import failed and a partially initialized key may need to be cleaned up. In this case, no key has been created to cleanup necessarily won't run.

SymCryptProvider/src/skeymgmt/p_scossl_generic_skeymgmt.c
```diff
@@ -0,0 +1,196 @@
...
+        {
+            ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+            goto cleanup;
+        }
+
+        SymCryptRandom(skey->pbKey, cbKey);
+        skey->cbKey = cbKey;
+
+    }
+
```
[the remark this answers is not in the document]
mamckee: See previous comment

MS-megliu (commented):
(see inline comments)

SymCryptProvider/src/skeymgmt/p_scossl_generic_skeymgmt.c
```diff
@@ -0,0 +1,196 @@
...
+    status = SCOSSL_SUCCESS;
+
```
MS-megliu, line 164: Bug: If `OSSL_SKEY_PARAM_KEY_LENGTH` is not present in `params`, `OSSL_PARAM_locate_const` returns NULL, the function skips the entire key generation block, falls through to `status = SCOSSL_SUCCESS` here, and returns a NULL `skey` as success. The caller gets a NULL key with no error indication.

Same issue in `p_scossl_generic_skeymgmt_import` (line 89 of this file) — if `OSSL_SKEY_PARAM_RAW_BYTES` is absent, returns NULL without raising an error.

Suggestion: treat a missing required param as an error — `ERR_raise(ERR_LIB_PROV, PROV_R_MISSING_KEY)` or equivalent before `goto cleanup`.

SymCryptProvider/src/skeymgmt/p_scossl_aes_skeymgmt.c
```diff
@@ -0,0 +1,91 @@
...
+static SCOSSL_SKEY *p_scossl_aes_skeymgmt_import(_In_ SCOSSL_PROVCTX *provctx, int selection, _In_ const OSSL_PARAM params[])
+{
+    SCOSSL_SKEY *skey = p_scossl_generic_skeymgmt_import(provctx, selection, params);
+
+    if (skey != NULL)
+    {
+        skey->type = SCOSSL_SKEY_TYPE_AES;
+
+        if (skey->cbKey != 16 && skey->cbKey != 24 && skey->cbKey != 32)
+        {
```
MS-megliu: Nit: AES key length is validated after `p_scossl_generic_skeymgmt_import` has already allocated secure memory and copied the key material. For invalid lengths, the key is allocated, copied, then immediately freed. Consider validating the length from the OSSL_PARAM before calling the generic import to avoid the wasted `OPENSSL_secure_malloc` + `memcpy`. Same applies to `p_scossl_aes_skeygen_generate` below — `SymCryptRandom` is called for invalid lengths.

SymCryptProvider/src/ciphers/p_scossl_aes_xts.c
```diff
@@ -120,6 +121,19 @@ static SCOSSL_STATUS p_scossl_aes_xts_decrypt_init(_Inout_ SCOSSL_AES_XTS_CTX *c
     return p_scossl_aes_xts_init_internal(ctx, 0, key, keylen, iv, ivlen, params);
 }
 
+static SCOSSL_STATUS p_scossl_aes_xts_skey_encrypt_init(_Inout_ SCOSSL_AES_XTS_CTX *ctx, _In_ SCOSSL_SKEY *skey,
+                                                        _In_reads_bytes_opt_(ivlen) const unsigned char *iv, size_t ivlen,
+                                                        _In_ const OSSL_PARAM params[])
+{
+    return p_scossl_aes_xts_init_internal((SCOSSL_AES_XTS_CTX *)ctx, 1, skey->pbKey, skey->cbKey, iv, ivlen, params);
+}
```
MS-megliu, line 129: Nit: Redundant cast — `ctx` is already declared as `SCOSSL_AES_XTS_CTX *ctx` in the function signature, so `(SCOSSL_AES_XTS_CTX *)ctx` is unnecessary.

SymCryptProvider/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ configure_file(
     ${CMAKE_CURRENT_SOURCE_DIR}/inc/p_scossl_base.h.in
     ${CMAKE_CURRENT_SOURCE_DIR}/inc/p_scossl_base.h)
 
-find_package(OpenSSL REQUIRED)
+find_package(OpenSSL 3.5 REQUIRED)
 include_directories(${OPENSSL_INCLUDE_DIR})
```
MS-megliu, line 15: Question: This bumps the minimum OpenSSL requirement from any version to 3.5 globally, making the entire provider unbuildable with OpenSSL 3.0-3.4. The SKEYMGMT dispatch registration in `p_scossl_base.c` already uses `#ifdef OSSL_OP_SKEYMGMT` guards — was making 3.5 the hard minimum intentional? If not, the new source files could be conditionally compiled and the `find_package` left as-is.

SymCryptProvider/src/skeymgmt/p_scossl_aes_skeymgmt.c
mamckee: There's a tradeoff here. Getting the length before generic import requires 2 parameter fetches. The normal case will be a valid length so we'd be optimizing the invalid case by checking beforehand. This is the approach OpenSSL is using too: https://github.com/openssl/openssl/blob/72d5e8dcd2977234e47e840d2173917daa8bb0fa/providers/implementations/skeymgmt/aes_skmgmt.c#L19.

It's a simple operation so realistically I guess we don't need to call the generic function really.

SymCryptProvider/src/skeymgmt/p_scossl_generic_skeymgmt.c
mamckee: So the default OpenSSL implementation doesn't write any errors here, so it's hard to tell if this should be considered an error or not. There are other places where the caller can pass empty key data as a valid case (the result is an empty key that isn't particularly useful). I guess in this case though since we return NULL, it probably should be marked as an error (as opposed to a context with no key data, but maybe some valid domain parameters set).

SymCryptProvider/CMakeLists.txt
mamckee: Forgot to remove the include guards in p_scossl_base. We don't have any new platforms we need to target SCOSSL 1.10 for that use OpenSSL < 3.5 so we can enforce this as the minimum version. Some of the other features (e.g. ML-KEM) can still be backported to an older version like 1.9, but this change specifically should only ever land on platforms with OpenSSL 3.5+

MS-megliu (approved):
LGTM

PR #166, opened 2026-04-21
Add composite signature interface for RSA and ECDSA signatures

mamckee:
OpenSSL 3.5 introduced hardcoded composite signature + digest interfaces for RSA and ECDSA. This PR adds the following composite signature interfaces to the SymCrypt provider, implementing the sign/verify message functions in place of sign/verify digest functions. The original RSA and ECDSA signature interfaces are still available.


SymCryptProvider/src/signature/p_scossl_ecdsa_sigalg_signature.c
```diff
@@ -0,0 +1,346 @@
...
+
+    if (!ctx->allowFinal)
+    {
+        ERR_raise(ERR_LIB_PROV, PROV_R_FINAL_CALL_OUT_OF_ORDER);
+        return SCOSSL_FAILURE;
+    }
+
+    if (!EVP_DigestFinal_ex(ctx->mdctx, digest, &cbDigest))
+    {
+        return SCOSSL_FAILURE;
```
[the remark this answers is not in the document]
mamckee: This matches the OpenSSL default behavior. -1 is only returned by internal ECDSA functions (e.g. padding failures)

SymCryptProvider/src/signature/p_scossl_ecdsa_sigalg_signature.c
```diff
@@ -0,0 +1,346 @@
...
+        if ((ctx->pbSignature = OPENSSL_memdup(sig, siglen)) == NULL)
+        {
+            ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+            return -1;
+        }
+        ctx->cbSignature = siglen;
+
+        if (p_scossl_ecdsa_sigalg_signverify_message_update(ctx, tbs, tbslen) <= 0)
+        {
+            return 0;
```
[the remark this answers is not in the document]
mamckee: This matches the openssl implementation

SymCryptProvider/src/signature/p_scossl_rsa_sigalg_signature.c
```diff
@@ -0,0 +1,376 @@
...
+    {
+        return SCOSSL_FAILURE;
+    }
+
+    ctx->allowUpdate = FALSE;
+    ctx->allowFinal = FALSE;
+    ctx->allowOneshot = FALSE;
+
+    return p_scossl_rsa_verify_internal(ctx, ctx->pbSignature, ctx->cbSignature, abDigest, cbDigest);
+}
```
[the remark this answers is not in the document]
mamckee: SymCrypt checks the signature size

SymCryptProvider/src/signature/p_scossl_ecdsa_sigalg_signature.c
```diff
@@ -0,0 +1,385 @@
...
+
+    if (!EVP_DigestFinal_ex(ctx->mdctx, digest, &cbDigest))
+    {
+        return 0;
+    }
+
+    ctx->allowUpdate = FALSE;
+    ctx->allowFinal = FALSE;
+    ctx->allowOneshot = FALSE;
+
```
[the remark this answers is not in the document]
mamckee: OpenSSL implementation doesn't check this, but the check isn't a bad idea and should be safe. The default behavior would be a crash in this case, so we don't regress existing scenarios.

SymCryptEngine/src/e_scossl_ecc.c
```diff
@@ -606,7 +606,11 @@ ECDSA_SIG* e_scossl_eckey_sign_sig(_In_reads_bytes_(dgstlen) const unsigned char
...
 }
 
-SCOSSL_STATUS e_scossl_eckey_verify(int type, _In_reads_bytes_(dgst_len) const unsigned char* dgst, int dgst_len,
+// Return
+// 1 (SCOSSL_SUCCESS) for valid signature
+// 0 (SCOSSL_FAILURE) for invalid signature
+// -1 for error
+int e_scossl_eckey_verify(int type, _In_reads_bytes_(dgst_len) const unsigned char* dgst, int dgst_len,
                           _In_reads_bytes_(sig_len) const unsigned char* sigbuf, int sig_len, _In_ EC_KEY* eckey)
 {
```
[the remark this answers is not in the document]
mamckee: Error returns match OpenSSL default behavior

SymCryptProvider/src/signature/p_scossl_rsa_signature.c
```diff
@@ -152,12 +142,20 @@ static SCOSSL_STATUS p_scossl_rsa_signverify_init(_Inout_ SCOSSL_RSA_SIGN_CTX *c
         return SCOSSL_FAILURE;
     }
 
-#ifdef RSA_PSS_SALTLEN_AUTO_DIGEST_MAX
     ctx->cbSalt = RSA_PSS_SALTLEN_AUTO_DIGEST_MAX;
```
MS-megliu, line 148: `RSA_PSS_SALTLEN_AUTO_DIGEST_MAX` is used unguarded here. This macro is unavailable on OpenSSL 3.0, so this regresses 3.0 builds unless we explicitly drop support. you may want to restore version guards (fallback to `RSA_PSS_SALTLEN_AUTO`) or raise/document the minimum supported OpenSSL version.

SymCryptEngine/src/e_scossl_ecc.c
MS-megliu, line 615: Function comment declares tri-state semantics (1 valid / 0 invalid / -1 error), but several internal-error branches still return `SCOSSL_FAILURE` (0). That makes operational failures indistinguishable from invalid signatures. May consider return `-1` consistently on internal/operational errors, or update the contract if tri-state is not intended.

SymCryptProvider/src/signature/p_scossl_rsa_sigalg_signature.c
```diff
@@ -0,0 +1,383 @@
...
+                                                         _In_ const OSSL_PARAM params[])                            \
+    {                                                                                                               \
+        return p_scossl_rsa_sigalg_signverify_init(ctx, keyCtx, params, EVP_PKEY_OP_VERIFY, MD);                    \
+    }                                                                                                               \
+                                                                                                                    \
+    static SCOSSL_STATUS p_scossl_rsa_##md##_sign_message_init(_Inout_ SCOSSL_RSA_SIGN_CTX *ctx,                    \
+                                                               _In_ SCOSSL_PROV_RSA_KEY_CTX *keyCtx,                \
+                                                               _In_ const OSSL_PARAM params[])                      \
+    {                                                                                                               \
+        return p_scossl_rsa_sigalg_signverify_message_init(ctx, keyCtx, params, EVP_PKEY_OP_SIGNMSG, MD);           \
```
MS-megliu, line 338: This path relies on SIGN_MESSAGE/VERIFY_MESSAGE operation macros and dispatch semantics that are not present in OpenSSL 3.0–3.3. You may want to gate this implementation/registration with OpenSSL version checks (or `#ifdef` for the message-signature symbols), or explicitly require 3.4+.

SymCryptProvider/src/signature/p_scossl_ecdsa_sigalg_signature.c
```diff
@@ -0,0 +1,391 @@
...
```
MS-megliu, line 174: you may want to enforce fixed-digest length on verify for sigalg operations (for example, reject when `tbslen != ctx->mdSize` for `ECDSA-SHA*`). Sign path validates digest size, but verify currently accepts arbitrary lengths before calling internal verify, which can diverge from expected sigalg behavior.

MS-megliu (commented):
please see comments inline.

SymCryptProvider/src/signature/p_scossl_rsa_signature.c
mamckee: Main branch and future SCOSSL versions won't support OpenSSL 3.0 so I removed this intentionally.

SymCryptEngine/src/e_scossl_ecc.c
mamckee: The expectation from OpenSSL is a bit inconsistent. This function only returns -1 in the cases that the default OpenSSL returns -1. _Technically_ OpenSSL documentation does say internal errors should return -1, but in practice that's only done a handful of places in ecdsa verify.

SymCryptProvider/src/signature/p_scossl_rsa_sigalg_signature.c
mamckee: Main branch and future versions only support OpenSSL 3.5+

SymCryptProvider/src/signature/p_scossl_ecdsa_sigalg_signature.c
mamckee: Digest size in this function is dictated by ctx->mdctx, which is set internally based on the specific algorithm combo, so we shouldn't have a case where cbDigest != ctx->mdSize

SymCryptProvider/src/signature/p_scossl_ecdsa_signature.c
```diff
@@ -111,6 +105,10 @@ static SCOSSL_STATUS p_scossl_ecdsa_signverify_init(_Inout_ SCOSSL_ECDSA_CTX *ct
     }
 
     ctx->operation = operation;
+    ctx->isSigalg = FALSE;
+    ctx->allowUpdate = TRUE;
+    ctx->allowFinal = TRUE;
+    ctx->allowOneshot = TRUE;
```
MS-megliu, line 117: This ECDSA init path resets the sigalg operation flags, but it doesn't clear the cached VERIFYMSG signature (`pbSignature`/`cbSignature`). If the same context previously performed a VERIFYMSG operation and is reinitialized, `p_scossl_ecdsa_sigalg_verify_message_final()` can pass its `pbSignature != NULL` guard and verify using the stale signature when the caller has not supplied a new `OSSL_SIGNATURE_PARAM_SIGNATURE`. RSA clears this state in `p_scossl_rsa_signverify_init`; ECDSA may want to do the same here.

SymCryptProvider/src/signature/p_scossl_rsa_sigalg_signature.c
```diff
@@ -0,0 +1,383 @@
...
+    if (!EVP_DigestFinal_ex(ctx->mdctx, abDigest, &cbDigest))
+    {
+        return SCOSSL_FAILURE;
+    }
+
+    ctx->allowUpdate = FALSE;
+    ctx->allowFinal = FALSE;
+    ctx->allowOneshot = FALSE;
+
+    return p_scossl_rsa_verify_internal(ctx, ctx->pbSignature, ctx->cbSignature, abDigest, cbDigest);
```
MS-megliu, line 173: After the merge from main, this call now fails for RSA `VERIFY_MESSAGE`: `p_scossl_rsa_verify_internal()` rejects any operation other than `EVP_PKEY_OP_VERIFY`, but this sigalg path is initialized with `EVP_PKEY_OP_VERIFYMSG`. That means `p_scossl_rsa_sigalg_verify_message_final()` reaches this helper with a valid VERIFY_MESSAGE operation and gets `ERR_R_OPERATION_FAIL` before doing the actual verification. Consider allowing `EVP_PKEY_OP_VERIFYMSG` in the internal helper, or routing VERIFY_MESSAGE through a helper that doesn't reject that operation.

MS-megliu (approved):
LGTM

PR #167, opened 2026-04-27
Support low-order, on-twist, and non-canonical x25519 keys

mamckee:
The SymCrypt provider does not currently support two groups of less common x25519 keys.
1. **Low order/on-twist keys**: `SymCryptEcpointSetValue` validates prime-order subgroup membership (`[GOrd]*P == O`) by default, which rejects these keys. OpenSSL does not check this any permits these keys.
2. **Non-canonical keys**: The 20 non-canonical values 2^255 - 19 through 2^255 - 1 for X25519 are rejected by SymCryptEcpointSetValue. If these are cononicalized by the SymCrypt provider (reduced modulo _p_) then they will be accepted/

- Added `SYMCRYPT_FLAG_KEY_NO_FIPS` to `SymCryptEcpointSetValue` calls for x25519
  - Allows import or low order/on-twist keys 
  - Increases performance of x25519 import
  - x25519 isn't FIPS anyways, so we should skip this check regardless
- Added helper function `p_scossl_x25519_canonicalize_public_key` to canonicalize non-canonical public key values before sending them to SymCrypt
- Added tests for low order/on-twist keys and exhaustive tests for non-canonical keys to EVP test file


SymCryptProvider/src/p_scossl_ecc.c
```diff
@@ -120,13 +147,18 @@ SCOSSL_ECC_KEY_CTX *p_scossl_ecc_dup_ctx(SCOSSL_ECC_KEY_CTX *keyCtx, int selecti
                 goto cleanup;
             }
 
+            if (keyCtx->isX25519)
+            {
+                flags |= SYMCRYPT_FLAG_KEY_NO_FIPS;
```
vcsjones, line 152: One thing I observed from enabling this is this will permit SymCrypt to produce "all zero" shared secrets when allowing low-order public keys. Most other X25519 implementations will refuse to give back the all zero shared secret (RFC 7748 recommends it)


e.g. Wycheproof test case 64.

The way we handled this on the .NET side is to do a secure compare of the produced shared secret with the all-zero key and not permit it.

mamckee: Do you have a repro? SymCrypt is supposed to reject all zero secrets, and I saw rejections when testing: https://github.com/microsoft/SymCrypt/blob/2864ff0da96642049a66808d86a8bad2736b85e8/lib/ec_dh.c#L129.

One of the test cases I added is a low-order point from wycheproof that gets rejected.

vcsjones: Hm, could be a bcrypt specific behavior. Feel free to resolve this one then and will take bcrypt offline.

samuel-lee-msft: https://github.com/microsoft/SymCrypt/blob/2864ff0da96642049a66808d86a8bad2736b85e8/lib/ec_dh.c#L129 
does not reject an all zero shared secret, it rejects a shared secret which is the _identity point_ (aka the point at infinity).

In the X/Z single projective coordinates of X25519, that is a check for Z being 0.
The shared secret from ECDH is the X coordinate when you transform back into affine coordinates (actually do the modular division X -> X/Z).

The check for all-zero secret seems to be a check to reject the use of low order public keys at secret agreement time:
https://www.rfc-editor.org/rfc/rfc7748#section-6.1
"The check for the all-zero value results from the fact that the
   X25519 function produces that value if it operates on an input
   corresponding to a point with small order, where the order divides
   the cofactor of the curve (see Section 7).  The check may be
   performed by ORing all the bytes together and checking whether the
   result is zero, as this eliminates standard side-channels in software
   implementations."
These public key were previously rejected at public key import time with the FIPS checks, but now we're going to allow them to be imported but fail every attempt to do secret agreement with them?
It seems super wonky for .NET to be complaining that we were failing too fast.

Given the RFC documentation recommending we should reject these secret agreement values, we should probably add that test to X25519 in SymCrypt when FIPS checks are disabled to align with RFC expectations.
There is a clear performance benefit to not doing the public key order validation on public key import for the expected case, and instead checking at secret agreement time.

samuel-lee-msft: Recommend zero shared secret check to go into SCOSSL in this PR, and move the check down into SymCrypt as a follow up.

vcsjones: > These public key were previously rejected at public key import time with the FIPS checks, but now we're going to allow them to be imported but fail every attempt to do secret agreement with them?
It seems super wonky for .NET to be complaining that we were failing too fast.

It's not just about failing too fast. In the case of low order it is failing too fast, but on-twist does not fail and should be permitted. It just happens that the FIPS check gates several things.

vcsjones: >  but fail every attempt to do secret agreement with them?

I don't think every attempt will fail. For example, Wycheproof test case 4 is on-twist and does not produce an all-zero shared secret.

samuel-lee-msft: You're right, there are several classes of X25519 point, and I oversimplified.

The FIPS public key order check rejects all but the prime-order subgroup, which are the only public keys you should expect from a peer which has performed keygen correctly and has managed to send it to you without corruption (which is why we haven't got a lot of complaints about being overly restrictive to date)


In the absence of the FIPS check at public key import time, we permit low order and on-twist points
My understanding is that low order points should always produce an all-zero shared secret
On-twist points will map to on-twist results which are high entropy (and the malicious/faulty peer will not be able to generate the same result so the E2E handshake would fail later)


Net net, removing FIPS checks can defer failure with bad public key to later in the process, reducing cost of the expected success path for good public keys
It is reasonable to reject 0 shared secrets to avoid games with a malicious peer guaranteeing you negotiate a zero-security shared secret (although a malicious peer could just publish the shared secret if they were so inclined, so it's probably not a huge deal)

mamckee: Following offline conversation, this check is sufficient, since _x_ is only ever 0 when the affine point is (0,0).

samuel-lee-msft (approved):

PR #174, opened 2026-08-06
Improve KeysInUse logging performance 

mamckee:
This PR makes a few small performance improvements to keysinuse.

- Minimizes the time the logging thread mutex is held during logging to continue ensuring fork safety, but avoid blocking crypto operations if a large number of sign/decrypt operations are fired at once.
- Set the logging thread signal mask to empty to prevent the logging thread from running any signal handlers.
- Added config options to allow applications to configure keysinuse forking behavior (run in parent and child by default, or exclusively run in the parent or child process)


SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -1025,48 +1049,53 @@ static void *p_scossl_keysinuse_logging_thread_start(ossl_unused void *arg)
...
+            }
+            else
+            {
+                p_scossl_keysinuse_log_error("Failed to lock keysinuse info,OPENSSL_%d", ERR_get_error());
+                keysinuseInfoTmp.refCount = -1;
             }
 
             pthread_mutex_unlock(&logging_thread_mutex);
+
+            p_scossl_keysinuse_info_free(pKeysinuseInfo);
```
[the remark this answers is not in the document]
mamckee: Addressed in cbf87aa: `p_scossl_keysinuse_info_free` is now called while `logging_thread_mutex` is still held, so the reference release (which can touch the info lock via `CRYPTO_atomic_add`'s fallback) can no longer overlap with the atfork prepare handler. The slow log write remains outside the mutex.

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -107,6 +110,72 @@ static void p_scossl_keysinuse_logging_thread_cleanup()
     }
 }
 
+// Starts the logging thread and marks keysinuse as enabled. Reainitializes
```
nnmkhang: NIT spelling:  
Reainitializes  ->  Reinitializes 
repsonsible  ->  responsible

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -663,6 +505,28 @@ static void p_scossl_start_keysinuse(_In_ const OSSL_CORE_HANDLE *handle)
             p_scossl_keysinuse_set_logging_delay(atol(confLoggingDelay));
         }
 
+        // Environment overrides config. Config value is alreday fetched core_get_params above
```
nnmkhang: NIT spelling:
alreday -> already

SymCryptProvider/src/p_scossl_base.c
```diff
@@ -663,6 +505,28 @@ static void p_scossl_start_keysinuse(_In_ const OSSL_CORE_HANDLE *handle)
...
+        {
+            if (OPENSSL_strcasecmp(confProcessScope, "main") == 0)
+            {
+                p_scossl_keysinuse_set_process_scope(KEYSINUSE_PROCESS_SCOPE_MAIN);
+            }
+            else if (OPENSSL_strcasecmp(confProcessScope, "child") == 0)
+            {
+                p_scossl_keysinuse_set_process_scope(KEYSINUSE_PROCESS_SCOPE_CHILD);
+            }
+            else
```
nnmkhang, line 524: Should there be another case for if there is a malformed label? IE: something that is not "child" or "main"? The way that it is now would mean that wrong labels (dont know if that can happen) will be treated as both child + main

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -107,6 +110,72 @@ static void p_scossl_keysinuse_logging_thread_cleanup()
...
+        (pthreadErr = pthread_cond_init(logging_thread_cond_wake_early, &attr)) != 0)
+    {
+        p_scossl_keysinuse_log_error("Failed to initialize logging thread condition,SYS_%d", pthreadErr);
+        is_logging = FALSE;
+        goto cleanup;
+    }
+
+    // Block all signals across creation of the logging thread so signal handlers
+    // never run in the logging thread.
+    sigfillset(&blockSigSet);
```
nnmkhang, line 173: this one is from AI. I don't understand the code super well but this seemed important to flag: 


 `sigfillset`  blocks every signal for the logging thread's whole life, including the synchronous fault signals ( SIGSEGV / SIGBUS / SIGFPE / SIGILL / SIGABRT ). Blocking those doesn't prevent a fault (the kernel forces them), and POSIX says delivering a blocked synchronous signal is undefined behavior, so a bug in the logging thread could fail to core-dump and become undebuggable. Blocking the async signals is the right goal; suggest keeping the fault signals deliverable. Also minor: the PR description says the mask is set "to empty," but it's actually set to full/all-blocked.

And a ready-to-apply suggestion (replaces line 155):

```suggestion
    sigfillset(&blockSigSet);
    // Keep synchronous fault signals deliverable so a bug in the logging
    // thread still faults and core-dumps normally.
    sigdelset(&blockSigSet, SIGSEGV);
    sigdelset(&blockSigSet, SIGBUS);
    sigdelset(&blockSigSet, SIGFPE);
    sigdelset(&blockSigSet, SIGILL);
    sigdelset(&blockSigSet, SIGABRT);
```

nnmkhang (dismissed):
LGTM barring `sigfillset`

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -1025,48 +1049,54 @@ static void *p_scossl_keysinuse_logging_thread_start(ossl_unused void *arg)
...
             }
+            else
+            {
+                p_scossl_keysinuse_log_error("Failed to lock keysinuse info,OPENSSL_%d", ERR_get_error());
+                keysinuseInfoTmp.refCount = -1;
+            }
+
+            p_scossl_keysinuse_info_free(pKeysinuseInfo);
 
             pthread_mutex_unlock(&logging_thread_mutex);
```
MS-megliu, line 1131: A fork can occur while remaining entries are held only in this thread-local pending stack. In the child, those entries keep `logPending == TRUE`, so future use is never requeued and the inherited references leak. Could the child handler clean the in-flight batch, or could the remaining entries be returned to the global queue before releasing this mutex?

SymCryptProvider/src/p_scossl_base.c
mamckee: I tend to lean towards falling back to default config value in case of an invalid value. I don't think it really helps to be strict here. I would usually log an error or warning in that case too, but we don't want keysinuse errors surfacing to the caller as openssl errors.

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: Yeah that's fair. I think this is more about being correct to the posix standard but worth doing.

mamckee: Applied in 1a41d36. The logging thread now keeps the synchronous fault signals (SIGSEGV/SIGBUS/SIGFPE/SIGILL, plus SIGABRT) deliverable via `sigdelset`, so a fault in the thread still core-dumps normally. These signals are thread-directed, so leaving them unblocked does not affect isolation of async process-directed signals like SIGIO. Thanks for the catch!

SymCryptProvider/src/p_scossl_keysinuse.c
mamckee: I originally tried to solve this by making the pending stack a global that we could clean up in the child fork handler, but there was a risk of a NULL deref if the logging thread exited early, and the fork handlers ran after. 

The risk of a leak here is one time per fork and won't grow, and the key won't get events for the rest of the application's lifetime. It's a real edge case but I'll see if I can make the pending stack globally accessible safely.

MS-megliu (dismissed):
Re-reviewed the current head. The fork-safety cleanup addresses the prior concern; approved.

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -223,15 +287,18 @@ static void p_scossl_keysinuse_init_once()
...
-    status = SCOSSL_SUCCESS;
+    keysinuse_initialized = TRUE;
 
-cleanup:
-    if (attr_initialized)
+    if ((keysinuse_process_scope & KEYSINUSE_PROCESS_SCOPE_MAIN) == 0)
     {
-        pthread_condattr_destroy(&attr);
+        status = SCOSSL_SUCCESS;
+        goto cleanup;
```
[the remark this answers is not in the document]
mamckee: This is acceptable. We don't want any keysinuse running in the parent if the caller has opted out. This config has limited use for cases where the parent process isn't using the keys before fork.

SymCryptProvider/src/p_scossl_keysinuse.c
```diff
@@ -1025,56 +1092,61 @@ static void *p_scossl_keysinuse_logging_thread_start(ossl_unused void *arg)
...
+                keysinuseInfoTmp.refCount = -1;
+            }
+
+            p_scossl_keysinuse_info_free(pKeysinuseInfo);
 
             pthread_mutex_unlock(&logging_thread_mutex);
+
+            if (keysinuseInfoTmp.refCount > 0)
+            {
+                p_scossl_keysinuse_log_notice("%s,%d,%d,%ld,%ld",
```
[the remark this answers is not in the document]
mamckee: This is a narrow case where a single FD potentially leaks in the child. Truly fixing this involves adding another mutex and complicating this further.

MS-megliu (approved):

PR #175, opened 2026-08-25
Skip unsupported EC curves in KeysInUse tests

mamckee:
This PR skips EC curves for KeysInUse tests programmatically rather than relying on hardcoded NIDs

### DIFF
--- a/test/KeysInUseTest/KeysInUseTest.cpp
+++ b/test/KeysInUseTest/KeysInUseTest.cpp
@@ -13,6 +13,7 @@
 #include "scossl_helpers.h"
 #include "keysinuse.h"
 
+#include <openssl/ec.h>
 #include <openssl/evp.h>
 #if OPENSSL_VERSION_MAJOR == 3
     #include <openssl/core_names.h>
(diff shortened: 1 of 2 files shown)


test/KeysInUseTest/KeysInUseTest.cpp
```diff
@@ -1626,13 +1624,24 @@ SCOSSL_STATUS keysinuse_test_generate_keys()
...
-                goto cleanup;
+                printf("Skipping unsupported curve %s\n", OBJ_nid2sn(testKeys[i].keygenParams));
+                ERR_clear_error();
+                EVP_PKEY_CTX_free(ctx);
+                ctx = NULL;
+                continue;
             }
         }
 
         if (EVP_PKEY_keygen(ctx, &testKeys[i].pkey) <= 0)
```
MS-megliu, line 1645: Do we need to distinguish an unsupported curve from other `EVP_PKEY_keygen()` failures here? Once the curve parameters have been accepted, key generation may still fail for unrelated reasons, such as a provider regression or allocation/initialization failure. Clearing the error and continuing could mask a real test failure as a skip.

MS-megliu (approved):

